PR #50, opened 2016-01-05
Post transaction endpoint

irisli:
This implements the post transaction endpoint by enabling endpoints to be defined as either GET or POST endpoints.

### DIFF
--- a/src/actions/endpointExplorer.js
+++ b/src/actions/endpointExplorer.js
@@ -27,11 +27,21 @@ export function updateValues(param, values) {
 
 export const START_REQUEST = "START_REQUEST"
 export const FINISH_REQUEST = "FINISH_REQUEST"
-export function submitRequest(request) {
+export function submitRequest(url, method, formData) {
   return dispatch => {
     dispatch({type: "START_REQUEST"});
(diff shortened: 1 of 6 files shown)


src/actions/endpointExplorer.js
```diff
@@ -27,11 +27,21 @@ export function updateValues(param, values) {
 
 export const START_REQUEST = "START_REQUEST"
 export const FINISH_REQUEST = "FINISH_REQUEST"
-export function submitRequest(request) {
+export function submitRequest(url, method, formData) {
```
nullstyle: I don't like this API change.  In other areas of the application, the term `request` refers to a javascript object (for example, the `buildRequest` method in `EndpointExplorer`) and is handled passed between units as a single object (for example as a param to `EndpointSetup`), whereas here we're saying a request is a triplet of `(url,method,formData)`.

You should pick one or the other.  Either the components should take 3 params instead of one, or submitRequest should take an object named `request`.

nullstyle:
other than the one comment, +1

irisli:
Thanks for the feedback!

Issue #44, opened 2015-12-19
Form shouldn't clear when you go from tx build to exploration and back


jedmccaleb:

irisli:
The current fix is to clear it.

What is happening under the covers is that the input fields get reset by React but the (backend) store is still there.

The solution is to make the input fields (pickers) controlled components based on the store.

irisli:
Done with the redux picker upgrades

PR #55, opened 2016-01-12
Refactor pickers

irisli:
This PR is a refactor of the Pickers driven by data and the PickerGenerator. This replaces it with a more code-based approach to it.

This also makes pages in the Laboratory persistent between app page change. https://github.com/stellar/laboratory/issues/44


src/actions/endpointExplorer.js
```diff
@@ -17,11 +17,11 @@ export function changePendingRequestProps(props) {
 }
 
 export const UPDATE_VALUES = "UPDATE_VALUES";
-export function updateValues(param, values) {
```
nullstyle: If this action is becoming single-valued, you should reflect it in the name as well:  UPDATE_VALUE

irisli: Thanks

nullstyle:
You've copy-pasted the same markup for a "text box with validation" too many times for me to endorse this change.

Instead, I think you should add `TextPicker` which all of these classes (`PubKeyPicker`, `SecretKeyPicker`, `PositiveIntPicker`, `AmountPicker`, etc.) use instead of repeating code.  Their only differences are the "what validator to use" and "what placeholder to use", as far as I can see.  So, for example:

```js
export default function SecretKeyPicker(props) {
  return <TextPicker 
    {...props} 
    placeholder={'Example: SAEXAMPLE6TLGEF6ASOTVTLFUK7LE2K2PFVPFGTEZMMVHH7KLLBBROEQ'}
    validator={validator}
    />
}
```

nullstyle:
Other than the copy-pasta, +1

irisli:
Thanks for the useful feedback!

PR #56, opened 2016-01-14
General improvements 3

irisli:
The biggest feature in here is support for multiple signers. The other two are changes to styles.

### DIFF
--- a/src/components/FormComponents/MultiPicker.js
+++ b/src/components/FormComponents/MultiPicker.js
@@ -1,0 +1,46 @@
+import React from 'react';
+import _ from 'lodash';
+
+// The value is extended so that at the end there is always an element that is null or undefined.
+// NOTE: The elements that are empty also include empty strings
+export default function MultiPicker(props) {
+  let {value, onUpdate, component} = props;
+  if (!_.isArray(value)) {
+    return <div></div>;
+  }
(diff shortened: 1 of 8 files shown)


src/components/FormComponents/MultiPicker.js
```diff
@@ -0,0 +1,46 @@
...
+
+// The value is extended so that at the end there is always an element that is null or undefined.
+// NOTE: The elements that are empty also include empty strings
+export default function MultiPicker(props) {
+  let {value, onUpdate, component} = props;
+  if (!_.isArray(value)) {
+    return <div></div>;
+  }
+  let lastItem = value[value.length - 1];
+  let extendedValue = value.slice();
```
nullstyle: It's not clear from the code what that `extendedValue` represents.  Could you add some code documentation that explains what is happening in this function?

src/components/FormComponents/MultiPicker.js
```diff
@@ -0,0 +1,46 @@
...
+
+  while (extendedValue.length > 1 && arelastTwoEmpty(extendedValue)) {
```
nullstyle: This loop needs a code comment is well... it's not clear why `areLastTwoEmpty` is involved.

nullstyle:
Other than documenting and improving MultiPicker, +1

PR #60, opened 2016-01-26
fix transaction sequence number bug caused by using JS Number

irisli:
New accounts on the testnet are now created with a starting sequence number greater the native JavaScript Number max number. A new account I created on testnet had the sequence of 9048748768362496 where as the javascript max functional `Number.MAX_SAFE_INTEGER` is 9007199254740991 which is LESS than the sequence number of new accounts.

There was a bug in the laboratory and in js-stellar-base that relied on the insufficient `Number` native value in JavaScript.


jedmccaleb:
r+

PR #62, opened 2016-01-27
Add all operation types to Transaction Builder

irisli:
This PR implements all the operation types in the Transaction Builder. It also adds a note feature so that small help messages can be placed under pickers to assist the user.

image


nullstyle:
+1  Great work @irisli !

PR #68, opened 2016-02-02
implement account sequence number fetcher

irisli:
This also includes inlining `src/components/TxBuilderPanes/TxAttributes.js` into `src/components/TxBuilderAttributes.js`

image


bartekn:
I talked with @cyberomin and @mozzadrella earlier today and we were using Laboratory to build a sample transaction. I realized that in most cases people will just input the next sequence number to create a valid transaction.

What do you think about doing something like this:
* Every time a user enters or changes Source Account field to a valid account ID we load a current sequence number, increment it by one and put it in a Sequence Number input field.
* Alternatively, we can hold this value in a text field (not input field), so inexperienced users don't feel like they need to care about it. Howerer, we add "Change [sequence number]" button/link that will change text field to input field thus making it editable.

So now, for the most common case a user will only have to fill his/her Source Account and Sequence Number will be filled automatically.

irisli:
Interesting suggestion @bartekn . It might be nice to have it automatically fetched. I think what we want this to do really depends on how we want users learning about Stellar to see/understand the sequence number.

### Ideology 1: Sequence number is important to understand
In this case, the field should be empty by default and the user has to go through the action of fetching the sequence number manually just like they would have to do in the code.

### Ideology 2: Sequence number can be useful but beginners shouldn't worry
In this case, it would be good to have the user manually load the sequence number.

### Ideology 3: Sequence number is something users generally shouldn't worry about
In this case, then the sequence number should be automatically loaded then locked from the user. The way to change it is manually.

--------

Since this is a learning tool, I lean on the side of having the user press the button manually (ideology 1) so that they understand that they will have to do the same thing when writing code. A button press is just a simplified way to fetch the sequence number just like the laboratory as a whole is a tool for more easily accessing Stellar features.

jedmccaleb:
Yeah I think we want option 1

cyberomin:
@irisli, the option of pressing a button to generate the sequence number is the most elegant way to tackle this problem. Thanks @bartek for bringing this up.

irisli:
Thanks for the feedback @cyberomin @jedmccaleb @bartekn!

nullstyle:
LGTM!  great work!

PR #69, opened 2016-02-04
Simple routing

irisli:
This implements routing in the laboratory with light optimizations to reduce the url length.

This doesn't use react-router. Instead, it uses a small router written for just the Laboratory with a simple data flow. Components do not know about a router at all and the only routing information is stored in a one item reducer. I'm overall pretty happy with the way this turned out.

## docs.md
I created a file called `docs.md` to put in documentation about the laboratory code itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,6 @@ npm install
 npm install
 ./node_modules/.bin/gulp build
 ```
+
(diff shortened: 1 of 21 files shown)


nullstyle:
+1.  Noice!

PR #83, opened 2016-02-17
Tree view and XDR viewer

irisli:
## In the Transaction Signer
image

## In the xdr viewer
image

## Decision behind folding
Right now, there is not much data in a xdr blob and it makes sense to leave everything expanded. At most, it will take up two screens (assuming 720 resolution, high browser bar, and 2 ops&signatures). The lack of expanding makes things much clearer and easier to view. Scrolling is a easy task and this way is much easier than requiring the user to guess what is behind a fold.

## Barebones Design
The xdr viewer is very barebones right now. I'll improve upon it later with later iterations.

I'm still not sure whether to keep the border around the tree view. It looks kinda awkward, but right now, we can see what sticks better.


src/components/TransactionSigner.js
```diff
@@ -51,6 +53,20 @@ class TransactionSigner extends React.Component {
           href={txPostLink(result.xdr)}>Submit this transaction to the network</a>;
       }
 
+      let xdrNodes, txDetails;
+      try {
+        xdrNodes = extrapolateFromXdr(result.xdr, 'TransactionEnvelope');
```
nullstyle: This is unused.  Notice one line down you are making the exact same call.

irisli: Thanks

src/components/XdrViewer.js
```diff
@@ -0,0 +1,193 @@
...
+  'TransactionResultExt': 'TransactionResultExt',
+  'TransactionResultPair': 'TransactionResultPair',
+  'TransactionResultResult': 'TransactionResultResult',
+  'TransactionResultSet': 'TransactionResultSet',
+  'TransactionSet': 'TransactionSet',
+  'TrustLineEntry': 'TrustLineEntry',
+  'TrustLineEntryExt': 'TrustLineEntryExt',
+  'TrustLineFlags': 'TrustLineFlags',
+  'Uint64': 'Uint64',
+};
```
bartekn: I changed it recently to: https://github.com/stellar/xdr-viewer/blob/a75bf97/app/scripts/app.js#L3 so it updates itself when new XDR structures are added.

irisli: Great! I'll update to that

src/components/TransactionSigner.js
```diff
@@ -51,6 +53,20 @@ class TransactionSigner extends React.Component {
...
+        let treeView = <TreeView className="TransactionSigner__details__tree" nodes={extrapolateFromXdr(result.xdr, 'TransactionEnvelope')} />
+        txDetails = <div className="so-back TransactionSigner__details">
+          <div className="so-chunk">
+            <p className="TransactionSigner__details__title">Transaction result details</p>
+            {treeView}
+          </div>
+        </div>
+      } catch (e) {
+        // Likely due to invalid secret key. Error messaging is handled in other sections
```
nullstyle: Let's try to avoid this if we can.  Having a blank, silent catch statement can come back to bite you in the ass.  At the very least, you should specifically target an error that you expect to occur and only go silent if the thrown error matches it.

In this case, I actually expect the error to be related to decoding XDR (and not involving a secret key as mentioned in the comment), and so you should re-throw any errors that don't match that.

src/components/TreeView.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+  if (separatorNeeded) {
+    separator = ': ';
+  }
+
+  return <span><strong>{node.type}</strong>{separator}{value}</span>
+}
+
+function convertComplexValue(complex) {
```
nullstyle: Two issues: "complex" is not really a meaningful, and "convert" is IMO best used when you given a destination.  

How about "format" instead of "convert", and "NodeWithType" instead of "complex"?

src/components/XdrViewer.js
```diff
@@ -0,0 +1,193 @@
...
+import SelectPicker from './FormComponents/SelectPicker';
+import extrapolateFromXdr from '../utilities/extrapolateFromXdr';
+import TreeView from './TreeView';
+import validateTxXdr from '../utilities/validateTxXdr';
+import {updateXdrInput, updateXdrType} from '../actions/xdrViewer';
+
+function XdrViewer(props) {
+  let {dispatch, state} = props;
+
+  let validation = validateTxXdr('');
```
nullstyle: Why are you validating an empty string?

src/components/XdrViewer.js
```diff
@@ -0,0 +1,193 @@
...
+  let messageClass = validation.result === 'error' ? 'xdrInput__message__alert' : 'xdrInput__message__success';
+  let message = <p className={messageClass}>{validation.message}</p>
+
+  let xdrNodes, treeView, errorMessage;
+  if (state.input === '') {
+    errorMessage = <p>Enter a base-64 encoded xdr blob to decode.</p>;
+  } else if (state.type === '' || !_.has(xdrTypes, state.type)) {
+    errorMessage = <p>Please select a valid xdr type</p>;
+  } else {
+    try {
```
nullstyle: I don't understand what is going on with this block:  Why is `xdrNodes` being declared using the static type of `'TransactionEnvelope'`, but then on the immediate next line you're parsing the XDR again using `state.type`?  Why aren't you using `xdrNodes` in your tree view?

irisli: It's dead code from copy paste.

irisli:
@bartekn: I need your help with the xdr viewer code (extrapolateFromXdr) that I took from your xdr-viewer project.

The viewer works on stellar-sdk `0.4.0` but it breaks when I upgrade to `0.4.2`. I tried to debug it for an hour with no success. Could you debug it and see what is the problem?

When I used the `debugger;` (in extrapolateFromXdr), the problem (compared to normal operation) was that it went from `buildTreeFromObject` and checked if the xdr `hasChildren`. But then that function returns false (when in a working version, the `hasChildren` returns true). As a result, buildTreeFromObject runs `anchor.value = getValue(object, name);` instead of the correct path which is `parseNormal(anchor, object)`.

bartekn:
@irisli this is probably caused by a bug I fixed recently (lodash registered itself to the global scope from v4.0.0). Simply reinstalling dependencies should fix this.

irisli:
Woot! It worked. I should have tried that in the beginning.

src/components/TransactionSigner.js
```diff
@@ -60,6 +62,17 @@ class TransactionSigner extends React.Component {
         postLink = <a {...postLinkProps}>Submit this transaction to the network</a>;
       }
 
+      let treeView, txDetails;
+      if (result.xdr) { // Only show tree view if xdr is valid and exists
+        let treeView = <TreeView className="TransactionSigner__details__tree" nodes={extrapolateFromXdr(result.xdr, 'TransactionEnvelope')} />
+        let txDetails = <div className="so-back TransactionSigner__details">
+          <div className="so-chunk">
+            <p className="TransactionSigner__details__title">Transaction result details</p>
+            {treeView}
```
nullstyle: `treeView` isn't used anywhere else is it?  Perhaps just inline it into `txDetails` declaration and then also remove the higher-scope let statement on line 65.

src/components/XdrViewer.js
```diff
@@ -0,0 +1,75 @@
...
+
+  let validation = validateBase64(state.input);
+  let messageClass = validation.result === 'error' ? 'xdrInput__message__alert' : 'xdrInput__message__success';
+  let message = <p className={messageClass}>{validation.message}</p>
+
+  let xdrTypeIsValid = _.indexOf(xdrTypes, state.type) >= 0;
+  let treeView, errorMessage;
+  if (state.input === '') {
+    errorMessage = <p>Enter a base-64 encoded xdr blob to decode.</p>;
+  } else if (state.type === '' || !xdrTypeIsValid) {
```
nullstyle: Shouldn't `state.type === ''` imply that `xdrTypeIsValid` should be false?

nullstyle: That is to say, you should move your empty string check onto line 18 along with the other validity checks

irisli: Good points.

I thought that the xdrTypes array had an empty string element. Turns out the `xdrTypes` array does not and that comparison to `''` would be caught by `xdrTypeIsValid`.

I do `xdrTypeIsValid` since the value is simply a get value in the hash url.

nullstyle:
+1

PR #88, opened 2016-02-25
Question mark help

irisli:
This is not meant to be comprehensive. But it does get in the question marks in some places.

### New style
image

### Old style
image


jedmccaleb:
are question marks better than just making the words links?

irisli:
I speculate so because it makes it more clear that it is some sort of "question" thingy. This is as opposed to simply having it as a link which is more vague in it's meaning. I see scenarios such as `do I click on this to make a memo?` or `why did clicking on the endpoint url not bring me to the docs?`.

Also, a link would be useful mostly to people familiar with the laboratory. However, the laboratory is targeted especially towards people new to Stellar (and thus the laboratory). We should make things as clear as possible for new users.

nullstyle:
+1

PR #98, opened 2016-03-03
Streaming endpoints

irisli:
This adds basic support for streaming in the laboratory. It does **not** do anything funky with the UI to keep the user scrolled where new results are.

image

### DIFF
--- a/src/reducers/endpointExplorer.js
+++ b/src/reducers/endpointExplorer.js
@@ -92,10 +92,10 @@ function results(state={id: null, available: false, body: []}, action) {
   }
 
   if (action.type === ERROR_REQUEST) {
-    let errorContent = action.error.status === 0 ?
+    let errorContent = action.errorStatus === 0 ?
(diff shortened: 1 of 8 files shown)


src/reducers/endpointExplorer.js
```diff
@@ -79,15 +75,39 @@ function pendingRequestValues(state={}, action) {
...
-    return Object.assign({}, state, {isLoading: false, response: payload, error});
-  default:
+  if (action.id !== state.id) {
+    // This action has expired as we've moved on to a new request
+    // Or, this is likely an irrelevant action
     return state;
   }
+
+  if (action.type === ERROR_REQUEST) {
+    let errorContent = action.errorStatus === 0 ?
```
nullstyle: IMO you're getting a little fancy with this statement... ternary expressions split onto multiple lines makes me think you should just use a regular old if/else block.

nullstyle:
+1

PR #105, opened 2016-03-05
JSON result syntax highlighting and linking

irisli:
screen shot 2016-03-04 at 5 10 10 pm


bartekn:
Looks good! However it should scroll to endpoint params when a link is clicked. When items in `_links` are clicked can we show results in endpoints explorer instead of a new tab?

irisli:
Yeah, items in _links are supposed to go to the endpoint explorer. That is a feature to be implemented on Monday.

src/utilities/linkBuilder.js
```diff
@@ -26,10 +28,39 @@ export function txPostLink(xdr) {
...
+  let query = serializeStore(SLUG.XDRVIEWER, {
+    xdrViewer: {
+      input: xdr,
+      type,
+    },
+  });
+  return hashBuilder(SLUG.XDRVIEWER, query);
+}
+
+// explorerEndpoint is pretty low level. Do not use directly
```
nullstyle: This should be improved:

-  A functions code comment should explain what it provides or what role it fulfills.
-  Commanding "do not use directly" is not very helpful, instead, offer an alternative or provide instances in which using it would make sense, to help guide the reader.
-  Low level compared to what? Is there a higher level alternative?

src/utilities/linkHighlightedJson.js
```diff
@@ -0,0 +1,83 @@
...
+  });
+  return body.innerHTML;
+}
+
+// Takes in a property highlighted item. If functionality is defined for it's
+// corresponding property (only string type supported) and the content is valid,
+// then we will turn the value node into a link.
+// @param {HTMLElement} propertyToken - The token element of property type (assigned by prism)
+function annotatePropertyToken(propertyToken) {
+  let valueToken = propertyToken.nextElementSibling.nextElementSibling;
```
nullstyle: This is incredibly brittle, are you sure there are no alternatives?

irisli: As far as I can tell, there isn't a way to iterate over react elements (I'm turning a html string into a group of react elements) so I have to do my chances to the highlighted code one of three ways (there may be another that I do not know of though).

1. Edit the HTML string using DOMParser before giving the HTML string to React
2. Edit the HTML string using DOMParser after giving the HTML string to React by using `componentDidUpdate`. This would be more messy since now the React element output is different from what is the final output.
3. Write a PrismJS plugin to add links to JSON code. An example is in the prism-autolinker plugin which does a related thing: https://github.com/PrismJS/prism/blob/gh-pages/plugins/autolinker/prism-autolinker.js

I chose to do option 1 (the current code) instead of writing a PrismJS plugin since I thought it would be simpler. My current implementation takes 45 lines of code that I believe to not be dense  and not too complicated as opposed to the Prism API and it's data structures.

It was the expedient choice over the "right" solution which is to build a prismjs plugin. I would love to write the plugin and even contribute it to `prism` since I think it's a feature that could help many. But for now, I chose the "easy way" by keeping it as a separate layer.

I think the highlighting feature is here to stay, but i'm not sure I could find a better solution considering the tradeoffs.

nullstyle: Parsing the HTML using the DOMParser isn't brittle;  I was specifically referring to how you are looking up `valueToken`.  

Relying on the relative structure of the dom to extract elements is brittle, which is to say that any change in the relationship between `propertyToken`  and `valueToken` will break the above code.  Prism gets updates and adds a new nesting? This breaks.  Prism gets updated and inserts another intervening sibling element? This breaks.

Instead, you should strive to use targets that are not dependent upon the full relationship between two nodes.  Are there any identifying classes on `valueToken` that can be used to find it?  Is there no other way to identify `valueToken` beyond "The second sibling element of `propertyToken"?

At the very least, you should raise a custom error notifying the developer that the expected dom structure is no longer reality so that they don't get a "Cannot read property 'nextSiblingElement' of undefined" error.

irisli: I could look at the `span.token.property` and find it's closest next neighbor that is `span.token.string`. But some times it might not be a string and we would have to do some validations.

In correctly formed JSON, there should always be a value that follows the property. If I wanted to decouple things, I could iterate over next tags/nodes until I find one with text that is not `:` or ` `. This way, the amount of next we go is based on content in tags and not what Prism gives us.

I'm not sure how to prevent breaking changes from Prism. Someone would have to be manually updating Prism because I pasted it into a js file in the laboratory. The in this case of a manual upgrade, someone will see the errors when upgrading.

Is this the right approach?

irisli:
Got a :+1: from @nullstyle

PR #112, opened 2016-03-12
Advanced results 2

irisli:
Whats new?
- Fixed error hiding bug
- Added scrolling to top when link in JSON is clicked (this feature is half useful (scrolling to top of xdr viewer) and half harmful (scrolling to top of endpoint explorer when the interesting setup part is at the bottom) but it will be addressed in a future update)
- Clear old results when something changes (such as page or endpoint parameters change)
- Make horizon hal links clickable and link them back into the laboratory

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "react-redux": "^4.0.0",
     "redux": "^3.0.4",
     "redux-thunk": "^1.0.0",
+    "route-recognizer": "^0.1.9",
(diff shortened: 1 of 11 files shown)


irisli:
Food for thought: If the horizon server was configured to a rogue Horizon server or a MITM carried out on Horizon could potentially create an XSS attack

nullstyle:
+!

PR #116, opened 2016-03-17
Memo case insensitive

bartekn:


irisli:
Tested and works. :+1:

PR #117, opened 2016-03-18
fix unescaping code to match all instances of &amp; and not just the first

irisli:
This fixes a bug where we weren't unescaping enough. Not a security issue.


nullstyle:
+1

Issue #110, opened 2016-03-08
Remove hash symbol in the url when on the landing page


irisli:
On the landing page, the url sometimes shows `https://www.stellar.org/laboratory/#`. The last hash tag can be omitted since it is empty.

This an by done by doing a replaceState when the `routerMiddleware` gets a url with a hash consisting of just the hash sign.

PR #120, opened 2016-03-18
remove hash symbol in the url when on the landing page

irisli:
Closes https://github.com/stellar/laboratory/issues/110


nullstyle:
No code is better than no code, but whatever.  +1

PR #123, opened 2016-03-22
Account creation & friendbot

irisli:
I made it all on one page from the suggestion of @jedmccaleb.

However, I feel that the way it is done in this PR (or just having it on the first place) is a bit weird and unorganized.

screen shot 2016-03-22 at 2 34 28 pm


jedmccaleb:
I thought we were just going to do the generating keys part? It doesn't seem like we need the friend bot  to be part of the lab.

irisli:
Oh, hmm. I think it's more useful with the friendbot in here though.

litnerdy:
I would say that it really helps to have friendbot in the lab, for context. People really don't get what goes into creating an account, and it's the source of a lot of friction currently.

It helps to have it broken down into each step (generate keypair, then fund) so people realize that generating keypair isn't enough to create an account.

I would also suggest having some more textual context around what it means to create an account. This is probably too much text, but:

> Create a test account on the Stellar network. An account consists of a keypair: a public key and secret key. Create your test account now. 
> Save both your public key and secret key somewhere safe, like in a password-protected document or password vault. Stellar.org does not have a record of your secret key. Do not share your secret key with anyone. 
> The Stellar network includes a security feature that requires every account to keep a small balance of the native currency, lumens (XLM), to conduct transactions. The transaction fee is .00001 lumens.
>

irisli:
Currently waiting on discussion about:

1. Whether to make an additional page for these items
2. Whether to include friendbot.

(my votes for both are to yes, to separate them into a separate page and keep the friendbot here)

jedmccaleb:
I still say it makes it overly cluttered but if we are keeping friend bot then yeah I agree it should probably be its own tab.

irisli:
I think the friendbot is useful because then the laboratory is now able to do everything. Instead of having to split tutorials into both the laboratory and the friendbot in the middle of a landing page on the developer site.

It makes explaining the tool to newcomers easier and keeps tools centralized.

irisli:
I have moved the account creator to a separate tab and improved the layout a bit

image

image

mozzadrella:
Stoked to get this deployed so we can test it. :)

src/reducers/accountCreator.js
```diff
@@ -0,0 +1,63 @@
...
+  return state;
+}
+
+const defaultRequestState = {
+  message: '',
+  code: '',
+  status: 'inital',
+}
+function friendbotStatus(state, action) {
+  if (state === undefined) {
```
nullstyle: This seems like a mistake...  If the first time this reducer is run you entirely ignore the action?  Could you explain the design decision's here?

irisli: You mean on line 38? The check for `state === undefined` is for setting the default value. The reducer gets first run before any non-redux-initiation action gets called.

irisli: Oh weird, I could just do
```
function friendbotStatus(state = defaultRequestState, action) {
```

It used to not work for me before

irisli: I've changed it to the newer es6 default variable, which functions essentially the same as `state === undefined`.

irisli: Ah, the default value is not equivalent to
```
if (state === undefined) {
  return defaultObj;
}
```

but rather 
```
if (state === undefined) {
  state = defaultObj;
}
```

nullstyle:
+1

PR #128, opened 2016-03-24
Fix lodash build issues

irisli:
The previous PR broke the build. I confirmed that it was the lodash version causing it. I'm going to revert it for now and hope that this issue with lodash gets automatically in the future.

This downgrades lodash and adds a function that was only available on the upgraded lodash.

----

I have tested this to work locally and on staging.


nullstyle:
+1

PR #132, opened 2016-03-28
improve path payment operation setup

irisli:
Reordering things and making things more easy to understand. I misunderstood how path payments worked when I first created this pane.

image


bartekn:
Clicking `Add new intermediate asset` results in:
```
ManualMultiPicker.js:43 Uncaught TypeError: Cannot read property 'concat' of undefined
```
Did `npm install` and it's the same.

irisli:
Thanks for catching it @bartekn. Not sure how that slipped through me.

It's fixed now.

bartekn:
:+1:

PR #133, opened 2016-03-28
use new custom dropdown in solar

irisli:
image


bartekn:
Awesome! :+1: 

One suggestion: can the default option say: `Select...` or something like this instead of the empty box?

irisli:
Good suggestion. I'll do it in a separate PR.

PR #135, opened 2016-03-30
add intro text to xdr viewer page

irisli:
image

Copy:
> External Data Representation (XDR) is a standardized protocol that the Stellar network uses to encode data.
> 
> The XDR Viewer is a tool that displays contents of a Stellar XDR blob in a human readable format.


bartekn:
:+1:

PR #136, opened 2016-03-31
Current network in url query params

irisli:
**TL;DR;**: Network parameter will always be in the url. Doing otherwise would be a ton more code for a minor aesthetic improvement.

-----

Most of my notes here are more for my own thoughts and understanding of how I got to this solution.

I have cleanly separated each commit so that is easy to review each step of the way.

## Commit 1: Feature
Added the serialization and deserialization of network into the url.

## Commit 2: Refactoring

When first writing the laboratory, we were not sure what was best to put in the reducer. After a bit more experience, it is more clear that we shouldn't be storing static data in the reducer since it clutters up the reducers and can be done in a cleaner way (by having a constants file).

The presence of a potentially dynamic available networks can mess up how we serialize and deserialize the network part of the state. I do not foresee having an option in the laboratory to add new horizon servers any time soon and this is a good tradeoff to make for safety.

## Commit 3: Realizing that things can't be "pretty"
I originally wanted the URL to not have network to have the lack of network refer to the default network which is test. Unfortunately for my desires, this does not easily work.

The original method would mean that whenever we build urls, we would need to have information about the current network (and thus the whole state).

Having the whole state to build URLs is annoying because we build them in stateless places. For example, we build a URL to link from the transaction builder to the signer. Adding my original vision would mean that every link now needs knowledge of the current network.

## Final notes
I actually like how this turned out and I think it's actually better to explicitly have the network in the url. Things fell together pretty well :)

### DIFF
--- a/src/constants/network.js
+++ b/src/constants/network.js
@@ -1,0 +1,14 @@
+const NETWORK = {
+  available: {
+    test: {
+      url: 'https://horizon-testnet.stellar.org',
+      useNetworkFunc: 'useTestNetwork',
+    },
+    public: {
+      url: 'https://horizon.stellar.org',
+      useNetworkFunc: 'usePublicNetwork',
+    }
+  },
+  defaultName: 'test',
+};
+export default NETWORK;
--- a/src/constants/network.js
+++ b/src/constants/network.js
@@ -1,12 +1,14 @@
+import {Network, Networks} from 'stellar-sdk';
+
 const NETWORK = {
   available: {
     test: {
       url: 'https://horizon-testnet.stellar.org',
-      useNetworkFunc: 'useTestNetwork',
+      networkObj: new Network(Networks.TESTNET),
     },
     public: {
       url: 'https://horizon.stellar.org',
-      useNetworkFunc: 'usePublicNetwork',
+      networkObj: new Network(Networks.PUBLIC),
     }
   },
   defaultName: 'test',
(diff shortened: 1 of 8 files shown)


bartekn:
Wondering if we should add a small notice to parts of Laboratory that rely on the network (Explorer, Signer, Builder > Sequence Number) when switching between networks. Example: User loads the account, realizes the network is wrong, switches the network, horizon response is still "not found", show notice box: `Network has been changed. You need to resubmit your request` (or something like this).

irisli:
Good idea on the endpoint explorer. I will just clear the results when network changes

src/constants/network.js
```diff
@@ -0,0 +1,14 @@
+const NETWORK = {
+  available: {
+    test: {
+      url: 'https://horizon-testnet.stellar.org',
+      useNetworkFunc: 'useTestNetwork',
```
nullstyle: Not sure when this got introduced... by I want to flag it as undesirable.  The 'useXNetwork' functions are helpers for simple scripts... calling to those methods by name  based upon data is needlessly complex and brittle.  Instead... use the `Network.use` method and pass it an instance of `Network`.

irisli: Relevant code in js-stellar-base: https://github.com/stellar/js-stellar-base/blob/master/src/network.js

So would it be better to have the `useNetworkFunc: 'useTestNetwork',` instead be `stellarBaseNetworkKey: 'PUBLIC',` (or maybe a shorter name)?

nullstyle:
Other than my comment about the 'useXNetwork' functions, +1.  I'm happy to talk about how to fix that one issue in person if my comments weren't clear.

nullstyle:
+1

PR #137, opened 2016-03-31
Select placeholder

irisli:
image

I also wrote some code in solar to support having the placeholder item be greyed out. I purposely let the placeholder item to NOT be disabled because it lets users go back to the default state.


bartekn:
:+1:

PR #142, opened 2016-04-06
add button to view in xdr viewer

irisli:
image

Note the new button at the bottom and the reworded label


bartekn:
They should probably scroll to top when clicked. Other than that: 👍

Issue #129, opened 2016-03-26
should strip white space


jedmccaleb:
https://www.stellar.org/laboratory/#explorer?resource=accounts&endpoint=single&values=eyJhY2NvdW50X2lkIjoiR0E1WElHQTVDN1FUUFRXWFFIWTZNQ0pSTVRSWkRPU0hSNkVGSUJORFFUQ1FIRzI2Mk40R0dLVE0gIn0%3D

PR #143, opened 2016-04-07
trim whitespace from endpoint explorer input

irisli:
Fixes https://github.com/stellar/laboratory/issues/129

Spaces are still enterable but just won't affect the built url


nullstyle:
+1

PR #145, opened 2016-04-07
Fix incorrect txsigner persistence

irisli:
Previously, the txsigner state would:
1. clear the txsigner state due to any LOAD_STATE triggered by changing the url (directly or via a link)
2. remain the secret key input that didn't get cleared

---

Also includes a refactoring to simplify the redux reducer (net negative lines of code; woot!)


nullstyle:
+1

PR #147, opened 2016-04-12
add button to clear transaction builder form

irisli:
image

--- 
Unrelated musing: I sometimes wonder if I am adding too many features that are unhelpful.

### DIFF
--- a/src/actions/transactionBuilder.js
+++ b/src/actions/transactionBuilder.js
@@ -1,6 +1,14 @@
 import axios from 'axios';
 import {UnsignedHyper} from 'stellar-sdk';
 
+// Resets everything to it's default state
+export const RESET_TXBUILDER = 'RESET_TXBUILDER';
+export function resetTxbuilder() {
+  return {
+    type: RESET_TXBUILDER,
+  }
+}
+
 // Attributes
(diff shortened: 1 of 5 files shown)


jedmccaleb:
Yeah this one is pushing it :) they can always refresh

irisli:
Seems superfluous, but it is actually quite difficult to clear the state from a user's perspective. 

1. Refresh doesn't get rid of anything (by design since is meant to keep state and refresh doesn't change the url)
2. Clicking to another tab and then clicking back doesn't get rid of things (by design since redux state persists)
3. Deleting things from the url doesn't get rid of things because the routing is designed to fill in the state when the query items don't exist (see note below)

(note about point 3): These "pitfalls" are by design since it is not simple to distinguish between a url change caused by a link or directly via the address bar (though doable, but adds complexity ... complexity not unknown to the laboratory).

---

Perhaps we could add analytics to see which buttons and events get triggered the most. Though we must be careful to not accidentally send any sensitive information. People might get antsy about tracking in the laboratory too. So maybe a way to disable the tracking? (but then EU laws would be in arms about cookies lol ... but we aren't in the EU).

---
These notes are as much for others as they are for myself.

nullstyle:
+1

PR #150, opened 2016-04-14
change "account ID" to "public key"

litnerdy:
For internal consistency with docs. @irisli is this okay?

### DIFF
--- a/src/components/AccountCreator.js
+++ b/src/components/AccountCreator.js
@@ -15,7 +15,7 @@ class AccountCreator extends React.Component {
     if (state.keypairGeneratorResult !== null) {
       keypairTable = <div className="simpleTable AccountCreator__generator__table">
         <div className="simpleTable__row">
-          <div className="simpleTable__row__label">Account ID</div>
+          <div className="simpleTable__row__label">Public Key</div>
           <div className="simpleTable__row__content">{state.keypairGeneratorResult.pubKey}</div>
         </div>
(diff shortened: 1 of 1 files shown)


irisli:
Technically, the correct term is public key and I originally did that. But then I switched to "Account ID" because it's the more relevant term. For the same reason I labelled the page as "Account Creator" instead of "Keypair Generator (with friendbot)" since 95% of the uses of this page will be for accounts.

Though with the public key vs account ID, I'm kinda wavering. Public key is the more broadly recognized term but Account ID is the more relevant term.

litnerdy:
I hear ya. I don't mind "Account creator" at the top - that's correct, because a keypair = an account. But account id will be confusing, since we refer to it in the docs (and everywhere else) as a public key. I strongly vote for changing it. @jessica-collier ?

jessica-collier:
Account ID and public key are often synonymous, as I understand them, but then there are cases when they're not. I think that we need to make it clearer what each term means and surface the appropriate one when necessary. In this instance, I'm not sure that I understand the logic at play. @irisli, how is Account ID more relevant (because it refers explicitly to an "account"?) even though public key is technically correct and more recognizable, and why is "relevance" the better option?

irisli:
I sat "relevance" because it is more related to what it will be used for.

However, after thinking about it, I believe that it might be better to change it to public key so that it is easily recognizable.

PR #156, opened 2016-04-21
Implement basic unit testing

irisli:
Adds:
- mocha and chai for unit tests
- integration with travis
- a bit of code improvements and bug fixes as a result of writing these tests


src/utilities/hydration.js
```diff
@@ -2,8 +2,9 @@ export const rehydrate = function(obj) {
   try {
     return JSON.parse(new Buffer(obj, 'base64').toString());
   } catch (e) {
-    alert('Unable to parse values passed in url query parameters');
-    console.error(e);
+    if (typeof alert !== 'undefined') {
```
nullstyle: Add an else block to this statement where you default to using `console.error`.  Code that fails silently is no bueno ;)

test/unit/Libify.js
```diff
@@ -0,0 +1,308 @@
...
+    it('contains no memo', () => {
+      expect(decodedTx.memo._switch.name).to.equal('memoNone');
+    });
+
+    it('contains one operation', () => {
+      expect(decodedTx.operations.length).to.equal(1);
+    });
+  })
+
+  describe('large transaction', () => {
```
nullstyle: This can be better as a call to `context`, not `describe`.  While the two functions are aliases, they have an important distinction for communicating intent to other coders:  `describe` is used to specify the thing that is being tested (in this case `Libify.buildTransaction`) , context is used to specify the context in which you test is operating (i.e. the input or environment).

test/unit/Libify.js
```diff
@@ -0,0 +1,308 @@
+import React from 'react';
+import {xdr, Transaction, Operation} from 'stellar-sdk';
+import Libify from '../../src/utilities/Libify';
+
+describe('Libify-built', () => {
```
nullstyle: IMO you should change the name of this to reflect the function being tested: `buildTransaction`

test/unit/Libify.js
```diff
@@ -0,0 +1,308 @@
...
+          value: 'such test'
+        }
+      }
+    ]);
+    let decodedTx = xdr.TransactionEnvelope.fromXDR(largeTransaction.xdr, 'base64');
+    let opAtIndex = (index) => {
+      return Operation.operationToObject(decodedTx._attributes.tx._attributes.operations[index]);
+    };
+
+    describe('1st operation', () => {
```
nullstyle: Using ordinals here for your description is a bit cumbersome... I would prefer you either use the operation name, the index of the operation, or add comments in the structure above that specifies the ordinal of each operation.

test/unit/routingRoundTrip.js
```diff
@@ -0,0 +1,126 @@
+import _ from 'lodash';
```
nullstyle: Should this file actually be in the `test/unit` subfolder?  It doesn't seem like a unit test to me.

nullstyle:
The only blocker I have to merging is the silent failure vulnerability.  The test suite issues are less meaningful to me.

src/utilities/hydration.js
irisli: The thing that prompted me to remove the console.error was that it would show up in the tests. This is also an error that can be caused by users and I would like the laboratory to recover after the user pastes corrupt data into the URL.

How should I proceed?

PR #161, opened 2016-07-14
Remove "All Accounts" endpoint

nullstyle:
This endpoint has been removed in an upcoming horizon release, and is already removed on testnet.


irisli:
Theres also a line from the link parser to delete:
https://github.com/stellar/laboratory/blob/master/src/utilities/horizonUrlParser.js#L23

Other than that link: :+1: from me

nullstyle:
Ah, thanks for the heads up.  I'll fix it momentarily

PR #186, opened 2017-11-23
Initial ledger wallet support

nullstyle:
This PR adds the most basic ledger wallet support...  adding a single signature for the default account in a wallet.

Next steps: add support for custom paths.


src/actions/transactionSigner.js
```diff
@@ -18,5 +23,51 @@ export function setSecrets(secrets) {
...
+        type: LEDGER_WALLET_SIGN_ERROR,
+        error: err,
+       }); 
+    };
+
+    let onConnect = () => {
+      console.log(Promise)
+      BP.all([
+        ledgerApi.getPublicKey_async(DEFAULT_BIP),
+        ledgerApi.signTx_async(DEFAULT_BIP, transaction),
```
bartekn: This will sign only txs with one operation. We need to use `signTxHash_async` to sign any transaction. This should be probably handled in `stellar-ledger-api` (if one operation => `signTx_async` else `signTxHash_async`).

Also because of this we should display transaction hash in the transaction summary table.

bartekn: Created this issue: https://github.com/lenondupe/stellar-ledger-api/issues/3

nullstyle: will fix later

bartekn: @lenondupe released a new version of `stellar-ledger-api`. It's now OK to use this function with every transaction.

package.json
```diff
@@ -21,19 +27,20 @@
...
     "redux": "^3.0.4",
     "redux-thunk": "^1.0.0",
     "route-recognizer": "^0.1.9",
-    "sass-loader": "^3.0.0",
+    "sass-loader": "^3.2.3",
     "solar-css": "git+https://github.com/stellar/solar#master",
     "solar-stellarorg": "git+https://github.com/stellar/solar-stellarorg#master",
-    "stellar-sdk": "0.7.3",
-    "urijs": "^1.16.0",
+    "stellar-ledger-api": "^1.1.0",
```
bartekn: Update to `1.1.1`.

bartekn:
@nullstyle added some changes to error reporting and upgraded `ledger-stellar-api`. Also removed `package-lock.json`. `npm` has been unreliable to me recently, it's super slow and doesn't seem to be able to download some packages even though I'm using the latest stable Node and `npm`. Please rebuild `package-lock.json` if you want but I'd be happy if we started using `yarn` which is really fast and I had no problems using it so far.

bartekn:
When it comes to tests, it looks like it's connected to https://github.com/node-hid/node-hid/issues/226. and the fix will be released soon so I think we can merge this and create a PR to `ledgerco` package when `node-hid` is released.

nullstyle:
@bartekn yeah, lets move to yarn.   I'll look into that when I work on the lab next.

PR #189, opened 2017-12-27
remove old orderbook trades endpoint / add new trades endpoint

tomerweller:


src/components/SetupPanes/Trades.js
```diff
@@ -8,32 +8,37 @@ import OrderPicker from '../FormComponents/OrderPicker.js';
 
 export default function Trades(props) {
   return <div>
-    <OptionsTablePair label="Selling Asset">
+    <OptionsTablePair label="Base Asset" optional={true}>
       <AssetPicker
-      value={props.values['selling_asset']}
-      onUpdate={(value) => {props.onUpdate('selling_asset', value)}}
+      value={props.values['base_asset']}
+      onUpdate={(value) => {props.onUpdate('base_asset', value)}}
```
bartekn: `AssetPicker` acts like a radio button so it's impossible to "uncheck" it. The only option is to select another endpoint and then go back to Trades.

It's not in a scope of this PR but let's create an issue for "Unchecking" optional `AssetPicker`.

tomerweller: Will do.

PR #326, opened 2018-04-29
Add xdr type search field

mlsteele:
Add a field for searching through the numerous XDR types. The top result from the fuzzy search is set as the active type, and the other results (up to 10) are put before the `---` in the type list.

The experience if you don't use the search field is unchanged.

Before typing
image

After typing
image

### DIFF
--- a/src/actions/xdrViewer.js
+++ b/src/actions/xdrViewer.js
@@ -8,6 +8,14 @@ export function updateXdrInput(input) {
   }
 }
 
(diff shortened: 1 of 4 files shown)


vcarl:
Hey @mlsteele! Thanks for adding this, sorry that it's languished for so long. I'm going to close this out for the moment, but it's a good idea and we may make it happen at a later date.

PR #338, opened 2018-07-13
Use custom horizon/network

### LINKED ISSUE #173: Show current network in transaction builder near tx hash

### LINKED ISSUE #176: Allow custom Horizon URL on Stellar Laboratory
Could be a really nice tool for customers to test / explore their local setup with some kind of UI.

bartekn:
Allows setting current Horizon/Network passphrase to custom values:

screen shot 2018-07-13 at 18 50 32

It required changing how `state.network` looks like and the value is checked in many places in the application. I think I updated all the usages but please run the code on your local machine and try to break it (and check look&feel) when doing a review.

Close #176.
Close #173.

### DIFF
--- a/src/actions/network.js
+++ b/src/actions/network.js
@@ -1,15 +1,46 @@
-export const CHOOSE_NETWORK = "CHOOSE_NETWORK";
-export function chooseNetwork(name) {
-  return {
-    type: CHOOSE_NETWORK,
-    name
-  }
-}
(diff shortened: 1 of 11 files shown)


robertDurst:
Got it up and running locally with my own horizon instance on `http://localhost:8000` with testnet passphrase 👍 

screen shot 2018-07-13 at 5 39 02 pm

PR #342, opened 2018-07-23
Trades for Offer

robertDurst:
Started to work on this: https://github.com/stellar/go/issues/525 and realized this endpoint was not yet added here.

### DIFF
--- a/src/components/SetupPanes/ForOffer.js
+++ b/src/components/SetupPanes/ForOffer.js
@@ -1,14 +1,15 @@
 import React from 'react';
 
 import For from './For';
-import OfferPicker from '../FormComponents/OfferPicker.js';
+import PositiveIntPicker from '../FormComponents/PositiveIntPicker.js';
 
-export default function ForAccount(props) {
+export default function ForOffer(props) {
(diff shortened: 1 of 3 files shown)


src/components/SetupPanes/ForOffer.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import For from './For';
+import OfferPicker from '../FormComponents/OfferPicker.js';
+
+export default function ForAccount(props) {
```
bartekn: `ForAccount` -> `ForOffer`

src/components/SetupPanes/ForOffer.js
```diff
@@ -0,0 +1,14 @@
...
+  let label = 'Offer ID';
+  let content = <OfferPicker
+    value={props.values['offer_id']}
+    onUpdate={(value) => {props.onUpdate('offer_id', value)}}
```
bartekn: Why not use `PositiveIntPicker` directly? `OfferPicker` doesn't seem to be doing anything special (other than setting a default placeholder).

PR #345, opened 2018-08-01
Remove authorization immutable flag from clear options

### LINKED ISSUE #320: Transaction builder allows clear immutable flag
Which will either be a no-op if immutable flag is false, or a failure if immutable is true.

screen shot 2018-02-14 at 10 02 20 am

robertDurst:
Under no circumstance would/can you ever clear an Authorization Immutable flag.

However, this is not quite as aesthetic looking:

screen shot 2018-08-01 at 1 58 36 pm

Closes #320


robertDurst:
Closes #320

bartekn:
FYI, Close/Closed/Fix etc auto-close tags doesn't work in comments. Updated the main PR description to automatically close #320.

Issue #336, opened 2018-06-29
some hex fields are displayed as base64


MonsieurNicolas:
Found that when viewing certain objects in the XDR Viewer, like `LedgerHeader` (as returned by the `ledgers` endpoint), hash values are base64 encoded (in other places in API etc they are hex encoded):

For example, [this ledger header](https://www.stellar.org/laboratory/#xdr-viewer?input=AAAACVAjytC4kHIYS7km%2BoZ2YZA3UtFQp39Oh7wbLJzn6FYe3Aj4XAYO2rYcNJZpL3pnSrJt1rLazpFCmihcQX42uL8AAAAAWzV5pwAAAAAAAAAAGtdqljXXrf33cr15ZVe9zP23tceUrsnbIt4jjfHpqcJxOzqyjwxZUgTi3%2BEJR4G9u2cCTjd30J%2BezkyqoqwkuAEcmyIOcSZHno%2B3jQAADOkaFo8TAAAA0QAAAAAA%2FgLnAAAAZABMS0AAAAAygDBgwG0KFL5niZMjteoE3%2BTB4rkAg8mGyP%2FpMCfY0viKLceI1w2UW98v%2FDTqlm2uJZVneo%2F4Nre9C0HONLNoc6t8aESrBPWEmTs%2BxbucJWEZ3eoqLyz4zBtGcY4cYvW3kuQw2R%2BMfAuB3ViYKmEXRaY15kNlAJf77PWXPJcRv3MAAAAA&type=LedgerHeader&network=public) returns
```json
previousLedgerHash: UCPK0LiQchhLuSb6hnZhkDdS0VCnf06HvBssnOfoVh4=
```

bartekn:
I'm wondering if we should switch to hex everywhere or create a switch (base64/hex) in XDR viewer section. I think the only places we use base64 are:
* [Account Data Entries](https://horizon-testnet.stellar.org/accounts/GACH7QS2F6DGG7T77B2PJBJU4YGKZ5RGMSUNJAV4SBSZGSULTGYS565I),
* Memo Hash, serialized XDRs and signatures in [transaction](https://horizon-testnet.stellar.org/transactions/509340b64d4b86649cefedeccf83a61e866fe1768b0ea265308fcf66d44f874a).

Everything else could be hex.

MonsieurNicolas:
Now that I think of it, the problem is that we don't want to give the wrong data for developers:
* humans (like me) when debugging want something consistent with core/horizon's database
[...]

PR #348, opened 2018-08-06
Add signature verification to xdr viewer for transaction envelopes

### LINKED ISSUE #343: xdr viewer: add bad signature indicator
When displaying a `TransactionEnvelope` with signatures attached it would be great to have laboratory indicate if a signature is wrong.

robertDurst:
This may need some work making it proper React code. It also might need some touchup on design. I kept it pretty simple.

Basically, whenever the user types or pastes in an xdr of type `Transaction Envelope`, this will asynchronously query for all signers on that transaction (base source account and all operation source accounts). It will then, for each signature on the transaction check to see if it is valid for a given signer -- the red means invalid, green means valid, and during loading, black means not yet loaded.

Here is what it looks like for a transaction has two valid signers and one invalid signer:
screen shot 2018-08-06 at 4 28 44 pm

Closes #343

### DIFF
--- a/src/actions/xdrViewer.js
+++ b/src/actions/xdrViewer.js
@@ -1,4 +1,6 @@
+import {xdr, hash, StrKey, Network, Keypair} from 'stellar-sdk';
 import axios from 'axios';
+import SIGNATURE from '../constants/signature';
 
 export const UPDATE_XDR_INPUT = 'UPDATE_XDR_INPUT';
 export function updateXdrInput(input) {
@@ -17,13 +19,100 @@ export function updateXdrType(xdrType) {
 }
 
 export const FETCH_LATEST_TX = 'FETCH_LATEST_TX';
-export function fetchLatestTx(horizonBaseUrl) {
+export function fetchLatestTx(horizonBaseUrl, networkPassphrase) {
   return dispatch => {
     axios.get(horizonBaseUrl + '/transactions?limit=1&order=desc')
       .then(r => {
-        dispatch(updateXdrInput(r.data._embedded.records[0].envelope_xdr))
+        const xdr = r.data._embedded.records[0].envelope_xdr;
+        dispatch(updateXdrInput(xdr))
         dispatch(updateXdrType('TransactionEnvelope'))
+        dispatch(fetchSigners(xdr, horizonBaseUrl, networkPassphrase))
       })
       .catch(r => dispatch({type: FETCH_SEQUENCE_FAIL, payload: r}))
   }
+}
+
+export const FETCHED_SIGNERS_SUCCESS = 'FETCHED_SIGNERS_SUCCESS';
+export const FETCHED_SIGNERS_FAIL = 'FETCHED_SIGNERS_FAIL';
+export const FETCHED_SIGNERS_START = 'FETCHED_SIGNERS_START';
+export function fetchSigners(input, horizonBaseUrl, networkPassphrase) {
+  return dispatch => {
+    dispatch({ type: FETCHED_SIGNERS_START });
+    try {
+      // Capture network for determining signature base
+      StellarSdk.Network.use(new Network(networkPassphrase));
+
+      let tx = new StellarSdk.Transaction(input);
+      const hashedSignatureBase = hash(tx.signatureBase());
+
+      // Extract all signatures on transaction
+      const signatures = tx.signatures.map(x => ({ sig: x.signature() }));
+
+      // Extract all source accounts from transaction (base transaction and all operations)
+      let sourceAccounts = {};
+  
+      let baseAccountId = tx.source;
+      sourceAccounts[baseAccountId] = true;
+      
+      tx.operations.forEach(op => {
+        if (op.source) {
+          sourceAccounts[op.source] = true;
+        }
+      });
+
+      // Get all signers per source account - array of promises
+      sourceAccounts = Object.keys(sourceAccounts).map(accountID => axios.get(horizonBaseUrl + '/accounts/' + accountID));
+
+      Promise.all(sourceAccounts)
+      .then(response => {
+        let allSigners = {};
+        response.forEach(r => r.data.signers.forEach(signer => allSigners[signer.key] = signer));
+
+        allSigners = Object.values(allSigners);
+        
+        // We are only interested in checking if each of the signatures can be verified for some valid
+        // signer for any of the source accounts in the transaction -- we are not taking into account
+        // weights, or even if this signer makes sense.
+        for (var i = 0; i < signatures.length; i ++) {
+          const sigObj = signatures[i];
+          let isValid = false;
+
+          for (var j = 0; j < allSigners.length; j ++) {
+            const signer = allSigners[j];
+            
+            // By nature of pre-authorized transaction, we won't ever receive a pre-auth
+            // tx hash in signatures array, so we can ignore pre-authorized transactions here.
+            switch (signer.type) {
+              case 'sha256_hash':
+                const hashXSigner = StrKey.decodeSha256Hash(signer.key);
+                const hashXSignature = hash(sigObj.sig);
+                isValid = hashXSigner.equals(hashXSignature);
+                break;
+              case 'ed25519_public_key':
+                const keypair = Keypair.fromPublicKey(signer.key);
+                isValid = keypair.verify(hashedSignatureBase, sigObj.sig);
+                break;
+            }
+
+            if (isValid) {
+              break;
+            }
+          }
+
+          sigObj.isValid = isValid ? SIGNATURE.VALID : SIGNATURE.INVALID;
+        }
+        dispatch({
+          type: FETCHED_SIGNERS_SUCCESS,
+          result: signatures,
+        });
+      })
+      .catch(e => {
+        console.error(e);
+        dispatch({ type: FETCHED_SIGNERS_FAIL })
+      });
+    } catch(e) {
+      console.error(e);
+      dispatch({ type: FETCHED_SIGNERS_FAIL });
+    }
+  };
 }
--- a/src/components/TreeView.js
+++ b/src/components/TreeView.js
@@ -4,26 +4,29 @@
 import React from 'react';
 import _ from 'lodash';
 import {EasySelect} from './EasySelect';
+import SIGNATURE from '../constants/signature';
 
 // @param {array} props.nodes - Array of TreeView compatible nodes
 export default class TreeView extends React.Component {
   render() {
-    let {nodes, className} = this.props;
+    let {nodes, className, fetchedSigners, parent} = this.props;
     let rootClass = 'TreeView ' + (className) ? className : '';
 
     let result = <div className={rootClass}>
       {_.map(Array.prototype.slice.call(nodes), (node, index) => {
         let childNodes;
 
+        let position = getPosition(node, parent);
+        
         if (typeof node.nodes !== 'undefined') {
           childNodes = <div className="TreeView__child">
-            <TreeView nodes={node.nodes} />
+           <TreeView nodes={node.nodes} fetchedSigners={fetchedSigners} parent={position} />
           </div>;
         }
 
         return <div className="TreeView__set" key={index}>
           <div className="TreeView__row" key={index + node.type}>
-            <RowValue node={node} />
+            <RowValue node={node} position={position} fetchedSigners={fetchedSigners} />
           </div>
           {childNodes}
         </div>
@@ -35,8 +38,8 @@ export default class TreeView extends React.Component {
 }
 
 function RowValue(props) {
-  let value, childNodes, separatorNeeded, separator;
-  let {node} = props;
+  let value, separatorNeeded, separator;
+  let {node, position, fetchedSigners} = props;
 
   if (typeof node.value === 'string') {
     value = String(node.value);
@@ -56,7 +59,41 @@ function RowValue(props) {
     separator = ': ';
   }
 
+  if (position === 'TransactionEnvelope.signatures') {
+    checkSignatures(node, fetchedSigners);
+    return formatSignatureCheckState(node, separator)
+  }
+
+  if (position.match(/^TransactionEnvelope\.signatures\[[0-9]+\]\.signature$/)) {
+    return formatSignature(node, separator);
+  }
+
   return <span><strong>{node.type}</strong>{separator}{value}</span>
+}
+
+function checkSignatures(signatures, fetchedSigners) {
+  // catch fetch signature errors
+  if (fetchedSigners === null || fetchedSigners === "PENDING") {
+    signatures.state = SIGNATURE.NOT_VERIFIED_YET;
+    return
+  } else if (fetchedSigners === "ERROR") {
+    signatures.state = SIGNATURE.INVALID;
+    return
+  }
+  
+  signatures.state = SIGNATURE.VALID;
+
+  for (var i = 0; i < signatures.nodes.length; i ++) {
+    const sig = signatures.nodes[i].nodes.find(n => n.type == 'signature');
+    const fetchedSignature = fetchedSigners.find(x => x.sig.equals(sig.value.raw));
+    
+    let isValid = SIGNATURE.NOT_VERIFIED_YET;
+    if (fetchedSignature) {
+      isValid = fetchedSignature.isValid;
+    }
+
+    sig.value.isValid = isValid;
+  }
 }
 
 // Types values are values that will be displayed with special formatting to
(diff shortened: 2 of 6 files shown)


robertDurst:
Fails for hash(x)

robertDurst:
Now passes for pre-auth tx and hash x signers, should be good for review @bartekn

src/actions/xdrViewer.js
```diff
@@ -27,3 +29,151 @@ export function fetchLatestTx(horizonBaseUrl) {
...
+      // Get all signers per source account - array of promises
+      srcAccounts = Object.keys(srcAccounts).map(x => getSigner(horizonBaseUrl, x));
+
+      Promise.all(srcAccounts)
+      .then(response => {
+        const allSigners = response.reduce((obj, resp) => {
+          resp.data.signers.forEach(signer => {
+            // Get key hint for matching
+            const key = signer.key.substring(signer.key.length-9, signer.key.length-4);
+            if (!obj.hasOwnProperty(key)) {
```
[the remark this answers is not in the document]
robertDurst: I believe stellar core uses this for optimization -- some trade off between likely-hood of collision vs space saved.

src/actions/xdrViewer.js
```diff
@@ -27,3 +29,151 @@ export function fetchLatestTx(horizonBaseUrl) {
...
+    // Used for matching against ExtrapolateXDR's signature formatting
+    const value = new Buffer(sig).toString('base64');
+
+    // In case of hashX
+    const hashx = sig.toString();
+
+    // Utilize same strategy as ExtrapolateXDR file, reference there for how this works
+    const hintBytes =  new Buffer(x._attributes.hint.toString("base64"), "base64");
+    const partialPublicKey = Buffer.concat([new Buffer(28).fill(0), hintBytes]);
+    const hint = StrKey.encodeEd25519PublicKey(partialPublicKey, 'base64').substr(47, 5);
```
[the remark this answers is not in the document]
robertDurst: Decided to ignore the hint. In my discussion with Tomer last week, we don't need to utilize the hint optimization for this.

src/actions/xdrViewer.js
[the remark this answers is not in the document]
robertDurst: Similar to below, decided to ignore the hint.

robertDurst:
@bartekn thanks for the feedback. Utilizing Stellar-base and removing trivial hint optimization lead to much cleaner code.

Also squashed.

src/actions/xdrViewer.js
```diff
@@ -27,3 +29,127 @@ export function fetchLatestTx(horizonBaseUrl) {
...
+}
+
+function extractSignatures(tx) {
+  return tx.signatures.map(x => {
+    const signature =  x._attributes.signature;
+    return {
+      // In case of hashX
+      hashx: hash(signature.toString()),
+      // Used for matching against ExtrapolateXDR's signature formatting
+      value: new Buffer(signature).toString('base64'),
```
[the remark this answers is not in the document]
robertDurst: ah yes, nicolas opened an issue about that

src/actions/xdrViewer.js
```diff
@@ -27,3 +29,127 @@ export function fetchLatestTx(horizonBaseUrl) {
...
+                isValid = signer.value.toString() === sigObj.hashx.toString();
+                break;
+              case 'ed25519_public_key':
+                isValid = signer.keypair.verify(hashedSignatureBase, sigObj.sig);
+                break;
+            }
+
+            if (isValid) {
+              break;
+            }
```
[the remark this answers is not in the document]
robertDurst: Not sure this is invalid, at least the way the code is currently written. Here the check on `isValid` exits the loop once a valid key is found. Not only does this optimize (skipping a few rounds of checks once a valid key is seen) it also breaks since `isValid` is re-written during each pass of the internally nested loop.

robertDurst:
Fixed per your latest comments... you were right, hash(x) still failed under the previous commit. It should be fixed now with buffer/hex/bin sorted out a bit more.

src/actions/xdrViewer.js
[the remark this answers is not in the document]
bartekn: You are right, my mistake.

src/components/TreeView.js
```diff
@@ -4,20 +4,26 @@
...
 
 // @param {array} props.nodes - Array of TreeView compatible nodes
 export default class TreeView extends React.Component {
   render() {
-    let {nodes, className} = this.props;
+    let {nodes, className, fetchedSigners} = this.props;
     let rootClass = 'TreeView ' + (className) ? className : '';
 
+    // First check in case of transaction with no signatures, i.e. pre-auth
+    if(nodes[0] && nodes[0].type == "hint" && fetchedSigners) {
```
bartekn: I think this condition is a little complicated. `fetchedSigners` is present if the type is a `TransactionEnvelope` (checked in `state.type === "TransactionEnvelope"` condition). So we really want to check if we are in `TransactionEnvelope` view but we're doing it by checking the other variable set somewhere else.

Also I believe that in a very near future we'll want to fix https://github.com/stellar/laboratory/issues/336 and it will be much easier to do if we know the exact location we're currently at in a tree.

So I propose to construct a location string. We always have key name in `nodes[i].type`. So we can construct a location string like `TransactionEnvelope.signatures[i]` by appending the current key name in each child and trigger an action when we are when we want to be. For location with a variable part ( arrays like `TransactionEnvelope.signatures[0]`, `TransactionEnvelope.signatures[1]`...) we can use `RegExp`.

src/components/TreeView.js
```diff
@@ -34,6 +40,18 @@ export default class TreeView extends React.Component {
...
 }
 
+function checkSignature(nodes, fetchedSigners) {
+  const signatureBuffer = Buffer.from(nodes[1].value.value, 'base64');
+  const sig = fetchedSigners.filter(x => x.sig.equals(signatureBuffer))[0];
+  
+  let isValid = SIGNATURE.INVALID;
+  if (sig) {
+    isValid = sig.isValid;
+  }
```
bartekn: `isValid` defaults to `SIGNATURE.INVALID` so in case of errors (ex. connection error) the signature will be displayed as invalid (you can go to Chrome Dev tools, switch to Offline mode and load a `TransactionEnvelope` to reproduce it).

What do you think something like this:
* When checking the signatures (loading signers etc) we add "Checking signatures..." to `TransactionEnvelope.signatures` node:
screen shot 2018-08-14 at 15 43 50
* If there was an error we display it at the `TransactionEnvelope.signatures` level:
screen shot 2018-08-14 at 15 45 28
* If checking was successful we display the statuses (like we do now):
screen shot 2018-08-14 at 15 45 52

src/components/TreeView.js
```diff
@@ -34,6 +40,18 @@ export default class TreeView extends React.Component {
   }
 }
 
+function checkSignature(nodes, fetchedSigners) {
+  const signatureBuffer = Buffer.from(nodes[1].value.value, 'base64');
+  const sig = fetchedSigners.filter(x => x.sig.equals(signatureBuffer))[0];
```
bartekn: Better to use [`find`](https://lodash.com/docs/4.17.10#find) instead of `filter` here.

src/components/XdrViewer.js
```diff
@@ -39,13 +44,18 @@ function XdrViewer(props) {
           <p>The XDR Viewer is a tool that displays contents of a Stellar XDR blob in a human readable format.</p>
         </div>
         <p className="XdrViewer__label">
-          Input a base-64 encoded XDR blob, or <a onClick={() => dispatch(fetchLatestTx(baseURL))}>fetch the latest transaction to try it out</a>:
+        Input a base-64 encoded XDR blob, or <a onClick={() => dispatch(fetchLatestTx(baseURL))}>fetch the latest transaction to try it out</a>:
```
bartekn: If you click this link it doesn't trigger signers check.

src/actions/xdrViewer.js
```diff
@@ -27,3 +29,82 @@ export function fetchLatestTx(horizonBaseUrl) {
...
+  return dispatch => {
+    try {
+      // Capture network for determining signature base
+      StellarSdk.Network.use(new Network(networkPassphrase));
+
+      let tx = new StellarSdk.Transaction(input);
+      const hashedSignatureBase = hash(tx.signatureBase());
+
+      // Extract all signatures on transaction
+      const signatures = tx.signatures.map(x => ({ sig: x._attributes.signature }));
```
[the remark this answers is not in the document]
robertDurst: It is x.signature()... much nicer 😄

src/components/TreeView.js
robertDurst: Looks good! So basically a sort of visual of the state of fetched signatures:
* loading
* failed
* success (same as now)

src/components/TreeView.js
robertDurst: Not quite following here -- can discuss tomorrow AM (UTC -7).

I see the key name, but where does the location string live? Is it attached to the node? I know that the TreeView has some recursion, so does this key persist through iterations?

robertDurst:
Fixed for presentation level feedback. However, left a reply on a comment where I was a bit confused.

src/components/TreeView.js
bartekn: We need to build a location ourselves. `TreeView` renders itself recursively so if we send a current (then parent) name to the child we can construct the full location string. Check the quick patch below:

```diff
diff --git a/src/components/TreeView.js b/src/components/TreeView.js
index e7ed97b..8a0e4d4 100644
--- a/src/components/TreeView.js
+++ b/src/components/TreeView.js
@@ -9,7 +9,7 @@ import SIGNATURE from '../constants/signature';
 // @param {array} props.nodes - Array of TreeView compatible nodes
 export default class TreeView extends React.Component {
   render() {
-    let {nodes, className, fetchedSigners} = this.props;
+    let {nodes, className, fetchedSigners, parent} = this.props;
     let rootClass = 'TreeView ' + (className) ? className : '';
 
     if(nodes[1] && nodes[1].type == "signatures") {
@@ -20,15 +20,22 @@ export default class TreeView extends React.Component {
       {_.map(Array.prototype.slice.call(nodes), (node, index) => {
         let childNodes;
 
+        let sep = '.';
+        if (!parent || node.type.charAt(0) == '[') {
+          sep = '';
+        }
+
+        let position = (parent ? parent+sep : '') + node.type;
+
         if (typeof node.nodes !== 'undefined') {
           childNodes = <div className="TreeView__child">
-            <TreeView nodes={node.nodes} fetchedSigners={fetchedSigners}/>
+            <TreeView nodes={node.nodes} fetchedSigners={fetchedSigners} parent={position} />
           </div>;
         }
 
         return <div className="TreeView__set" key={index}>
           <div className="TreeView__row" key={index + node.type}>
-            <RowValue node={node} />
+            <RowValue node={node} position={position} />
           </div>
           {childNodes}
         </div>
@@ -67,7 +74,7 @@ function checkSignatures(nodes, fetchedSigners) {
 
 function RowValue(props) {
   let value, childNodes, separatorNeeded, separator;
-  let {node} = props;
+  let {node, position} = props;
 
   if (typeof node.value === 'string') {
     value = String(node.value);
@@ -95,7 +102,7 @@ function RowValue(props) {
     return formatSignature(node, separator);
   }
 
-  return <span><strong>{node.type}</strong>{separator}{value}</span>
+  return <span><strong>{node.type} {position}</strong>{separator}{value}</span>
 }
 
 // Types values are values that will be displayed with special formatting to
```
Then instead of:
```js
if(nodes[1] && nodes[1].type == "signatures")
```
We can check if we are in the exact location in a tree:
```js
if (position == 'TransactionEnvelope.signatures')
```

src/components/TreeView.js
```diff
@@ -34,8 +48,34 @@ export default class TreeView extends React.Component {
   }
 }
 
+function checkSignatures(nodes, fetchedSigners) {
+  // catch fetch signature errors
+  if (fetchedSigners === null) {
+    nodes[1].state = SIGNATURE.NOT_VERIFIED_YET;
```
bartekn: I think something is wrong here. I set connection latency to 10s in devtools. I expected to see "Checking signatures..." but instead I got "Error checking signatures..." and after 10s it disappeared and updated each signature state.

src/components/TreeView.js
```diff
@@ -4,26 +4,40 @@
...
 
+        let sep = '.';
+        if (!parent || node.type.charAt(0) == '[') {
+          sep = '';
+        }
+ 
+        let position = (parent ? parent + sep : '') + node.type;
+
+        if (position == 'TransactionEnvelope.signatures') {
+          checkSignatures(nodes, fetchedSigners);
```
bartekn: Use `node` here so that you don't need to use `nodes[1]` inside `checkSignatures`..

src/components/TreeView.js
```diff
@@ -56,6 +96,14 @@ function RowValue(props) {
     separator = ': ';
   }
 
+  if (node.type === 'signatures') {
```
bartekn: Use `position` prop here.

src/components/TreeView.js
```diff
@@ -56,6 +96,14 @@ function RowValue(props) {
...
+    return formatSignatureCheckState(node, separator)
+  }
+
+  if (node.type === "signature") {
```
bartekn: Use `position` prop here too but use regexp to check if you are in `TransactionEnvelope.signatures[x].signature`.

src/components/TreeView.js
robertDurst: Ah I see. The error is that I have an initial state of null with only two states, fail and success. However, there is in fact also a third state (different from initial state) for loading.

src/components/TreeView.js
```diff
@@ -4,26 +4,41 @@
...
     let result = <div className={rootClass}>
       {_.map(Array.prototype.slice.call(nodes), (node, index) => {
         let childNodes;
 
+        let sep = '.';
+        if (!parent || node.type.charAt(0) == '[') {
+          sep = '';
+        }
+ 
+        let position = (parent ? parent + sep : '') + node.type;
```
bartekn: Let's move it to a separate function.

src/reducers/xdrViewer.js
```diff
@@ -37,3 +37,15 @@ function type(state = 'TransactionEnvelope', action) {
 
   return state;
 }
+
+function fetchedSigners(state = "uninitialized", action) {
```
bartekn: We can use `undefined` here (or even remove value as it will be `undefined` by default?). Or maybe add:
```js
export const FETCHED_SIGNERS_PENDING = 'FETCHED_SIGNERS_PENDING'; // or a better name?
```

src/components/TreeView.js
```diff
@@ -71,3 +120,30 @@ function convertTypedValue({type, value}) {
...
+// signature is valid for one of the signers related to the transaction
+// envelope.
+function formatSignature(node, separator) {
+  let style = {color: "black"};
+  if (node.value.isValid === SIGNATURE.INVALID) {
+    style = {color: "red"};
+  } else if (node.value.isValid === SIGNATURE.VALID) {
+    style = {color: "green"};
+  }
+  return <span style={style}><strong>{node.type}</strong>{separator}{node.value.value}</span>
```
bartekn: Let's add add a small icon here too, like ❌ / ✅ (can be unicode).

src/constants/signature.js
```diff
@@ -0,0 +1,9 @@
+// This standardizes all the slug strings into a simple map
+// These are used in to build the url and are the "page name"
```
bartekn: This comment seems wrong.

src/components/TreeView.js
```diff
@@ -56,6 +97,14 @@ function RowValue(props) {
     separator = ': ';
   }
 
+  if (position === 'TransactionEnvelope.signatures') {
+    return formatSignatureCheckState(node, separator)
+  }
+
+  if (position.match(/(\.signature)+$/)) {
```
bartekn: We really need a full [regexp](https://regex101.com/r/m4xxKw/1) here. No idea how `TransactionEnvelope` will looks like in a future and it will break if there's another `signature` field.

src/reducers/xdrViewer.js
robertDurst: Originally that was undefined, specifically an empty string, however in js land ```"" == []``` which failed for this:

```
// Fetch signers on initial load
  if (state.type === "TransactionEnvelope" && state.fetchedSigners === 'uninitialized') {
    dispatch(fetchSigners(state.input, baseURL, networkPassphrase))
  }
```

since initial state == case with no signers ([])

robertDurst: Also, default state cannot be undefined:
```
combineReducers.js:56 Uncaught Error: Reducer "fetchedSigners" returned undefined during initialization. If the state passed to the reducer is undefined, you must explicitly return the initial state. The initial state may not be undefined. If you don't want to set a value for this reducer, you can use null instead of undefined.
```

robertDurst: Decided to go with:

INITIAL -->     null
SUCCESS --> result (type array)
FAILURE -->      "ERROR"
PENDING -->     "PENDING"

src/actions/xdrViewer.js
```diff
@@ -17,13 +19,96 @@ export function updateXdrType(xdrType) {
...
+          sigObj.isValid = isValid ? SIGNATURE.VALID : SIGNATURE.INVALID;
+        }
+        dispatch({
+          type: FETCHED_SIGNERS_SUCCESS,
+          result: signatures,
+        });
+      })
+      .catch(err => dispatch({ type: FETCHED_SIGNERS_FAIL }));
+    } catch(error) {
+      dispatch({ type: FETCHED_SIGNERS_FAIL });
```
bartekn: Please add `console.error` in both `catch` functions.

src/components/TreeView.js
```diff
@@ -56,9 +59,44 @@ function RowValue(props) {
...
+  }
+
+  if (position.match(/^TransactionEnvelope\.signatures\[[0-9]+\]\.signature$/)) {
+    return formatSignature({node, separator});
+  }
+
   return <span><strong>{node.type}</strong>{separator}{value}</span>
 }
 
+function checkSignatures({signatures, fetchedSigners}) {
```
bartekn: Any reason for destructuring here and other methods? I think it's just copied from `convertTypedValue({type, value})` definition where it's helpful because you can pass a node value there: `convertTypedValue(node.value)`.

src/components/TreeView.js
```diff
@@ -56,9 +59,44 @@ function RowValue(props) {
...
+    return
+  } else if (fetchedSigners === "ERROR") {
+    signatures.state = SIGNATURE.INVALID;
+    return
+  }
+  
+  signatures.state = SIGNATURE.VALID;
+
+  for (var i = 0; i < signatures.nodes.length; i ++) {
+    const sig = signatures.nodes[i].nodes[1];
```
bartekn: This can be confusing if you don't remember that `signatures.nodes` are [decorated signatures](https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x#L10). Let's change it to:
```js
const sig = signatures.nodes[i].find(n => n.type == 'signature');
```

src/components/TreeView.js
```diff
@@ -56,9 +59,44 @@ function RowValue(props) {
...
+    const signatureBuffer = Buffer.from(sig.value.value, 'base64');
```
bartekn: As a part of a separate PR we should probably move all the extra presentation functions here from `extrapolateFromXdr`. Right now we should probably leave the original value in a resulting object like:
```js
  if (object && object._isBuffer) {
    return {type: 'code', raw: object, value: new Buffer(object).toString('base64')};
  }
```
Without it we need to convert it back to bytes from base64 representation. Then we can do:
```js
const fetchedSignature = fetchedSigners.find(x => x.sig.equals(sig.raw));
```

robertDurst:
@bartekn ok updated for latest comments. Also squashed.

PR #352, opened 2018-08-15
Fix setOptions unselected errors

### LINKED ISSUE #351: Clearing master weight or thresholds in set options sets it to 0 instead of undefined
**Repro steps:**

1. Go to https://www.stellar.org/laboratory/#txbuilder?params=eyJhdHRyaWJ1dGVzIjp7InNvdXJjZUFjY291bnQiOiJHQ1ZIRUtTUkFTSkJENk8yWjUzMkxXSDROMlpMQ0JWRExMVExLU1lDU01CTE9ZVE5NRUVHVUFSRCIsInNlcXVlbmNlIjoiNzIxMDQ4MDg2MTM0MTI4NjcifSwib3BlcmF0aW9ucyI6W3siaWQiOjAsImF0dHJpYnV0ZXMiOnsiaW5mbGF0aW9uRGVzdCI6IkdDQ0Q2QUpPWVpDVUFRTFgzMlpKRjJNS0ZGQVVKNTNQVkNGUUkzUkhXS0wzVjQ3UVlFMkJOQVVUIn0sIm5hbWUiOiJzZXRPcHRpb25zIn1dfQ%3D%3D&network=test
2. Click on View in XDR Viewer and observe that master weight is undefined (will not be set with this transaction)
3. Go back to the transaction builder and enter "1" for the Master Weight in the Set Options operation
4. Delete the "1" you just entered
5. Go back to the View in XDR Viewer

**Expected:**

Should have undefined in the master weight, since it is not explicitly set in the UI

**Actual:**

Sets the master weight to 0

Note: this also happens for any of the threshold fields

**Why this is bad:**

If you accidentally type in a value into master weight and then delete it, you may accidentally set your master weight to 0, locking yourself out of your account.

I actually had a support ticket on StellarGuard where someone was trying to manually configure the StellarGuard signer and ended up locking their own private key out (luckily thresholds was also set to 0 due to this bug and I was able to reset the master weight via the StellarGuard signer).

robertDurst:
Fixes two issues:

1. When unselecting master weight, or thresholds, it defaulted to zero instead of undefined (#351 )
2. When unselecting flags, it defaulted to zero instead of undefined

Closes #351

### DIFF
--- a/src/components/FormComponents/FlagfieldPicker.js
+++ b/src/components/FormComponents/FlagfieldPicker.js
@@ -64,7 +64,7 @@ function intToExistenceArray(inputInt) {
 function existenceArrayToInt(existenceArray) {
   return _.reduce(existenceArray, (result, intValue) => {
     return result | parseInt(intValue);
-  }, 0);
+  }, undefined);
 }
 
 function isInteger(input) {
--- a/src/utilities/Libify.js
+++ b/src/utilities/Libify.js
@@ -49,7 +49,7 @@ let castIntOrUndefined = function(value) {
   if (typeof value === 'number') {
     return value;
   }
-  if (_.isString(value) && value.match(/^[0-9]*$/g)) {
+  if (value !== '' && _.isString(value) && value.match(/^[0-9]*$/g)) {
     return Number(value);
   }
   return undefined;


bartekn:
```js
if (value !== '' && _.isString(value) && value.match(/^[0-9]*$/g))
```
is equal to:
```js
if (_.isString(value) && value.match(/^[0-9]+$/g))
```

I changed it in: https://github.com/stellar/laboratory/commit/4295ca8b963c91ad2a47c2df659f5edfac3e0dd0

PR #363, opened 2019-01-11
Improve signature check

### LINKED ISSUE #358: Signature check in XDR viewer fails when account does not exist
https://www.stellar.org/laboratory/#xdr-viewer?input=AAAAAMrYKG3BZaSqRhOGYDLzjmftV4T0cygxdm0dnFd%2Bqo%2BcAAAB9ACny6sAAAALAAAAAAAAAAAAAAAFAAAAAAAAAAAAAAAAD6UEWPiw4VpT2FflRjWjEglquS%2BgcUXKJmWuvgbDCt4AAAAAAcnDgAAAAAEAAAAAD6UEWPiw4VpT2FflRjWjEglquS%2BgcUXKJmWuvgbDCt4AAAAGAAAAAVhMTVgAAAAApikH8CkLwXHSqsBEU8s%2BjSVgfFithvueHV%2BWmsq3bi8AAFrzEHpAAAAAAAEAAAAAD6UEWPiw4VpT2FflRjWjEglquS%2BgcUXKJmWuvgbDCt4AAAAGAAAAAUFJWAAAAAAApikH8CkLwXHSqsBEU8s%2BjSVgfFithvueHV%2BWmsq3bi8AAFrzEHpAAAAAAAEAAAAAD6UEWPiw4VpT2FflRjWjEglquS%2BgcUXKJmWuvgbDCt4AAAAFAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAASZWGWA7maQgp9E2v2eQbvkIGj0BAQeAgIgOwgH1ydLAAAAAQAAAAEAAAAAD6UEWPiw4VpT2FflRjWjEglquS%2BgcUXKJmWuvgbDCt4AAAAFAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAgAAAAEAAAACAAAAAAAAAAAAAAAAAAAAAgbDCt4AAABA3I66cqPuV1tIdre5v%2Bo4zpmW%2F7Ja0QiV4knmfV5FSaI1K4kDHytzU%2FjMiCiLIeh8XwQssvK3%2BU7JOJNraLFdAX6qj5wAAABAAzMEwbLvq3Oxqp%2BKY%2BhlZWDhxqBL2LPIb39cB9%2BCv9gq5teBOGMgg8Ahg336qiQni6B5ZRwD5gchRj5jT2D2CQ%3D%3D&type=TransactionEnvelope&network=test

bartekn:
This commit improves signature checking code by adding FETCHED_SIGNERS
state names and changing `fetchedSigners` reducer to an object.

It also adds a new state indicating that one of the source accounts of
the transaction does not exist.

Close #358.

### DIFF
--- a/src/actions/xdrViewer.js
+++ b/src/actions/xdrViewer.js
@@ -1,6 +1,7 @@
 import {xdr, hash, StrKey, Network, Keypair} from 'stellar-sdk';
 import axios from 'axios';
 import SIGNATURE from '../constants/signature';
+import FETCHED_SIGNERS from '../constants/fetched_signers';
 
 export const UPDATE_XDR_INPUT = 'UPDATE_XDR_INPUT';
 export function updateXdrInput(input) {
@@ -32,12 +33,9 @@ export function fetchLatestTx(horizonBaseUrl, networkPassphrase) {
   }
 }
 
-export const FETCHED_SIGNERS_SUCCESS = 'FETCHED_SIGNERS_SUCCESS';
-export const FETCHED_SIGNERS_FAIL = 'FETCHED_SIGNERS_FAIL';
-export const FETCHED_SIGNERS_START = 'FETCHED_SIGNERS_START';
 export function fetchSigners(input, horizonBaseUrl, networkPassphrase) {
   return dispatch => {
-    dispatch({ type: FETCHED_SIGNERS_START });
+    dispatch({ type: FETCHED_SIGNERS.PENDING });
     try {
       // Capture network for determining signature base
       StellarSdk.Network.use(new Network(networkPassphrase));
@@ -102,17 +100,21 @@ export function fetchSigners(input, horizonBaseUrl, networkPassphrase) {
           sigObj.isValid = isValid ? SIGNATURE.VALID : SIGNATURE.INVALID;
         }
         dispatch({
-          type: FETCHED_SIGNERS_SUCCESS,
+          type: FETCHED_SIGNERS.SUCCESS,
           result: signatures,
         });
       })
       .catch(e => {
         console.error(e);
-        dispatch({ type: FETCHED_SIGNERS_FAIL })
+        if (e.status == 404) {
+          dispatch({ type: FETCHED_SIGNERS.NOT_EXIST });
+        } else {
+          dispatch({ type: FETCHED_SIGNERS.FAIL });
+        }
       });
     } catch(e) {
       console.error(e);
-      dispatch({ type: FETCHED_SIGNERS_FAIL });
+      dispatch({ type: FETCHED_SIGNERS.FAIL });
     }
   };
 }
--- a/src/components/TreeView.js
+++ b/src/components/TreeView.js
@@ -5,6 +5,7 @@ import React from 'react';
 import _ from 'lodash';
 import {EasySelect} from './EasySelect';
 import SIGNATURE from '../constants/signature';
+import FETCHED_SIGNERS from '../constants/fetched_signers';
 
 // @param {array} props.nodes - Array of TreeView compatible nodes
 export default class TreeView extends React.Component {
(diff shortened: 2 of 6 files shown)


src/actions/xdrViewer.js
```diff
@@ -102,17 +100,21 @@ export function fetchSigners(input, horizonBaseUrl, networkPassphrase) {
...
+        if (e.status == 404) {
+          dispatch({ type: FETCHED_SIGNERS.NOT_EXIST });
+        } else {
+          dispatch({ type: FETCHED_SIGNERS.FAIL });
+        }
       });
     } catch(e) {
       console.error(e);
-      dispatch({ type: FETCHED_SIGNERS_FAIL });
+      dispatch({ type: FETCHED_SIGNERS.FAIL });
```
poliha: I am wondering if we might be missing something here by considering only signers that are fetched successfully.  This is because a signer can still be valid even if the account is not active on the network. @bartekn thoughts?

poliha: When the signer does not exist, can we still verify the signature?

bartekn: Please note that we are only loading source accounts that must exist for transaction to succeed. It's true that you can use signer that does not have corresponding account in the ledger.

src/components/TreeView.js
```diff
@@ -84,19 +85,15 @@ function RowValue(props) {
 
 function checkSignatures(signatures, fetchedSigners) {
   // catch fetch signature errors
-  if (fetchedSigners === null || fetchedSigners === "PENDING") {
-    signatures.state = SIGNATURE.NOT_VERIFIED_YET;
-    return
-  } else if (fetchedSigners === "ERROR") {
-    signatures.state = SIGNATURE.INVALID;
-    return
+  signatures.state = fetchedSigners.state;
```
morleyzhi: What's the purpose of this line? I don't see that `signatures` is referenced elsewhere.

I really try to avoid mutating data this way in React, especially in functions called in `render` because of how frequently (and unpredictably) they get run. You could end up in a situation where data gets changed multiple times, and it'd be really hard to figure out what's causing the data to change. Better to do handle all data mutations in the reducer.

src/components/TreeView.js
```diff
@@ -150,15 +147,24 @@ function formatSignature(node, separator) {
...
 // richer formating with ternary based coloring based on whether the
-// state is valid (VALID), is pending (NOT_VERIFIED_YET), or has errored (INVALID).
+// state is NONE, SUCCCESS, PENDING, FAIL or NOT_EXIST.
 function formatSignatureCheckState(node, separator) {
   let message = "";
-  if (node.state === SIGNATURE.INVALID) {
-    message = <span style={{color: "red"}}> Error checking signatures...</span>;
-  } else if (node.state === SIGNATURE.NOT_VERIFIED_YET) {
-    message = <span style={{fontStyle: "italic"}}> Checking signatures...</span>;
+  switch (node.state) {
```
morleyzhi: You don't have to change this, but just a suggestion for the future. `state` is a term of art in React, so I avoid using that as the name. It might have been clearer to name this prop 'signatureCheckState' or something.

src/components/TreeView.js
bartekn: `signatures` is a reference to `node` (see call in [line 64](https://github.com/stellar/laboratory/pull/363/files/9f8e8e232e8baeb503eeb3830b52214d8143282c#diff-3fb8d04a71b32353f0ef387d0baf89d3R64)). It's not actually changing the state in React terminology, it's probably a name confusion: https://github.com/stellar/laboratory/pull/363#discussion_r253111030. We use `state` property too often, let's fix this in a separate PR.

src/components/TreeView.js
bartekn: https://github.com/stellar/laboratory/issues/367

PR #370, opened 2019-02-26
Add amplitude metrics

### LINKED ISSUE #369: Add analytics
We aren't collecting any usage metrics, which makes it hard to know how well the laboratory fulfills users' needs. Google Analytics would be a simple way to college page-level metrics, which is a good start.

vcarl:
We don't know very much about how people use the laboratory. It would be valuable to know what features are used so we can make improvements.

`utilities/metrics.js` exports `addEventHandler`, `broadcastEvent`, and `logEvent`. Each feature can add its own even handler, which gets called for each dispatched action. Since the app is already set up such that all the actions a user can take are redux actions, this lets us tap into it and `logEvent`s to Amplitude.

Resolves #369 

* [x] accountCreator
* [x] endpointExplorer
* [x] network
* [x] routing
* [x] transactionBuilder
* [x] transactionSigner
* [x] xdrViewer
* [x] Add Amplitude key to build environment


src/utilities/metrics.js
```diff
@@ -0,0 +1,43 @@
...
+  // In development, track when events are fired for debugging purposes.
+  if (process.env.NODE_ENV === 'development') {
+    console.log(`[METRICS]: "${type}", ${JSON.stringify(properties)}`)
+  }
+  instance.logEvent(type, properties)
+}
+
+const eventHandlers = []
+
+export const addEventHandler = handler => {
```
morleyzhi: Anon functions don't seem better than just exporting named functions.

src/components/XdrViewer.js
```diff
@@ -18,15 +27,17 @@ function XdrViewer(props) {
...
     errorMessage = <p>Enter a base-64 encoded XDR blob to decode.</p>;
   } else if (!xdrTypeIsValid) {
     errorMessage = <p>Please select a XDR type</p>;
   } else {
     try {
       treeView = <TreeView nodes={extrapolateFromXdr(state.input, state.type)} fetchedSigners={state.fetchedSigners} />
+      tLogEvent(metricsEvents.decodeSuccess, {type: state.type})
     } catch (e) {
       console.error(e)
+      tLogEvent(metricsEvents.decodeFailed, {type: state.type})
```
vcarl: This and [L37](https://github.com/stellar/laboratory/pull/370/files#diff-ca1739756f192224e8ee7af3c7dfce76R37) are debounced metrics loggers run within the render function of this component, which isn't ideal. Pretty much the only useful metric to collect here, though, is whether the XDR was successfully decoded, and that doesn't happen in response to a Redux action.

morleyzhi: Seems like this should then use `componentDidUpdate` instead?

vcarl: That'd be a much more significant refactor, since this is a function component and everything is calculated there. Would have to add additional information as well to tell whether it succeeded or failed, too.

PR #384, opened 2019-05-31
Include the babel-polyfill

tyvdh:
Unable to run in development due to regeneratorRuntime errors. The babel-polyfill fixes this issue and you already have it in the package.json. You just need to include it in the project.

Screen Shot 2019-05-31 at 9 25 18 AM

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,5 +1,7 @@
 require("./styles/main.scss");
 
+import 'babel-polyfill';
+
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {createStore, applyMiddleware} from 'redux';


bartekn:
Are you sure you try it on master? I don't see such error in development.

tyvdh:
Yep on master. Guy in Keybase originally mentioned the issue so it's not isolated to me.

Screen Shot 2019-05-31 at 9 52 17 AM

bartekn:
Strange, I'm using exactly the same Chrome version (74.0.3729.169) and I don't see it.

tyvdh:
Fresh install on alternate machine? `babel-polyfill` could be a global package or coming from an extension? Maybe try on a couple different browsers?

bartekn:
It seems that the error comes from `@ledgerhq/hw-transport-u2f` package. Can you ensure you are using version `4.15.0`? Have you used `yarn` or `npm` to fetch dependencies? What's the result for `git log -1` for you?

tyvdh:
I'm on `"version": "4.60.3"` according to my node_modules.

```
$ git log -1
commit a0da1ce51ed0ca4840168f3fd3b8170f27c124dd (HEAD -> master, tyvdh/master)
Author: Tyler van der Hoeven <hi@tyvdh.com>
Date:   Fri May 31 09:30:12 2019 -0400

    include the babel-polyfill
```

tyvdh:
You're not committing the lock file for yarn or npm so it's possible your node_modules are out of date to what everyone else is installing. Might want to try opening a fresh project from a git clone.

bartekn:
> I'm on "version": "4.60.3" according to my node_modules.

Yeah, that's the issue. You should be at `4.15.0` ([yarn.lock](https://github.com/stellar/laboratory/blob/master/yarn.lock#L11)).

> You're not committing the lock file for yarn.

We do commit `yarn.lock`. There's an issue in README file that says to use `npm` instead of `yarn`.

tyvdh:
Wait I lied yarn.lock is committed. Let me try using yarn. Your README suggests npm which doesn't have the lockfile committed.

bartekn:
Just pushed https://github.com/stellar/laboratory/pull/385. Let me know if `yarn` works for you.

tyvdh:
Bingo. If you use yarn it works. Still. You're using an outdated version and will need to address in the future. Ultimately for this PR looks like just fixing the docs to prefer/enforce yarn rather than npm should do the trick.

PR #389, opened 2019-08-02
Reskin

vcarl:
Screen Shot 2019-08-02 at 5 51 57 PM
Screen Shot 2019-08-02 at 5 52 02 PM


src/components/LaboratoryChrome.js
```diff
@@ -28,6 +34,33 @@ function LaboratoryChrome(props) {
   }
 
   return <div>
+    <style dangerouslySetInnerHTML={{__html: `
```
msfeldstein: Is it possible to put these in a css file and just import that?

src/styles/_TreeView.scss
```diff
@@ -2,10 +2,10 @@
 
 }
   .TreeView__child {
-    border-left: 1em solid $s-color-primary9;
+    border-left: 1em solid #F2F2F2;
```
msfeldstein: Should this be in a var?

src/components/LaboratoryChrome.js
vcarl: There isn't a mechanism for hosting static assets that I could find besides importing them, and I'm not clear how to do so from a non-JS file.

msfeldstein: Does this transform it into base64? I was able to import SVGs into my CSS files as base64 dataurls using css-loader, svg-loader, and praying to the gods of webkit: https://github.com/stellar/sep6-demo-client/blob/master/webpack.config.js

what you have now certainly doesn't bother me though.

msfeldstein: https://webpack.js.org/loaders/css-loader/
`The css-loader interprets @import and url() like import/require() and will resolve them.`

src/styles/_TreeView.scss
vcarl: I did it as a one-off because this is the only use of this particular gray, because the palette gray was slightly too dark due.

src/components/LaboratoryChrome.js
vcarl: Ooo good tip. Cheers, I've been deep in styled-components for a long time.

PR #392, opened 2019-08-16
Fix redirect when missing trailing /

vcarl:


vcarl:
Closing in favor of #393

PR #395, opened 2019-08-26
Design polish, set testnet date to automatically refresh

vcarl:
* Fetch testnet reset date from statuspage
* Soften text color, shrink to 16px
* Fix ? size when font changes


bartekn:
@vcarl Sorry for a post-merge comment but I think fetching testnet reset date will not always work. I believe in the future we will have scheduled maintenances for other components (Horizon, Stellar-Core or network upgrades). I think it's actually a good idea to copy the full code from the Dashboard, including current incident. Ex. if there are problems with Horizon, Lab users should now about it because you can query Horizon endpoints in "Endpoint Explorer".

PR #410, opened 2019-10-09
create initial Dockerfile

brahman81:


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ubuntu:16.04 as build
```
vcarl: Just curious, what's the motivation for not using a node base image for this?

jacekn:
I created #427 which obsoletes this PR.

PR #412, opened 2019-10-14
remove 'Fund this account on the test ...' from public

### LINKED ISSUE #380: Friendbot should be disabled when non test networks are selected
as friendbot is only active on the test network

jeesunikim:
The text `Fund this account on the test network using the friendbot tool below` to only show on TESTNET

Fixes #380


vcarl:
I'd like to get #413 merged, rebase this off master, and then make sure the preview looks good before merging

PR #418, opened 2019-10-28
Fix testnet reset banner information

### LINKED ISSUE #411: Test Network reset date is invalid
Most likely because of: https://github.com/stellar/laboratory/pull/395#issuecomment-525033282

vcarl:
This was incorrectly assuming that scheduled maintenance would always be a testnet reset, which was wrong for the protocol 12 vote. Fixes #411

### DIFF
--- a/src/components/TestnetBanner.js
+++ b/src/components/TestnetBanner.js
@@ -1,56 +1,71 @@
-import React from 'react'
-
-const formatDate = date => `${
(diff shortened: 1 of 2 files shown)


vcarl:
This is still imperfect: 
* it will show _all_ scheduled maintenance, which means we can't schedule multiple future resets without showing multiple banners
* it only displays when the test network is selected, which means we're not appropriately broadcasting main network changes (like the protocol 12 vote)
I think this is a good step though.

PR #423, opened 2019-11-11
Enable CSP

vcarl:


vcarl:
This has to be done in headers, so closing this in favor of an ops ticket.

PR #425, opened 2019-12-09
Add strict send and strict receive to paths to endpoint explorer.

abuiles:
Extend endpoint explorer with [strict receive](https://www.stellar.org/developers/horizon/reference/endpoints/path-finding-strict-receive.html) and [strict send](https://www.stellar.org/developers/horizon/reference/endpoints/path-finding-strict-send.html) endpoints. 

This PR adds a new widget to receive one or more assets using the canonical representation of an asset which is `native` for XLM or `AssetCode:AssetIssuer` for issued assets. 

## Strict Receive

Screen Shot 2019-12-09 at 5 02 51 PM

## Strict Send

Screen Shot 2019-12-09 at 5 03 45 PM


abuiles:
@vcarl thanks for formatting :)

vcarl:
Hm, I'm not sure we should have 2 different widgets for adding an asset. It seems confusing to me to have a form with 2 different methods of inputting the same information. 
Screen Shot 2019-12-10 at 10 13 25 AM
Destination asset and source assets are both looking for more or less the same data.

Do we specifically want to give people the option to input `BTC:<address>`/`native` strings?

abuiles:
@vcarl  Destination asset uses a new format to specify an asset info, which is simpler.  

The old format would have required you to specify an asset passing 3 query params (`asset_type`, `asset_code` and `asset_issuer`) which is abstracted by the `AssetPicker` component,  the new canonical representation simplifies it to `native` for XLM or `AssetCode:AssetIssuer` for issued assets. Whether it is `alphanum_4` or `alphanum_12` is chosen based on the length. 

There are some scenarios where the API has both formats :( and that's why we kept the old widget for selecting an asset.  Alternatively, we could  change `CanonicalAssetTextPicker` to render something similar to `AssetPicker`, adapting to the new representation. We could also allow it to show a "+" or "Add" button, which can be used for end-points that take a list of assets (they are split by commas).

vcarl:
Yeah I think it would be good to have `CanonicalAssetTextPicker` appear similarly to `AssetPicker`—maybe what format to pass to `onUpdate` could be a flag on `AssetPicker`? Since these are both "assets" from a user perspective it seems like they should have the same UI for picking them.

abuiles:
We could simplify it since the later doesn't require the asset type, or we
can keep it like that .

vcarl:
New multi-asset picker:
Screen Shot 2020-02-21 at 3 18 38 PM

src/components/FormComponents/AssetPicker.js
```diff
@@ -1,87 +1,164 @@
...
+      <RadioButtonPicker
+        value={isNative ? "native" : "issued"}
+        onUpdate={() => {
+          onUpdate(isNative ? ":" : "native");
+        }}
+        className={!isNative ? "picker--spaceBottom" : ""}
+        items={stringFormRadioOptions}
+      />
+      {!isNative &&
+        (() => {
```
vcarl: Inline [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE). A little awkward, but since the `[code, issuer] = value.split(':')` only works if `!isNative`, this seemed a lot less awkward than the alternative. Handy way to conditionally declare variables within the return.

src/components/SetupPanes/FindStrictReceivePaymentPaths.js
```diff
@@ -0,0 +1,53 @@
...
+      </OptionsTablePair>
+      <OptionsTablePair label="Source Assets">
+        <AssetMultiPicker
+          stringForm
+          value={
+            values["source_assets"] === undefined
+              ? []
+              : // Filter out empty strings from resulting array.
+                // ''.split(',') produces ['']
+                values["source_assets"].split(",").filter(Boolean)
```
vcarl: I kinda don't like how this is, but I don't particularly want redux to have to care about munging the data it receives, and I also don't want `AssetMultiPicker` to emit a single comma-separated list. Least-bad option, though the `.filter(Boolean)` makes it a lot more bad.

src/components/SetupPanes/FindStrictSendPaymentPaths.js
```diff
@@ -0,0 +1,54 @@
...
+      </OptionsTablePair>
+      <OptionsTablePair label="Destination Assets">
+        <AssetMultiPicker
+          stringForm
+          value={
+            values["destination_assets"] === undefined
+              ? []
+              : // Filter out empty strings from resulting array.
+                // ''.split(',') produces ['']
+                values["destination_assets"].split(",").filter(Boolean)
```
vcarl: (see [comment](https://github.com/stellar/laboratory/pull/425/files#r382790521))

src/index.ejs
```diff
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta http-equiv="Content-Security-Policy-Report-Only" content="default-src 'self'; font-src data: 'self'; connect-src 'self' api.amplitude.com 9sl3dhr1twv1.statuspage.io; report-uri https://sentry.io/api/1773525/security/?sentry_key=478778eb1a5642de97c08f3c3ea33ece;"></meta>
```
vcarl: This is applied via Ops configuration as a header, and this was logging a warning due to `report-uri` in a <meta> tag. Nuked.

src/components/FormComponents/AssetPicker.js
piyalbasu: yeah i guess alternatively you could just break this out into a named function and invoke here since it's only being used once, i'm fine w this

src/components/SetupPanes/FindStrictReceivePaymentPaths.js
piyalbasu: ah yeah,`.filter(Boolean)` is kind of a neat trick - wondering if it's worth being a bit more verbose rather than explaining w comment

values["source_assets"].split(",").filter(str => str.length)

but yeah, nbd

src/components/FormComponents/AssetMultiPicker.js
```diff
@@ -0,0 +1,68 @@
...
+        </div>
+      ))}
+      <button
+        className="s-button"
+        type="button"
+        onClick={() => {
+          onUpdate(value.concat("native"));
+        }}
+      >
+        Add {value.length > 0 && "another"} asset
```
piyalbasu: can prob just do `Add {value.length && "another"} asset` here

src/components/FormComponents/AssetMultiPicker.js
```diff
@@ -0,0 +1,68 @@
...
+  value,
+  onUpdate,
+  stringForm = false,
+  ...props
+}) {
+  return (
+    <>
+      {value.map((asset, index) => (
+        <div
+          key={`${asset === "native" ? "native" : "issued"}${index}`}
```
piyalbasu: looks we use "native" and "issued" in a couple places, any value to saving them as a const?

src/components/FormComponents/AssetMultiPicker.js
vcarl: Recently I've been bit by falsy values in conditionals a few times (mostly `''`/`0` etc when I only meant it to trigger on `null` or `undefined`, but still), so I've been shooting for being explicit about what cases should trigger behavior.

src/components/FormComponents/AssetMultiPicker.js
vcarl: Actually this key can probably be simplified, this logic is from when I was trying to manage native/not-native via state in the component, and it kept resetting because the key changed.

src/components/FormComponents/AssetMultiPicker.js
vcarl: OOO I totally forgot that [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) is officially a part of JS now

src/components/FormComponents/AssetPicker.js
vcarl: Yep, it seemed like a worse option to break out into a new component or function. More indirect, by moving single-purpose functionality away from where it's used.

src/components/SetupPanes/FindStrictReceivePaymentPaths.js
vcarl: Yeah I think I'd still want to have a comment explaining what produces an empty string in the array, so 🤷‍♂

PR #429, opened 2019-12-13
Fix signer hash

vcarl:
The transaction signer is using an outdated method to set network mode, which results in an incorrect hash being displayed.

Many thanks to @jeesunikim for helping hunt down the [first bad commit](https://github.com/stellar/laboratory/commit/d6654fe4ff696bf6c32f3abfad33f9c30d43e40c) which made this fix much faster to spot, and to @bartekn for the report.


vcarl:
Had the same thought @bartekn. Ideally we'd have some stronger handling here.

vcarl:
Change behavior so it will throw in a similar scenario in the future with stellar/js-stellar-base#284

PR #436, opened 2020-03-04
Add missing endpoints

### LINKED ISSUE #426: Add new end-points to laboratory explorer.
New end-points are missing from end-point explorer.

- [x] `paths/strict-send` #425 
- [x] `paths/strict-receive` #425 
- [ ] `/offers`
- [ ] `/offers/{id}`
- [ ] `/accounts?{signer,asset}`

vcarl:
Resolves #426

### DIFF
--- a/src/components/FormComponents/AssetPicker.js
+++ b/src/components/FormComponents/AssetPicker.js
@@ -94,21 +94,51 @@ function codeValidator(value, ignoreType = false) {
   }
 }
 
-const stringFormRadioOptions = { native: "Native", issued: "Issued" };
-function AssetStringPicker({ value, onUpdate, ...props }) {
-  const isNative = !value.includes(":");
+const ASSET_TYPES = {
+  none: "none",
+  native: "native",
+  issued: "issued",
+};
+const stringFormRadioOptions = {
+  [ASSET_TYPES.native]: "Native",
+  [ASSET_TYPES.issued]: "Issued",
+};
+const optionalOptions = {
+  [ASSET_TYPES.none]: "None",
+  ...stringFormRadioOptions,
+};
+function getAssetType(assetString) {
+  if (assetString === "native") {
+    return "native";
+  }
+  if (assetString.includes(":")) {
+    return "issued";
+  }
+  return "none";
+}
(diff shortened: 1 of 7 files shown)


vcarl:
I think this is the first time the asset picker has needed to be optional. There's no way to deselect, so I added an `optional` flag that adds a `None` option to the radio buttons.

The `data/endpoints.js` is pretty noisy with changes because I ran Prettier on it, but I did that as a separate commit. The individual commits to add the new endpoints will have easier to read diffs.

New endpoints:
[accounts](https://laboratory-64d8e45.previews.kube001.services.stellar-ops.com/laboratory/#explorer?resource=accounts&endpoint=multiple&network=test)
[offers](https://laboratory-64d8e45.previews.kube001.services.stellar-ops.com/laboratory/#explorer?resource=offers&endpoint=all&network=test)
[single offer](https://laboratory-64d8e45.previews.kube001.services.stellar-ops.com/laboratory/#explorer?resource=offers&endpoint=single&network=test)

src/components/FormComponents/AssetPicker.js
```diff
@@ -94,21 +94,51 @@ function codeValidator(value, ignoreType = false) {
...
-  const isNative = !value.includes(":");
+const ASSET_TYPES = {
+  none: "none",
+  native: "native",
+  issued: "issued",
+};
+const stringFormRadioOptions = {
+  [ASSET_TYPES.native]: "Native",
+  [ASSET_TYPES.issued]: "Issued",
+};
```
piyalbasu: Nit: is this just so the radio labels are capitalized? is there any value in just using ASSET_TYPES.native and using a css text transform to capitalize instead of creating a new obj?

vcarl: Good point. In this case the label happens to be the capitalized form of the word we're using internally, but I don't think I want to couple it so that it _has_ to be that.

vcarl:
Dropping a late note: there's a small known issue on the "All Offers" endpoint—the asset picker allows you to pick `native`, but that's not a valid filter. I left it because the asset picker is already getting a little complex, might be worth revisiting at some point.

PR #438, opened 2020-03-26
Remove /laboratory/ prefix from docker image

jacekn:
The image is not in use yet and we'd like to simplify architecture a little bit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ADD . /app/src
 
 WORKDIR /app/src
 
-RUN apt-get update && apt-get install -y curl git apt-transport-https && \
+RUN apt-get update && apt-get install -y curl git make apt-transport-https && \
     curl -sSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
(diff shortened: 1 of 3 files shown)


vcarl:
I think you'll have to remove [this line](https://github.com/stellar/laboratory/blob/master/webpack.config.base.js#L11) for this to work. Build right now emits assets that assumes they're hosted from `laboratory/`

jacekn:
> I think you'll have to remove [this line](https://github.com/stellar/laboratory/blob/master/webpack.config.base.js#L11) for this to work. Build right now emits assets that assumes they're hosted from `laboratory/`

@vcarl thanks, this is a very good point. I think we need a transition period so I added rewrite rule to the dockerfile to support /laboratory/ prefix. This way we don't change anything for existing deployment while allowing dockerized one to strip the prefix.
I added @brahman81 to reviewers to get review for ops perspective as well.

brahman81:
`https://laboratory.stellar.org/laboratory` 404's whilst `https://laboratory.stellar.org/laboratory/` doesn't. (trailing `/`)

jacekn:
> `https://laboratory.stellar.org/laboratory` 404's whilst `https://laboratory.stellar.org/laboratory/` doesn't. (trailing `/`)

@brahman81 that's fixed now

PR #448, opened 2020-04-27
Add new Submit Transaction page

vcarl:
The laboratory currently doesn't, strictly speaking, have dedicated code to submit transactions to the Stellar network—it leans on the Endpoint Explorer to do so, which means it's not using the Stellar SDK. This isn't ideal, as it means we can't dogfood the JS SDK as much as we'd like to.

This PR changes the transaction flow so that it ends on a new Submit Transaction screen (which is a specialized copy of View XDR). For space reasons, I had to rename all the nav items to squish a few characters out, otherwise the new item causes "XDR Viewer" to wrap onto a new line.

Before:
Screen Shot 2020-05-01 at 2 23 36 PM
After:
Screen Shot 2020-05-01 at 2 23 27 PM

Success:
Screen Shot 2020-04-27 at 4 18 31 PM

Failure:
Screen Shot 2020-04-27 at 4 17 27 PM

SEP 29 failure:
Screen Shot 2020-04-28 at 1 20 45 PM

Video (of unfurl animation):
https://gfycat.com/smugillblowfish

### DIFF
--- a/src/components/TransactionSubmitter.js
+++ b/src/components/TransactionSubmitter.js
@@ -1,0 +1,122 @@
+import React from "react";
+import { connect } from "react-redux";
+import _ from "lodash";
+import FETCHED_SIGNERS from "../constants/fetched_signers";
+import extrapolateFromXdr from "../utilities/extrapolateFromXdr";
+import TreeView from "./TreeView";
+import validateBase64 from "../utilities/validateBase64";
+import { updateXdrInput, fetchSigners } from "../actions/xdrViewer";
+import { xdr } from "stellar-sdk";
+import { addEventHandler, logEvent } from "../utilities/metrics";
+import xdrViewerMetrics, { metricsEvents } from "../metricsHandlers/xdrViewer";
+import { TxSubmitterResult } from "./TxSubmitterResult";
+
+// XDR decoding doesn't happen in redux, but is pretty much the only thing on
+// this page that we care about. Log metrics from the component as well.
+addEventHandler(xdrViewerMetrics);
+
+const tLogEvent = _.debounce(logEvent, 1000);
+
+function TransactionSubmitter(props) {
+  let { dispatch, state, baseURL, networkPassphrase, horizonURL } = props;
+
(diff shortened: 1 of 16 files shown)


src/components/TransactionSubmitter.js
```diff
@@ -0,0 +1,148 @@
...
+          </div>
+        </code>
+      </pre>
+    </div>
+  </div>
+);
+const Error = ({ error }) => {
+  const { message } = error;
+  const { extras } = error.response?.data;
+  return (
```
abuiles, line 124: @vcarl if the error is `AccountRequiresMemoError`, can you show what's the operation and the account that requires the memo?

vcarl: Fixed! Added a screenshot to PR description

src/components/TxSubmitterResult.js
```diff
@@ -0,0 +1,181 @@
...
+    message = "An unknown error occurred";
+    extras = (
+      <React.Fragment>
+        original error:
+        <br />
+        <EasySelect>{JSON.stringify(error, null, 2)}</EasySelect>
+        <br />
+      </React.Fragment>
+    );
+  } else {
```
abuiles, line 155: btw, I think it would be worth checking that the type of error has `response` -- it might be the case that a different error appears and that would be the `else {` scenario

vcarl: Is there an error you have in mind that could cause that? I don't want to add stuff "just in case"—really, I'd love if the SDK could handle some of this complexity.

abuiles: @vcarl the SDK tries to handle as many errors as possible, however, since we rely on third parties too there might be errors that it doesn't know how to handle so it just bubbles it up to the consumer. 

In this particular scenario I don't think we are adding another handler just in case, but rather, handling unknown errors which might still happen, the else is assuming all other errors are of a given type, which isn't the case.

vcarl: Updated to make sure this will correctly handle any error thrown at it.

PR #460, opened 2020-06-18
pull base_fee from horizon, update text to pull from there too

acharb:
@vcarl @stellar/nebula 

WHAT?
Changing the base_fee source when building a transaction to be pulled from Horizon instead of a constant from the "stellar-sdk" package. If for whatever reason the pull fails, falls back to the constant.

WHY?
In case the network base fee value changes the input reflects it accordingly.

### DIFF
--- a/src/actions/transactionBuilder.js
+++ b/src/actions/transactionBuilder.js
@@ -95,7 +95,7 @@ export function fetchBaseFee(horizonBaseUrl) {
     axios.get(horizonBaseUrl + '/fee_stats')
       .then(r => dispatch({
         type: FETCH_BASE_FEE_SUCCESS,
-        base_fee: r.data.last_ledger_base_fee
+        base_fee: r.data.fee_charged.mode
       }))
       .catch(r => dispatch({type: FETCH_BASE_FEE_FAIL, payload: r}))
   }
--- a/src/actions/transactionBuilder.js
+++ b/src/actions/transactionBuilder.js
@@ -82,3 +82,23 @@ export function fetchSequence(accountId, horizonBaseUrl) {
       .catch(r => dispatch({type: FETCH_SEQUENCE_FAIL, payload: r}))
   }
 }
+
+
+export const FETCH_BASE_FEE = 'FETCH_BASE_FEE';
+export const FETCH_BASE_FEE_FAIL = 'FETCH_BASE_FEE_FAIL';
+export const FETCH_BASE_FEE_SUCCESS = 'FETCH_BASE_FEE_SUCCESS';
+export function fetchBaseFee(horizonBaseUrl) {
+  return dispatch => {
+    dispatch({
+      type: FETCH_BASE_FEE
+    });
+    axios.get(horizonBaseUrl + '/fee_stats')
+      .then(r => dispatch({
+        type: FETCH_BASE_FEE_SUCCESS,
+        base_fee: r.data.last_ledger_base_fee
+      }))
+      .catch(r => dispatch({type: FETCH_BASE_FEE_FAIL, payload: r}))
+  }
+}
+
+
--- a/src/components/TxBuilderAttributes.js
+++ b/src/components/TxBuilderAttributes.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import OptionsTablePair from './OptionsTable/Pair';
 import HelpMark from './HelpMark';
 import PubKeyPicker from './FormComponents/PubKeyPicker';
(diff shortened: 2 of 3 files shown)


src/reducers/transactionBuilder.js
```diff
@@ -106,6 +109,10 @@ function attributes(state = defaultAttributes, action) {
     return Object.assign({}, state, action.newAttributes);
   case FETCH_SEQUENCE_SUCCESS:
     return Object.assign({}, state, { sequence: action.sequence });
+  case FETCH_BASE_FEE_SUCCESS:
+    return Object.assign({}, state, {fee: action.base_fee});
```
accordeiro, line 113: Small, nonblocking detail: by looking at the [lines above](https://github.com/stellar/laboratory/pull/460/files#diff-afb08b5ed40971376e57c51329828e12R110-R111) it seems like we use were adding a whitespace between the curly braces and the content – I think we should keep it for standardization purposes. Maybe it's just a matter of running `prettier`?

src/components/TxBuilderAttributes.js
```diff
@@ -36,7 +40,7 @@ export default function TxBuilderAttributes(props) {
           value={attributes['fee']}
           onUpdate={(value) => {onUpdate('fee', value)}}
           />
-        <p className="optionsTable__pair__content__note">The <a href="https://www.stellar.org/developers/guides/concepts/fees.html">network base fee</a> is currently set to 100 stroops (0.00001 lumens). Transaction fee is equal to base fee times number of operations in this transaction.</p>
+        <p className="optionsTable__pair__content__note">The <a href="https://www.stellar.org/developers/guides/concepts/fees.html">network base fee</a> is currently set to {attributes['fee']} stroops ({attributes['fee'] / 1e7} lumens). Transaction fee is equal to base fee times number of operations in this transaction.</p>
```
vcarl, line 43: "Base fee" is a Stellar network term, I believe this should stay unchanged https://developers.stellar.org/docs/glossary/fees/#base-fee
```suggestion
        <p className="optionsTable__pair__content__note">The <a href="https://www.stellar.org/developers/guides/concepts/fees.html">network base fee</a> is currently set to 100 stroops (0.00001 lumens). Transaction fee is equal to base fee times number of operations in this transaction.</p>
```

src/components/TxBuilderAttributes.js
```diff
@@ -9,11 +9,15 @@ import TimeBoundsPicker from './FormComponents/TimeBoundsPicker';
...
 import NETWORK from '../constants/network';
-import {fetchSequence} from '../actions/transactionBuilder';
+import {fetchSequence, fetchBaseFee} from '../actions/transactionBuilder';
 
-export default function TxBuilderAttributes(props) {
+function TxBuilderAttributes(props) {
   let {onUpdate, attributes} = props;
 
+  useEffect(() => {
+    props.dispatch(fetchBaseFee(props.horizonURL))
```
vcarl, line 18: Nitpick: For consistency, it would be nice to add `horizonURL` to the destructuring above.

ire-and-curses:
>Also, @ire-and-curses IIRC there was some discussion about changing the fee_stats endpoint – can you confirm that this solution is future-proof?

Yeah this is fine. The changes have already happened. They were around the `accepted_*` names, which are now renamed `max_*`.

src/actions/transactionBuilder.js
```diff
@@ -82,3 +82,23 @@ export function fetchSequence(accountId, horizonBaseUrl) {
...
+export const FETCH_BASE_FEE_SUCCESS = 'FETCH_BASE_FEE_SUCCESS';
+export function fetchBaseFee(horizonBaseUrl) {
+  return dispatch => {
+    dispatch({
+      type: FETCH_BASE_FEE
+    });
+    axios.get(horizonBaseUrl + '/fee_stats')
+      .then(r => dispatch({
+        type: FETCH_BASE_FEE_SUCCESS,
+        base_fee: r.data.last_ledger_base_fee
```
vcarl, line 98: I'm not sure if this is the one we want, @accordeiro I believe this might be the "base fee" i.e. the defined constant? I'm not sure though. It might be better to use `fee_charged.mode` instead

accordeiro: Yep, good catch! `fee_charged.mode` seems more accurate.

src/components/TxBuilderAttributes.js
accordeiro: My understanding is that this is now (starting on Protocol 13) dynamic and can be voted, so we can't rely on a fixed number for the base fee anymore. Does that sounds right @ire-and-curses?

ire-and-curses: Yeah this is correct. In fact it could always be changed through a vote, but that couldn't happen before without potentially invalidating pre-signed transactions. Now those can be fee bumped, so in practical terms it can change now (although such changes will be infrequent).

acharb: Thanks for the clarity. Keeping the change then based on above comments @vcarl

vcarl: Ah I see, yeah that makes sense.

PR #461, opened 2020-06-23
Pass through Amplitude key from build server

vcarl:
Also optimizes the Docker build so it caches better, and removes some unused files

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+node_modules
+dist
+.tmp
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,6 @@ FROM ubuntu:16.04 as build
 
 MAINTAINER SDF Ops Team <ops@stellar.org>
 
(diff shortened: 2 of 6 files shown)


vcarl:
Metrics don't work here, but that's because there isn't an Amplitude key provided to the preview build, which I think is the right call. We could set up a parallel dev project, but I've confirmed locally that the key is passed through correctly.

jacekn:
@vcarl I think metrics won't work in production either in some cases. There is a `Makefile`  in this repository that needs `--build-arg AMPLITUDE_KEY=$AMPLITUDE_KEY` (README.md was definitely unclear on that).

Using yarn for triggering docker build requires tooling on the host which some build systems might not have. That's why I think option to trigger via Makefile is valuable.

I created #462 to try to address this.

vcarl:
Dangit! You're right, I forgot to update the makefile, I knew I had to do it and missed it. Thanks Jacek

PR #470, opened 2020-07-14
prettified and hid 'Sign with Lyra' until transaction is valid

jeesunikim:
Ticket: [Update Lab to hide `Sign with Lyra` btn until transaction is valid](https://app.asana.com/0/1168666457741233/1182884933849923)

Prettified two files: `TxBuilderResult.js` and `TxLyraSign.js`.
- It looks like we are using `requiredPragma` to prettify only the newer files. Correct me, if I am wrong 😄 

I added a comment where I updated the actual code to hide `Sign with Lyra` button until transaction is valid

### DIFF
--- a/src/components/TxBuilderResult.js
+++ b/src/components/TxBuilderResult.js
@@ -1,78 +1,130 @@
-import React from 'react';
-import {connect} from 'react-redux';
-import reduce from 'lodash/reduce'
-import {EasySelect} from './EasySelect';
-import Libify from '../utilities/Libify';
-import {txSignerLink, xdrViewer} from '../utilities/linkBuilder';
-import scrollOnAnchorOpen from '../utilities/scrollOnAnchorOpen';
-import TxLyraSign from './TxLyraSign';
+/**
+ * @prettier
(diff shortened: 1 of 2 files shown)


src/components/TxBuilderResult.js
```diff
@@ -1,78 +1,130 @@
...
-    let errorTitle = errorTitleText ? <h3 className="TransactionBuilderResult__error">{errorTitleText}</h3> : null
-    let successTitle = successTitleText ? <h3 className="TransactionBuilderResult__success">{successTitleText}</h3> : null
+    let errorTitle = errorTitleText ? (
+      <h3 className="TransactionBuilderResult__error">{errorTitleText}</h3>
+    ) : null;
+    let successTitle = successTitleText ? (
+      <h3 className="TransactionBuilderResult__success">{successTitleText}</h3>
+    ) : null;
+
+    const txBuilderHasErrors =
```
jeesunikim, line 111: `transactionBuild` is an object so it was returning true regardless whether it had errors or not.

PR #471, opened 2020-07-16
bugfix/txBuilderResult > logic to allow transactionBuild and 'readonly' to 'readOnly'

jeesunikim:
`transactionBuild` needs to be true in order for `transactionBuild.xdr` to not throw an error

### DIFF
--- a/src/components/TxBuilderResult.js
+++ b/src/components/TxBuilderResult.js
@@ -108,9 +108,6 @@ class TxBuilderResult extends React.Component {
       <h3 className="TransactionBuilderResult__success">{successTitleText}</h3>
     ) : null;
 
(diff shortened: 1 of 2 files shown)


src/components/TxBuilderResult.js
```diff
@@ -109,7 +109,7 @@ class TxBuilderResult extends React.Component {
     ) : null;
 
     const txBuilderHasErrors =
-      transactionBuild && transactionBuild.errors.length > 0;
+      transactionBuild && !transactionBuild.errors.length;
```
morleyzhi, line 112: This logic no longer matches the variable name, is that intentional? If so, the name should be updated too.

PR #474, opened 2020-07-20
Add ToS/privacy links, minor fixes

vcarl:
* Fix all offers endpoint template, which was coming through with duplicated URL
* Improve error message when async chunks fail to load
* Add ToS and privacy policy links

New async chunk error:
Screen Shot 2020-07-20 at 5 15 00 PM


vcarl:
Hmmm yea probably should. Good catch

PR #477, opened 2020-08-07
Remove amplitude module, simplify user/device metrics

vcarl:
[`amplitude-js` is about 60kb](https://bundlephobia.com/result?p=amplitude-js@4.6.0), and post-build I see the vendor bundle reduced by 58kb with this change. It's doing a lot of complex stuff to fingerprint users and devices, which we're not interested in. This PR replaces it with a much simpler and more transparent way of gathering metrics.

### DIFF
--- a/src/utilities/metrics.js
+++ b/src/utilities/metrics.js
@@ -1,24 +1,64 @@
-import Amplitude from 'amplitude-js'
+/**
+ * @prettier
+ */
+import throttle from "lodash/throttle";
 
-const instance = Amplitude.getInstance();
-instance.init(process.env.AMPLITUDE_KEY)
+const METRICS_ENDPOINT = "https://api.amplitude.com/2/httpapi";
+let cache = [];
+
+const uploadMetrics = throttle(() => {
+  const toUpload = cache;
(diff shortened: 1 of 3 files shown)


src/utilities/metrics.js
```diff
@@ -1,24 +1,64 @@
...
+
+const getUserId = () => {
+  const storedId = localStorage.getItem("metrics_user_id");
+  if (!storedId) {
+    // Create a random ID by taking the decimal portion of a random number
+    const newId = Math.random()
+      .toString()
+      .split(".")[1];
+    localStorage.setItem("metrics_user_id", newId);
+    return newId;
```
piyalbasu, line 37: if !storeId, do you need to set `storedId` to `newId` in this conditional so you have a value to return at the bottom of this?

vcarl: Nah this is an early return, no value needed in `storedId`. 
```suggestion
    storedId = newId;
```
would also work but then I'd also have to do `let storedId…` so 🤷

piyalbasu: ah you're right

PR #478, opened 2020-08-07
Use lyra api on prototype

piyalbasu:


src/components/TxLyraSign.js
```diff
@@ -36,13 +37,6 @@ const TxLyraSign = ({ xdr }) => {
       </button>
       {txResult ? (
         <div>
-          <div
-            className={`AccountCreator__spaceTop s-alert ${
-              txResult.successful ? "s-alert--success" : "s-alert--alert"
-            }`}
-          >
-            Result: {txResult.successful ? "Success!" : "Failed!"}
-          </div>
```
piyalbasu, line 45: no longer needed as we're returning a transaction, not a Horizon response

src/components/TxBuilderResult.js
```diff
@@ -128,7 +129,7 @@ class TxBuilderResult extends React.Component {
         </pre>
         {signingInstructions}
         {signingLink} {xdrLink}
-        {window.lyra && transactionBuild && !transactionBuild.errors.length ? (
+        {lyraApi.isConnected() && transactionBuild && !transactionBuild.errors.length ? (
```
piyalbasu, line 132: locally I was getting an error where `transactionBuild` hadn't been defined here 🤷

PR #480, opened 2020-08-17
update-imported-lyra/updating lyra's import method

jeesunikim:
For example, from `lyraApi.isConnected()` to just `isConnected()`.

### DIFF
--- a/src/components/TxBuilderAttributes.js
+++ b/src/components/TxBuilderAttributes.js
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import lyraApi from "@stellar/lyra-api";
+import { isConnected, getPublicKey } from "@stellar/lyra-api";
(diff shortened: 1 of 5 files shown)


src/components/TxBuilderAttributes.js
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import lyraApi from "@stellar/lyra-api";
+import { isConnected, getPublicKey } from "@stellar/lyra-api";
```
piyalbasu, line 2: ah, so I changed this to `import lyraApi from "@stellar/lyra-api";` so when you call things like `isConnected`, in the code, it's clear it's a Lyra method

I'm fine with either, honestly. Looks like we destructure methods from the other imports, so maybe it's best to do it this way

vcarl: Yeah I think either pattern is fine to use

PR #481, opened 2020-08-26
Add dummy UI elements to Lyra implementation

vcarl:


vcarl:
I don't believe so, it's just some dummy placeholders to add some elements that will be used in the final product

PR #483, opened 2020-09-17
Added 'sign with lyra' in TransactionSigner; Styled its api's result

jeesunikim:
Ticket: [success/error for lyra integration on laboratory](https://app.asana.com/0/1168666457741233/1190641579177919)

I didn't strictly follow the comp that was provided in the ticket:

**Comp:**
comp

**Updates:**
- Using `TxLyraSign` component in `TransactionSigner.js`
- Styling adjustments for responses
- The comp for error message looks like it's expecting an error code and more information. Our current lyra api response is strictly `string`. Also, in `Transaction Builder`, unless all the inputs are verified, signing buttons won't show.

**Screenshot: Transaction Builder Success**
txbuilder-result-success

**Screenshot: Transaction Builder Error**
txbuilder-result-error

**Screenshot: Transaction Signer Success**
txsigner-result-success

**Screenshot: Transaction Signer Error**
txsigner-result-error

### DIFF
--- a/src/components/TxLyraSign.js
+++ b/src/components/TxLyraSign.js
@@ -28,31 +28,6 @@ const signWithLyra = async (xdr, setTxResultMessage, setErrorStatus) => {
   }
 };
 
-const ErrorResultDetails = ({ resultMessage, hasError }) => {
-  const [isClicked, setOnClick] = React.useState(false);
-
-  let arrowIconType = isClicked
-    ? "s-button__dropdown-icon--right"
-    : "s-button__dropdown-icon";
-
-  return (
-    <div>
-      <button
-        type="button"
-        className={arrowIconType}
-        onClick={() => setOnClick(!isClicked)}
-      >
-        Details
-      </button>
-      {isClicked ? (
-        <div className="TxLyraSign__result-error">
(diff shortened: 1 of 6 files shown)


src/components/TxLyraSign.js
```diff
@@ -19,30 +19,72 @@ const signWithLyra = async (xdr, setTxResult) => {
...
-        </div>
+      {txResultMessage ? (
+        hasError ? (
+          <div className="TxLyraSign__result">
+            There was an error signing the transaction:
+            <ErrorResultDetails
+              resultMessage={txResultMessage}
+              hasError={hasError}
+            />
+          </div>
```
vcarl, line 77: If we only ever get a single string back, not more complex details, maybe this should just display the string instead of a hider element? I was thinking the hidden bit would be nice for long, obscure information like a stack trace.

vcarl (approved):
This looks great!

src/components/TxLyraSign.js
jeesunikim: Makes sense @vcarl

jeesunikim: Updated!
updated-to-string

PR #484, opened 2020-09-22
lyra-feature-address-input-btn > calling Lyra's getPublicKey() api to get a public key; updated its svg button to include alt like text

jeesunikim:
**Summary:**
- Placed an import button to get a public key in multiple input fields instead of just one "Connect Source Account with Lyra"
- Updated SVG to include `<title/>` so when users don't know what this button is about, they can just hover over and see the text "Import Public Key via Lyra Button"

screenshot-laboratory

### DIFF
--- a/src/utilities/useLyra.js
+++ b/src/utilities/useLyra.js
@@ -1,6 +1,6 @@
 /** @format */
 import React from "react";
-import { isConnected } from "@stellar/lyra-api";
+import { isConnected, getPublicKey } from "@stellar/lyra-api";
 
 export const useLyra = () => {
   const [hasLyra, setHasLyra] = React.useState(false);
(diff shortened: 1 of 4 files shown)


src/utilities/useLyra.js
```diff
@@ -10,3 +10,16 @@ export const useLyra = () => {
...
   return hasLyra;
 };
+
+export const lyraGetPublicKey = async (onUpdate) => {
+  try {
+    const lyraResponse = await getPublicKey();
+
+    if (lyraResponse.error) {
+      return lyraResponse.error;
+    }
```
piyalbasu, line 20: this is fine for this version - but just FYI, the version of lyra-api i'm about to release will properly `throw` an error if it can't find a public key rather than returning `{ error: string }`

piyalbasu (approved):

src/utilities/useLyra.js
jeesunikim: @piyalbasu updated!

PR #495, opened 2020-10-31
Update style of sign transaction buttons

### LINKED ISSUE #494: Make "Sign in with BIP path" button secondary
Thats about this button

Screen_Shot_2020-10-31_at_19 25 52

And thats the bottom, in my opinion only one should be primary
Screen_Shot_2020-10-31_at_19 26 14

gbaranski:
# Before:
Screen Shot 2020-10-31 at 22 48 16

___

# After 
Screen Shot 2020-10-31 at 22 47 52

I'm not a designer, but I still think that will be more intuitive than before. 

closes #494

### DIFF
--- a/src/components/TransactionSigner.js
+++ b/src/components/TransactionSigner.js
@@ -90,7 +90,7 @@ class TransactionSigner extends React.Component {
       if (!isUndefined(result.xdr)) {
         codeResult = <pre className="TxSignerResult__xdr so-code so-code__wrap" onClick={clickToSelect}><code>{result.xdr}</code></pre>;
         submitLink = <a
-          className="s-button TxSignerResult__submit"
+          className="s-button TxSignerResult__primary"
           href={txPostLink(result.xdr)}
           onClick={scrollOnAnchorOpen}
           >Submit in Transaction Submitter</a>;
--- a/src/styles/_TxSignerKeys.scss
+++ b/src/styles/_TxSignerKeys.scss
@@ -5,4 +5,13 @@
 
 .TxSignerKeys__signBipPath {
   margin-top: 0.5em;
+  background-color: white !important;
+  border-color: gray !important;
+  color: black;
+}
+
+.TxSignerKeys__signBipPath:hover {
+  border-color: black !important;
+  background-color: white !important;
+  color: black !important;
 }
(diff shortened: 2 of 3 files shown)


accordeiro:
Hi @gbaranski, thanks for the contribution! 

Could you share a bit about your use case, as well as the problem you're trying to solve?

Specifically, I'm not sure there's a clear hierarchy among these three actions. As I see it, it all depends on what you're using the lab for. But I'd love to understand your rationale here.

Thanks! :)

gbaranski:
Hello, the problem is those buttons color, every is made as a primary, especially the most misleading is Sign with BIP Path, and not only i encountered this problem when I was clicking this button instead of submit transaction.


Also i did a mistake, those two buttons next to submit in my opinion should be outlined.

__A layout should contain a single prominent button that makes it clear that other buttons have less importance in the hierarchy. This high-emphasis button commands the most attention.__




Source: https://material.io/components/buttons


Thanks for response

accordeiro:
Thanks! I understand your solution from a technical perspective, but what is the usability problem you're aiming to fix? Is it the fact that it's not obvious which buttons you should click on a happy path? Or is the problem that some clicks are optional, and that is not clear in the UI?

accordeiro:
Also: what task were you trying to accomplish when you faced these issues?

gbaranski:
@accordeiro That happens when I am signing transaction, and every time I press Sign with BIP Path i need to reload website, because it won't pass me with wrong Ledger Wallet input.

Actually, I think that submit buttons should always be with the highest emphasis. And yes, I think that may be also because some buttons are optional, like "View in XDR Viewer, Sign with BIP Path, Wrap with Fee bump" but in UI you see them as something most important.

gbaranski:
Are there any updates about this PR?

accordeiro:
Sorry for the late reply. I like your rationale, but I'm not sure about introducing a new button color. Maybe instead we could leave the "secondary" buttons as outlined buttons, the same way you did with the "Sign with BIP"? @vcarl @quietbits would love your thoughts here too.

As for the code: could you figure out the correct components / styles to update, so that we don't have to rely on `!important` overrides? [This blog post](https://uxengineer.com/css-specificity-avoid-important-css/) explains well the rationale for avoiding them.

quietbits:
Closing this PR for no activity.

PR #501, opened 2020-12-17
Explicitly include stellar-base

tyvdh:
This is intended as a stop gap while `js-stellar-sdk` is currently on a breaking change version which has an outdated version of `stellar-base` pinned to it.

All ultimately to allow a zero balance `createAccount` operation which the recent protocol 15 now allows for.


tyvdh:
Related issue: https://github.com/stellar/js-stellar-sdk/issues/602

quietbits:
Handled in this PR: https://github.com/stellar/laboratory/pull/502

PR #502, opened 2020-12-18
Temp fix: Explicitly include stellar-base

quietbits:
Version update for: https://github.com/stellar/laboratory/pull/501 (I can't update the fork).

NOTE: We will undo this with the next `js-stellar-sdk` version bump.


acharb (approved):
LGTM! 

I think I remember seeing somewhere that having stellar-base and stellar-sdk as dependencies may create wonkiness. You're probably aware of but just wanted to make sure

PR #512, opened 2021-02-16
Gracefully fail statuspage.io call

quietbits:
Privacy Badger Chrome extension was crashing Laboratory (nothing but a white page was rendered). This fix gracefully fails if `statuspage.io` call is blocked, showing a message in the banner.

image

### DIFF
--- a/src/components/MaintenanceBanner.js
+++ b/src/components/MaintenanceBanner.js
@@ -73,8 +73,15 @@ export default class TestnetBanner extends React.Component {
         return (
           <div className="LaboratoryChrome__network_reset_alert s-alert">
             <div className="so-chunk">
-              Failed to fetch testnet reset date. Your browser extension may be
-              blocking the call.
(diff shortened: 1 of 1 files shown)


src/components/MaintenanceBanner.js
```diff
@@ -63,6 +68,20 @@ export default class TestnetBanner extends React.Component {
...
 
+    if (!relevantMaintenance) {
+      if (currentNetwork === "test") {
+        return (
+          <div className="LaboratoryChrome__network_reset_alert s-alert">
+            <div className="so-chunk">
+              Failed to fetch testnet reset date. Your browser extension may be
+              blocking the call.
+            </div>
+          </div>
```
leighmcculloch, line 79: Instead of displaying this error message, could we link to our status page where the details can be found out instead?

quietbits: Great idea!

leighmcculloch (approved):

Issue #500, opened 2020-12-17
Error when creating a transaction with a starting balance of 0


shanzzam:
With the implementation of Protocol 15, we should be able to create a transaction with a starting balance of 0.

Currently seeing this error when doing so:
![Screen Shot 2020-12-17 at 11 21 14 AM](https://user-images.githubusercontent.com/5062622/102533384-32e47780-405a-11eb-9ffe-1c92fb0d0c99.png)

[example transaction ](https://laboratory.stellar.org/#txbuilder?params=eyJhdHRyaWJ1dGVzIjp7InNvdXJjZUFjY291bnQiOiJHQ01HUVdNN1VUV0M2S1JRUUlTTVpFU0VKUUlPUUpPMks1TzYyMzZRQkNYVkFUWklBV0QySjRWRyIsInNlcXVlbmNlIjoiMzU1MDEyOTc3MjU2MDM4NSIsImZlZSI6IjEwMCJ9LCJmZWVCdW1wQXR0cmlidXRlcyI6eyJtYXhGZWUiOiIxMDAifSwib3BlcmF0aW9ucyI6W3siaWQiOjAsImF0dHJpYnV0ZXMiOnsic3BvbnNvcmVkSWQiOiJHQkxCRVFBSFE1SlNWVERHN1FWUUVaR0dYSkJYVk9ZVFkyWUZBNkpST1IzNVFTN0xDSVBYN1FUTCJ9LCJuYW1lIjoiYmVnaW5TcG9uc29yaW5nRnV0dXJlUmVzZXJ2ZXMifSx7ImlkIjoxNjA4MjIyMDMzODk4LCJuYW1lIjoiY3JlYXRlQWNjb3VudCIsImF0dHJpYnV0ZXMiOnsiZGVzdGluYXRpb24iOiJHQkxCRVFBSFE1SlNWVERHN1FWUUVaR0dYSkJYVk9ZVFkyWUZBNkpST1IzNVFTN0xDSVBYN1FUTCIsInN0YXJ0aW5nQmFsYW5jZSI6IjAifX1dfQ%3D%3D&network=test)

quietbits:
The [quick fix](https://github.com/stellar/laboratory/pull/502) was to add the correct `js-stellar-base` version (which has the fix for 0 starting balance) manually in Laboratory. The proper fix will be to update `js-stellar-sdk` with the latest `js-stellar-base` and then use the updated `js-stellar-sdk` version in Laboratory.

[...]

PR #522, opened 2021-04-28
Protocol 17 updates

quietbits:
- [x] Upgrade `stellar-sdk` to the latest and remove manually added `stellar-base` (will properly fix https://github.com/stellar/laboratory/issues/500)
- [x] ClawbackOp
- [x] ClawbackClaimableBalanceOp
- [x] SetTrustLineFlagsOp
- [ ] ~Add to XDR types~ (It looks like this comes directly from the SDK, Laboratory doesn't handle this locally.)


leighmcculloch (approved):
LGTM, but should probably still get 👀 from someone more familiar with JS and this project.

src/components/OperationPanes/SetTrustLineFlags.js
```diff
@@ -0,0 +1,57 @@
...
+const flagsSet = {
+  'authorized': 'Authorized',
+  'authorizedToMaintainLiabilities': 'Authorized to maintain liabilites',
+};
+
+const flagsClear = {
+  'authorized': 'Authorized',
+  'authorizedToMaintainLiabilities': 'Authorized to maintain liabilites',
+  'clawbackEnabled': 'Clawback enabled',
+};
```
piyalbasu, line 19: Does it make sense to combine reused flags?

```
const DEFAULT_FLAGS = {
'authorized': 'Authorized',
  'authorizedToMaintainLiabilities': 'Authorized to maintain liabilites',
}

const flagsSet = {
...DEFAULT_FLAGS
}
etc.
```

src/utilities/Libify.js
```diff
@@ -459,6 +459,68 @@ Libify.Operation.revokeSponsorship = function(opts) {
...
+  });
+}
+
+Libify.Operation.setTrustLineFlags = function(opts) {
+  assertNotEmpty(opts.asset, 'Set Trust Line Flags operation requires asset');
+  assertNotEmpty(opts.trustor, 'Set Trust Line Flags operation requires trustor');
+
+  if (
+    !(opts.setFlags && opts.setFlags.length) &&
+    !(opts.clearFlags && opts.clearFlags.length)
```
piyalbasu, line 488: I think we could use optional chaining here to make this a little nicer 
```
!(opts.setFlags?.length)
```

src/utilities/Libify.js
```diff
@@ -459,6 +459,68 @@ Libify.Operation.revokeSponsorship = function(opts) {
...
+    throw new Error(
+      "Set Trust Line Flags operation requires at least one flag to be selected"
+    );
+  }
+
+  const flagLabels = {
+    authorized: "Authorized",
+    authorizedToMaintainLiabilities: "Authorized to maintain liabilites",
+    clawbackEnabled: "Clawback enabled",
+  };
```
piyalbasu, line 499: can we reuse `flagsClear` from above?

piyalbasu (approved):

src/utilities/Libify.js
Shaptic: didn't know about this JS feature, neat!

src/utilities/Libify.js
```diff
@@ -459,6 +459,68 @@ Libify.Operation.revokeSponsorship = function(opts) {
...
+    source: opts.sourceAccount,
+  })
+}
+
+Libify.Operation.clawbackClaimableBalance = function(opts) {
+  assertNotEmpty(opts.balanceId, 'Clawback Claimable Balance operation requires claimable balance ID');
+
+  return Sdk.Operation.clawbackClaimableBalance({
+    balanceId: opts.balanceId,
+  });
```
Shaptic, line 479: You can just pass `opts` directly, no?
```suggestion
  return Sdk.Operation.clawbackClaimableBalance(opts);
```

src/data/operations.js
```diff
@@ -144,20 +147,41 @@ export const operationsMap = [
...
+    operationPane: ClawbackClaimableBalance,
+    helpNote: 'Creates a clawback operation for a claimable balance.',
+    docsUrl: 'https://developers.stellar.org/docs/start/list-of-operations/#clawback-claimable-balance',
+  },
+  {
+    name: 'setTrustLineFlags',
+    label: 'Set Trust Line Flags',
+    operationPane: SetTrustLineFlags,
+    helpNote: 'Creates a trustline flag configuring operation.',
+    docsUrl: 'https://developers.stellar.org/docs/start/list-of-operations/#set-trustline-flags',
```
Shaptic, line 185: Slight oddity I noticed due to the inconsistency here, so just adding a comment to get your thoughts: should the URL (and corresponding heading, obviously) be `#set-trust-line-flags` instead?

src/utilities/Libify.js
Shaptic: Also, the validation is done by `clawbackClaimableBalance`, but I get it if you want to double-check the inputs.

src/components/OperationPanes/SetTrustLineFlags.js
```diff
@@ -0,0 +1,57 @@
...
+    <OptionsTablePair
+      label={<span>Set Flags</span>}
+      optional={true} key="setFlags">
+      <CheckboxButtonPicker
+        className=""
+        value={props.values['setFlags'] || []}
+        onUpdate={(value) => {props.onUpdate('setFlags', value)}}
+        items={flagsSet}
+        />
+    </OptionsTablePair>,
```
Shaptic, line 45: Is there a reason why you present the `setFlags` and `clearFlags` to the user directly? I only ask because I tried simplifying the API by letting people just set booleans ([brief discussion](https://github.com/stellar/js-stellar-base/pull/397#discussion_r598907531) for context if interested), so it's odd to see that complexity creep back in (alongside the `reduce(...)` code below to map *back* to the simplification).

Shaptic (commented):
Looks good (and I actually used it, too!), but I don't think I have the authority to give it the green ✅ heh.

src/components/OperationPanes/SetTrustLineFlags.js
quietbits: Keeping this for now for consistency. https://github.com/stellar/laboratory/issues/523

src/utilities/Libify.js
quietbits: Keeping this for now for consistency. https://github.com/stellar/laboratory/issues/523

src/utilities/Libify.js
quietbits: Good suggestion! `opts` are passed explicitly everywhere else, so trying to keep it consistent. We're validating on the frontend to make the UX better (if you're submitting a bunch of operations, it might not be great to get an error telling you to fix a simple thing like that).

src/data/operations.js
quietbits: The URL works for me with `#set-trustline-flags`. Is it changing?

src/utilities/Libify.js
quietbits: Keeping it as is for now, until we start using TypeScript.

src/components/OperationPanes/SetTrustLineFlags.js
quietbits: The "final" list is a combination of both (reduce just merges the two). I split them on the UI to make it obvious which flags are to be set and which to be cleared (`clawbackEnabled` can only be cleared). Are you suggesting that we should have a single list and by default all of them are cleared? Sorry if I don't get it.

`flags` submitted to the API are like this:

```javascript
}  
  authorized: true,
  authorizedToMaintainLiabilities: false,
  trustlineClawback: true,
}
```

Shaptic: Ah, I see what you're saying. I'm more asking why (or if?) you're submitting all of the flags regardless of whether or not the user interacted with them. Do you retrieve state first to set the flags accordingly? I wouldn't expect someone to want to set &/or clear all of the flags at once. 

To clarify, the API is as follows:

```js
sdk.Operation.setTrustLineFlags({
  /* other stuff */
  flags: {
    authorized: true, /* authorization will be set */
    clawbackEnabled: false, /* clawback will be cleared */
  }
}
```

With the above call, `authorizedToMaintainLiabilities` will **neither be set *nor* cleared**, meaning its state will be unchanged for the account. However, it appears that via the UI (correct me if I'm wrong) the flags are cleared by default, and all of the flags will always be either set &/or cleared.

Shaptic: Okay, turns out my eyes don't work well :eyes:. For posterity: it appears only flags that are clicked on in the UI will be submitted to the operation, alleviating my concern!

src/data/operations.js
quietbits: I'd vote for changing the link to keep them consistent.

PR #531, opened 2021-06-25
Ledger: replace U2F with WebUSB

quietbits:


piyalbasu (approved):

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,13 @@
+name: Test and build
+on: [push]
```
leighmcculloch, line 2: I have found it useful to run only on pull request, or push to master, so it doesn't run on every push to any branch, but push to any branch also works. Up to you, but sharing this example of how to do this if you want it: 
```suggestion
on:
  push:
    branches:
      - master
  pull_request:
```

leighmcculloch (commented):
Just commenting on the GitHub Actions workflow, one suggestion, and LGTM regardless.

.github/workflows/test-build.yml
quietbits: Thanks, @leighmcculloch ! We used to run it on every push. If it gets annoying, we can change to use only master.

PR #535, opened 2021-06-29
Set withMuxing flags dynamically

quietbits:


accordeiro (approved):

src/utilities/Libify.js
```diff
@@ -537,16 +538,17 @@ Libify.buildTransaction = function(attributes, operations, networkPassphrase) {
...
     let account;
     const sequence = Sdk.UnsignedHyper.fromString(attributes.sequence).subtract(1).toString();
+    const isMAddress = isValidMAddress(attributes.sourceAccount);
 
-    if (attributes.sourceAccount.startsWith("M")) {
+    if (isMAddress) {
       account = new Sdk.MuxedAccount.fromAddress(attributes.sourceAccount, sequence);
     } else {
       account = new Sdk.Account(attributes.sourceAccount, sequence);
     }
```
Shaptic, line 547: Just musing out loud here: when I initially designed the API, I specifically wanted to avoid people needing `if` blocks like this... Fundamentally, a muxed account is just another account. The SDK should transparently handle M accounts and G accounts and the user shouldn't have to care. Hopefully in future iterations (see e.g. stellar/js-stellar-base#436) we can remove the need for these kinds of checks.

shanzzam:
Passed testing w/ sending to both Muxed and regular accounts

PR #538, opened 2021-07-07
Add Trezor hardware wallet

quietbits:
image

### DIFF
--- a/src/utilities/trezorTransformTransaction.js
+++ b/src/utilities/trezorTransformTransaction.js
@@ -94,7 +94,7 @@ const transformTimebounds = (timebounds) => {
 /**
  * Transforms StellarSdk.Transaction to TrezorConnect.StellarTransaction
  */
-export const transformTransaction = (path, transaction) => {
+export const trezorTransformTransaction = (path, transaction) => {
   const amounts = [
     "amount",
     "sendMax",
--- a/src/utilities/trezorTransformTransaction.js
+++ b/src/utilities/trezorTransformTransaction.js
@@ -68,15 +68,17 @@ const transformType = (type) => {
 const transformMemo = (memo) => {
   switch (memo.type) {
     case StellarSdk.MemoText:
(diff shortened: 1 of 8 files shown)


src/utilities/trezorTransformTransaction.js
```diff
@@ -0,0 +1,171 @@
...
+  // those values are defined in Trezor firmware messages as numbers
+  return {
+    minTime: Number.parseInt(timebounds.minTime, 10),
+    maxTime: Number.parseInt(timebounds.maxTime, 10),
+  };
+};
+/**
+ * Transforms StellarSdk.Transaction to TrezorConnect.StellarTransaction
+ */
+export const transformTransaction = (path, transaction) => {
```
piyalbasu, line 97: I think it would be helpful to call this `trezorTransformTransaction` so when you call this in `transactionSigner` it's clear that this is specifically for Trezor

src/utilities/trezorTransformTransaction.js
```diff
@@ -0,0 +1,171 @@
...
+const transformType = (type) => {
+  switch (type) {
+    case "pathPaymentStrictReceive":
+      // case 'pathPaymentStrictSend':
+      return "pathPayment";
+    case "createPassiveSellOffer":
+      return "createPassiveOffer";
+    case "manageSellOffer":
+      // case 'manageBuyOffer':
+      return "manageOffer";
```
piyalbasu, line 63: out of scope for this PR, but we should implement an enum here during the Lab refactor

piyalbasu (approved):

src/utilities/trezorTransformTransaction.js
quietbits: We could refactor this but it might be better to keep it as close to the original created by Trezor as possible. Just in case there are updates. Aside from added types and some minor adjustments, it's pretty close to the original.

shanzzam:
@quietbits Testing pass SUCCESS. Ready to merge.

PR #542, opened 2021-07-23
Fix time bounds when min provided without max

quietbits:


src/utilities/Libify.js
```diff
@@ -576,11 +576,10 @@ Libify.buildTransaction = function(attributes, operations, networkPassphrase) {
...
     }
 
-    var transaction = new Sdk.TransactionBuilder(account, opts)
+    var transaction = new Sdk.TransactionBuilder(account, opts);
 
-    if (isEmpty(timebounds) ||
-      (opts.timebounds && opts.timebounds.maxTime == 0)) {
-      transaction.setTimeout(transaction.setTimeout(Sdk.TimeoutInfinite));
+    if (isEmpty(timebounds)) {
+      transaction.setTimeout(Sdk.TimeoutInfinite);
```
Shaptic, line 582: Could this code be simplified with the check on 572?

I'm not sure how `defaults` works, but it appears to me that 

```js
opts.timebounds = defaults(timebounds, { minTime: '0', maxTime: '0' })
```

without a conditional would achieve the same effect here, right? Because that time-bounds is equivalent to `TimeoutInfinite`.

Shaptic (approved):
Seems like it works! Minor note but otherwise :+1:

src/utilities/Libify.js
quietbits: The first check will set the default value to `0` if it's not provided (only if at least one of the time bounds is provided). Then pass that to `TransactionBuilder`. If we remove this check, SDK complains about the timeout. 🤷‍♀️ 

```
TimeBounds has to be set or you must call setTimeout(TimeoutInfinite)
```

The timeout will be set only if nothing was provided (we can do it after the transaction is built). How could we simplify this if the order matters?

Shaptic: > The timeout will be set only if nothing was provided

but won't the `defaults` line always set both `minTime` and `maxTime` to zero or whatever was provided in `timebounds` (**if** we remove the `!isEmpty` condition)? Which means `opts.timebounds` will always be set, and hence pass the timeout check?

Issue #543, opened 2021-07-28
Feature Request: Allow duplicating operations in transaction builder


hanseartic:
### What problem does your feature solve?
When a user wants to add a particular operation multiple times they must click add operation button, select the operation type and fill in all the fields. This can be a tedious task and a 'duplication' button would help to make is more easy to add repeating operations.

### What would you like to see?
A button to duplicate any operation.
![screenshot of duplication button](https://user-images.githubusercontent.com/1752217/127393295-cd9134f1-5713-472a-be6b-79642f628955.png)

### What alternatives are there?
* The current workflow
* Selecting a  bunch of operations and bulk-clone them

hanseartic:
Addressed in #544

hanseartic:
closed with a435de6

PR #544, opened 2021-07-28
Allow duplicating of operations in txbuilder

hanseartic:
Add functionality to duplicate any operation in the txbuilder of stellar laboratory as per #543 

image

image

### DIFF
--- a/src/actions/transactionBuilder.js
+++ b/src/actions/transactionBuilder.js
@@ -29,6 +29,15 @@ export let addOperation = (() => {
   }
 })();
 
+export const DUPLICATE_OPERATION = 'DUPLICATE_OPERATION';
+export function duplicateOperation(opId) {
+  return {
+    type: DUPLICATE_OPERATION,
+    sourceOpId: opId,
+    opId: Date.now(),
+  };
+}
+
 export const REMOVE_OPERATION = 'REMOVE_OPERATION';
 export function removeOperation(opId) {
   return {
--- a/src/actions/transactionBuilder.js
+++ b/src/actions/transactionBuilder.js
@@ -30,10 +30,10 @@ export let addOperation = (() => {
 })();
 
 export const DUPLICATE_OPERATION = 'DUPLICATE_OPERATION';
-export function duplicateOperation(opId) {
+export function duplicateOperation(sourceOpId) {
   return {
     type: DUPLICATE_OPERATION,
-    sourceOpId: opId,
+    sourceOpId: sourceOpId,
     opId: Date.now(),
   };
 }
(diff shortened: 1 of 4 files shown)


src/actions/transactionBuilder.js
```diff
@@ -29,6 +29,15 @@ export let addOperation = (() => {
   }
 })();
 
+export const DUPLICATE_OPERATION = 'DUPLICATE_OPERATION';
+export function duplicateOperation(opId) {
```
quietbits, line 33: Should we name this `opId` as `sourceOpId`? It might be a little bit more clear since we're assigning a different value to `opId`.

src/components/OperationsBuilder.js
```diff
@@ -75,7 +82,10 @@ let operation = (ops, index, dispatch) => {
           maxLength="2"
           className="TransactionOpMeta__order__input" />
       </div>
+      <div>
       {removeLink}
+      {duplicateLink}
```
quietbits, line 87: What if we swapped these two buttons (duplicate would always be on top)? Would it be better for the user if the button order didn't change? 🤔

src/styles/_TransactionBuilder.scss
```diff
@@ -52,8 +52,19 @@
...
+  font-size: $s-scale-down1;
+  margin: 0.25em 0;
+}
 .TransactionOpMeta__remove {
   text-align: center;
+  width: 72px;
+  padding: 0.35em 0;
+  font-size: $s-scale-down1;
+  margin: 0.25em 0;
 }
```
quietbits, line 68: These styles are exactly the same. We can create a more generic class for both buttons. What do you think? 🙂

quietbits (commented):
Thanks for the PR! 🙌 Everything works nicely, I just added a few nitpicky comments.

src/components/OperationsBuilder.js
hanseartic: Thanks. Makes a lot of sense

src/styles/_TransactionBuilder.scss
hanseartic: nice catch!

hanseartic:
> Thanks for the PR! 🙌 Everything works nicely, I just added a few nitpicky comments.

Thank you! Those were pretty valuable.

quietbits (approved):
Looks good. Thank you, @hanseartic !

piyalbasu (approved):

hanseartic:
May I bump this, @quietbits  or @piyalbasu ?

quietbits:
@hanseartic sorry for the delay! It should be merged today. 🤞

PR #553, opened 2021-08-02
Refactor: Use Create React App + TypeScript config

quietbits:
Most of the changes are from prettier and linting (updated the ones with no or very low impact).

- Updated dependencies
- Use CRA the same way we use on other FE projects (no more custom Webpack and babel config)
- Disable `eslint` rules to pass builds (we'll fix these gradually as we go)
- TypeScript support
- Disabling tests for now (we don't have many tests currently)
- Update Docker config


.prettierrc.js
```diff
@@ -1,7 +0,0 @@
-/**
- * @prettier
- */
-module.exports = {
-  ...require("@stellar/prettier-config"),
-  requirePragma: true,
-};
```
piyalbasu, line 7: I think we should still keep this file and continue using Prettier. We can use pragmas to make sure it's only implemented on the pages we want:
https://prettier.io/docs/en/options.html#require-pragma

quietbits: Good catch! We're using `prettier.config.js` name in other projects, so I used that here as well, but I forgot to add `requirePragma` param.

piyalbasu (approved):
Looks like the majority is Prettier changes that shouldn't affect functionality. If that's the case - Approved!

PR #577, opened 2021-08-05
BUG: Fix `claimClaimableBalance` source account being ignored

Smephite:
If a source account was specified in the `Claim Claimable Balance` operation it was ignored and thus the transactions source account was being used.  
I was able to trace this back to a misnamed field in the `Libify.js`.


quietbits (approved):
Thank you for fixing this! 🎉

piyalbasu (approved):

PR #586, opened 2021-08-11
Redux Toolkit config

quietbits:
- Configure Redux Toolkit for Redux store and middleware (+ Redux DevTools)
- Move Redux store to external file `config/store.ts` (same as in other projects)
- Move saving/loading network setting from local storage to `NetworkPicker` (used to be in the store)


src/config/store.ts
```diff
@@ -0,0 +1,49 @@
...
+  accountCreator,
+  endpointExplorer,
+  transactionBuilder,
+  transactionSigner,
+  xdrViewer,
+  network,
+  routing,
+});
+
+export const resetStoreAction = createAction(RESET_STORE_ACTION_TYPE);
```
piyalbasu, line 36: jc - where do you envision this being used?

src/helpers/networkLocalStorage.ts
```diff
@@ -0,0 +1,22 @@
...
+  name: string;
+  horizonURL?: string;
+  networkPassphrase?: string;
+}
+
+export const networkLocalStorageSaveValue = (
+  values: NetworkLocalStorageSaveProps,
+) => {
+  try {
+    localStorage.setItem(LOCAL_STORAGE_NETWORK, JSON.stringify(values));
```
piyalbasu, line 13: this key/value pair doesn't match what currently exists in prod. I think we should keep it the same unless there's a really really good reason not to. I just don't want to introduce bugs because of an API change:
Screen Shot 2021-08-11 at 3 22 31 PM

src/config/store.ts
quietbits: I thought we'll need something like that when clearing the form (might need to tweak it to exclude network, maybe?). Currently, it clears only the active form, but it should clear everything to avoid manual clearings on each page.

image

src/helpers/networkLocalStorage.ts
quietbits: This change was intentional because we don't want to mutate the state directly, which the "old" way does. The update is to dispatch the same action as picking the network when the component mounts. Did you notice anything wrong there? I thought I tested it well, but I might have missed something.

quietbits: To clarify, we used to save the whole network store in local storage. The update makes the change dynamically by dispatching the action.

image

PR #592, opened 2021-08-13
Refactor LaboratoryChrome (renamed to AppContent)

quietbits:
- Rename `LaboratoryChrome` to `AppContent` (after we refactor routing and `MaintenanceBanner` we can move this directly to `App.tsx`)
- Move components to their own files:
  - `Header`
  - `Footer`
  - `LoadingView`
  - `Navigation`
  - `NotFoundPage`
- Add `useRedux` hook
- Update tests to use TypeScript
- Add `AppContent` tests
- Run tests in GitHub Action


.github/workflows/test-build.yml
```diff
@@ -9,5 +9,5 @@ jobs:
         with:
           node-version: 14
       - run: yarn install
-      # - run: yarn test
+      - run: yarn test:jest
```
piyalbasu, line 12: I would run both `yarn test` and `yarn test:jest` bc `yarn test` runs the existing tests in `./test`

piyalbasu (approved):

.github/workflows/test-build.yml
quietbits: `yarn test` is broken right now. 😬 More refactoring might fix them.

PR #594, opened 2021-08-16
Disable INLINE_RUNTIME_CHUNK for Create React App

quietbits:
Disabling `INLINE_RUNTIME_CHUNK` will create an external file (instead of loading inline) for the script. Let's see if this will help. 🤞 

Source: https://create-react-app.dev/docs/advanced-configuration/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,8 @@ RUN mkdir -p /app
 WORKDIR /app
 
 ENV DEBIAN_FRONTEND=noninteractive
+# https://create-react-app.dev/docs/advanced-configuration/
+ENV INLINE_RUNTIME_CHUNK=false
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,6 +6,8 @@ RUN mkdir -p /app
 WORKDIR /app
 
 ENV DEBIAN_FRONTEND=noninteractive
+# https://create-react-app.dev/docs/advanced-configuration/
+ENV INLINE_RUNTIME_CHUNK=false
```
quietbits, line 10: Is this the correct way of setting `env` variable in Docker? This flag should work like setting it in `.env` file (worked on build locally).

jacekn: This will set the environment variable during the build only. So `yarn install` will see it when the image is built but it won't be set when the resulting image is run.

jacekn (approved):

PR #601, opened 2021-08-19
Refactor AccountCreator

quietbits:
- `PubKeyPicker` and `MuxedKeyPicker` converted to TS because the `placeholder` prop was breaking tests for `AccountCreator` (was considered required). We can combine these into one component in Tier 3.
- Most of the file updates are for updating the `PubKeyPicker.js` import.
- Moved `KeypairGenerator`, `FriendbotFundAccount`, and `MuxedAccount` to their own files.


src/components/FormComponents/TextPicker.js
```diff
@@ -23,6 +25,7 @@ export default function TextPicker(props) {
         }}
         placeholder={props.placeholder}
         className={`picker picker--textInput ${props.className}`}
+        {...(testId ? { "data-testid": testId } : {})}
```
quietbits, line 28: Added `data-testid` here because it was much better to test that way. Getting component by text was returning multiple elements.

src/helpers/testHelpers.tsx
```diff
@@ -8,7 +9,11 @@ import { reducers } from "config/store";
 
 export const render = (
   ui: React.ReactElement,
-  { preloadedState, store = createStore(reducers), ...renderOptions }: any = {},
+  {
+    preloadedState,
+    store = createStore(reducers, applyMiddleware(thunk)),
```
quietbits, line 14: Had to add `applyMiddleware(thunk)` to handle nested `dispatch` in actions.

piyalbasu (approved):
Awesome!!

PR #604, opened 2021-08-23
Refactor ExploreEndpoints

quietbits:
- Move `msw` server config to `testHelpers` for reusability.
- Add tests to cover the basic happy path to all endpoints in `EndpointExplorer` (we could add more specific tests for form inputs, etc., but I'm not sure how much value we'll get out of that considering how much time it will take to write those tests 🤷‍♀️).
- Refactor `EndpointExplorer` component.
- Move `buildRequest` and `buildRequestUrl` to helpers.
- Convert `data/endpoints` to TypeScript.


src/helpers/testHelpers.tsx
```diff
@@ -20,3 +22,422 @@ export const render = (
...
+  rest.get("https://horizon-testnet.stellar.org/effects", (_req, res, ctx) => {
+    return res(
+      ctx.status(200),
+      ctx.json({ test: ENDPOINT_RESPONSE.all_effects }),
+    );
+  }),
+  // Endpoints: effects > effects for account
+  rest.get(
+    // TODO: fix URL to have only one / at the end
+    "https://horizon-testnet.stellar.org/accounts//effects",
```
piyalbasu, line 122: a few of these have a `//` - typo?

src/helpers/testHelpers.tsx
```diff
@@ -20,3 +22,422 @@ export const render = (
...
+      return res(
+        ctx.status(200),
+        ctx.json({ test: ENDPOINT_RESPONSE.effects_for_account }),
+      );
+    },
+  ),
+  // Endpoints: effects > effects for ledger
+  rest.get(
+    // TODO: fix URL to have only one / at the end
+    "https://horizon-testnet.stellar.org/ledgers//effects",
```
piyalbasu, line 133: thanks for adding all of these stubs! Small improvement might be to set `const HORIZON_URL = https://horizon-testnet.stellar.org/` so we can easily configure this to work on mainnet (for ex) later, if needed

piyalbasu (approved):

src/helpers/testHelpers.tsx
quietbits: Yeah, it looks like the bug is in the URL generator if there is a second param. I didn't want to fix this now in case it's used somewhere else and we can't test it.

Issue #557, opened 2021-08-02
Lab refactor: Tier 2 work


shanzzam:
This work will include file conversion and creating end-to-end tests. Specifically:
Conversion to Typescript
[Style guide](https://github.com/stellar/product-conventions/blob/master/STYLEGUIDE.md) implementation
Replacement of Redux connect with Redux-React Hooks
Create end-to-end tests

- [x] #580 
- [x] #558
- [x] #590
- [x] #559
- [x] #560
- [x] #561
- [x] #562
- [x] #631 
- [x] #564
- [x] #565
- [x] #566
- [x] #575 
- [ ] Combine actions/reducers into ducks using Redux toolkit for accountCreator, endpointExplorer, network, routing, transactionBuilder, transactionSigner, xdrViewer
- [x] #630

quietbits:
Closing in favor of the v2 rebuild

PR #613, opened 2021-09-22
Bug fix: timebounds reset when memo changed in TX builder

quietbits:


src/views/TransactionBuilder.js
```diff
@@ -95,6 +95,7 @@ function onAttributeUpdate(dispatch, param, value) {
     case "memo":
       newAttributes.memoType = value.type;
       newAttributes.memoContent = value.content;
+      break;
```
leighmcculloch, line 98: ❔ Would it be realistic to write a test for this? I see that tests are pretty light already in this repo, so maybe there isn't the framework for it.

leighmcculloch (approved):
🎉 Looks good, thanks for fixing.

One question below, but it doesn't need to block merging.

src/views/TransactionBuilder.js
quietbits: Tests would have definitely caught this bug. 💯 We are adding tests as we keep refactoring, but we haven't reached the `TransactionBuilder` view yet. Hopefully, we'll have time to get to it soon. Tier 2 progress can be found [here](https://github.com/stellar/laboratory/issues/557).

piyalbasu (approved):

PR #624, opened 2021-12-06
Add an introduction/tour 

Kanaye:
This is a more or less "prototype" version to pitch the idea.
Most of the work I've done started at Meridian 2021's Docsprint.
You can find a [demo of the current state here](https://laboratory.kany.xyz).

## The idea 💡 
Most developers starting their journey into the Stellar Network will sooner or later meet Laboratory.
While Laboratory is an awsome tool, it can be hard to wrap your head around at first.
So a more or less interactive Tutorial/Tour should give people a headstart on where they need to look out for specific features.

I could go deep into Labs features but for now it's a very basic overview over the different pages. We could expand it later as far as needed.

## The implementation ⚙️ 
I've seen (non open source) tour components build in react, but all of them have/had trouble with steps across js-chunks and blow up the js bundle size even when they're not used by the user.
For this prototype I'm using [Sheperd](https://shepherdjs.dev/), a well known library for Tour overlays.
The js is dynamically import()-ed when the tour is started, the only overhead loaded every time is shepherds css as I did't get around to lookup how I could dynamically import it without breaking jest.

### Open TODOs / Things to discuss 🔧 
- In this version I'm manipulating Labs state by [emitting click events](https://github.com/Kanaye/laboratory/blob/430526851dabf5d338b695d8dd7a728c8635c1a9/src/views/Tour.ts#L43-L54) on buttons/links.
Most of this could be replaced by using the redux store, as this code is fairly ugly and easy to break.
But manipulating the store could lead to different behavior of the tour and the real UI elements.
- The selectos I've used are either [href attributes](https://github.com/Kanaye/laboratory/blob/430526851dabf5d338b695d8dd7a728c8635c1a9/src/views/Tour.ts#L89), [easy to break selector chains](https://github.com/Kanaye/laboratory/blob/430526851dabf5d338b695d8dd7a728c8635c1a9/src/views/Tour.ts#L74) or event [data-testid](https://github.com/Kanaye/laboratory/blob/430526851dabf5d338b695d8dd7a728c8635c1a9/src/views/Tour.ts#L106). This should probably be changed to either fixed ids or it's own data-attribute like `data-tour-anchor` to make it more obvious that this element is part of the tour.
- Integration tests: As the tour (by it's nature of living along with react) is seperated from the views but interacts with them, it's fairly easy to break it by accident. So adding integration tests to ensure that elements are highlighted and the "next" button works are probably a good idea.

---

I'll happily continue work on this, but it may take time as I'm fairly busy.
But I'd love to hear thought and feedback on the idea, my points from above and anything else implementation related.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.13.9",
     "route-recognizer": "^0.3.4",
+    "shepherd.js": "^8.3.1",
     "solar-stellarorg": "git+https://github.com/stellar/solar-stellarorg#master",
     "stellar-sdk": "^9.0.0-beta.0",
     "trezor-connect": "^8.1.27",
--- a/src/constants/tour.ts
+++ b/src/constants/tour.ts
@@ -10,6 +10,7 @@ const options: Tour.TourOptions = {
         },
         buttons: [
             {
+                classes: 's-button',
                 text: 'Continue',
                 action() {
                     this.next();
@@ -37,6 +38,17 @@ const awaitElementExistence = (selector: string) => {
     });
 };
 
+/**
+ * The other Hackaround. 
+ * Could probably be replaced by manipulating state-management directly.
+ */
+const click = (selector: string) => {
+    const element = document.querySelector(selector);
+    if (element instanceof HTMLElement) {
+        element.click();
+    }
+};
+
 
 const steps: Step.StepOptions[] = [
     {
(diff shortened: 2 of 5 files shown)


quietbits:
The tour idea is great, @Kanaye ! Thank you very much for starting the conversation in this PR!

We are in the middle of refactoring Lab to match our other frontend projects, so the Redux implementation could be an option. Would you like to give it a try? No need to write code just yet, but a high-level idea could be a good start. The code structure eventually will be the same as in the [Account Viewer](https://github.com/stellar/account-viewer-v2). It's also OK to say that we should figure this out ourselves when the time comes. :smile: 

I like your suggestion to use the same `data-` attribute for all tour elements. That way it would be clear what is used where and won't break if some class is changed during refactoring.

Adding tests would be awesome, but we can probably do that later. The same goes for style tweaks.

Thanks again for your work on this! We understand that you're busy, so please take your time with any updates. :raised_hands: :rocket:

BertsLifee (approved):

quietbits:
This feature won't work with Lab v2

PR #640, opened 2022-01-21
add fee_stats endpoint to EndpointExplorer

silence48:
This adds a fee_Stats all endpoint to the endpoint explorer.

#tyvdh/stellar-quest-bounties#155
#tyvdh/stellar-quest-bounties#156


src/data/endpoints.ts
```diff
@@ -167,6 +167,21 @@ export const endpointsMap: EndpointsMap = {
...
+    label: "Fee Stats",
+    endpoints: {
+      all: {
+        label: "Fee Stats",
+        helpUrl: "https://developers.stellar.org/api/aggregations/fee-stats/object/",
+        method: RequestMethod.GET,
+        path: {
+            template: "/fee_stats",
+        },
+        setupComponent: All,
```
silence48, line 180: All should probably be the default setup Component to avoid code duplication

src/data/endpoints.ts
```diff
@@ -5,6 +5,7 @@ import { AllLiquidityPools } from "components/SetupPanes/AllLiquidityPools";
 import AllOffers from "components/SetupPanes/AllOffers";
 import AllWithFailed from "components/SetupPanes/AllWithFailed";
 import ClaimableBalances from "components/SetupPanes/ClaimableBalances";
+import Fees from "components/SetupPanes/Fees";
 import FindPaymentPaths from "components/SetupPanes/FindPaymentPaths";
```
silence48, line 9: The SetupPane for "all" was inappropriate for fees. so I created a new one for Fees.

src/data/endpoints.ts
```diff
@@ -167,6 +168,20 @@ export const endpointsMap: EndpointsMap = {
       },
     },
   },
+  fee_stats: {
+    label: "Fee Stats",
+    endpoints: {
+      all: {
+        label: "Fee Stats",
```
quietbits, line 175: I think we should make this label "All Fee Stats" for consistency, what do you think?

src/data/endpoints.ts
```diff
@@ -167,6 +168,20 @@ export const endpointsMap: EndpointsMap = {
...
+        helpUrl: "https://developers.stellar.org/api/aggregations/fee-stats/object/",
+        method: RequestMethod.GET,
+        path: {
+            template: "/fee_stats",
+        },
+        setupComponent: Fees,
```
quietbits, line 181: If we show the message that there are no options, it will need to be styled.

image

We could leave it blank, though. Do you think it will be confusing to the users that there is nothing more to select? 

To render nothing, we could replace `Fees` with `() => null`.

image

src/views/__tests__/EndpointExplorer.test.tsx
```diff
@@ -302,6 +302,8 @@ describe("claimable balances", () => {
   });
 });
 
+
+
```
quietbits, line 306: Extra spaces

src/views/__tests__/EndpointExplorer.test.tsx
```diff
@@ -557,6 +559,66 @@ describe("effects", () => {
   });
 });
 
+describe("fee_stats", () => {
```
quietbits, line 562: The tests are failing. Could you please run them locally (`yarn test:jest`) to check? Otherwise, the build check will fail. We should probably run them before committing.

image

quietbits (commented):
Thanks for the contribution, @silence48! :tada: Left some comments, please let us know if you have any questions.

src/data/endpoints.ts
silence48: I think that's a good plan! I will update!

src/data/endpoints.ts
silence48: I think rendering nothing looks better.  I tried a styled note but it seems superfluous.  I will change it to render nothing.

src/views/__tests__/EndpointExplorer.test.tsx
silence48: I must have not updated the test after I modified the setup pane; I'll fix tests and comment back once all this is resolved. sometime later today.
Thank you!

src/data/endpoints.ts
silence48: image
This definitely looks better.

src/views/__tests__/EndpointExplorer.test.tsx
silence48: I just got to look at this, and while fixing the tests for all_fees i am seeing some other console errors while running tests. 
image
I think these might be related to 29f460f not sure how anything i touched could mess up something related to signers.

silence48:
Tests are passing. All issues repaired.
image
Fee_Stats component added:
image

src/data/endpoints.ts
silence48: I ended up deleting this and returning null. no need to duplicate code for an obvious point.

src/data/endpoints.ts
silence48: I have changed the .all endpoint label to "All Fee Stats" for consistency.

src/views/__tests__/EndpointExplorer.test.tsx
silence48: All tests pass, but there is still an error about the signature xdr unrelated to me.  That will have to be resolved in later.

quietbits (approved):
Awesome work, @silence48 ! :clap: Thank you! Don't worry about the test issues in `TransactionSubmitter.test.tsx`, it shouldn't be related to your PR.

PR #647, opened 2022-04-29
Stellar SDK update for Protocol 19

quietbits:


piyalbasu (approved):

leighmcculloch:
One thing I noticed is that the payload byte array shows up as base64 in the xdr viewer. If that's the way it has to be no worries. But if it is possible for it to be hex, it will make it easier to understand txs because the payload is most likely a hash to be signed, and most likely a tx hash, and tx hashes we usually express as hex.
Screen Shot 2022-04-29 at 2 13 06 PM

If this isn't a small quick change, please disregard.

quietbits:
@leighmcculloch it could be a relatively easy update. Do we need to format any `payload` key's value this way, or are there special criteria for this one?
Here's an example of nodes data we get:

image

leighmcculloch:
It would be fine to format all `ed25519SignedPayload` `payload`s as hex. They are min 1 byte, max 64 bytes.

leighmcculloch:
Actually, it appears there might be a bug with the rendering of signatures.

Take this transaction for example:
```
AAAABQAAAAAn7cGlUcWaNo5TiKabXmJV9emwIlWcVsR+84ZtgjPyUAAAAAAAAADIAAAAAgAAAAB0FnjE0UuuBcF379tOTBX+BfAEtuZNW+3lZWWqwHE4JQAAAAAAAA/FAAAACQAAAAIAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAPxQAAAAEAAAAAAAAAAAAAAAAAAAABAAAAAyftwaVRxZo2jlOIppteYlX16bAiVZxWxH7zhm2CM/JQAAAAIMUF4UC+7igM6EREaPHzpz/UlLrPjcpnNEdRfdxlwGJSAAAAAAAAAAEAAAAAAAAACwAAAAAAAAAAAAAAAAAAAAMQzn1lAAAAQJOpNCyRDMNZnN040dr8G1bYtNHOKGXT/8fYjGNkYKbYUb/Y0nfwDwtqWZbqpHV2dyUx8K6Nxf+zlVov4xpCug6CM/JQAAAAQKx56sseLbcGqp27v0fvFOd/f7L3Eg+qm4r85waI+k+hjA8UJ8+vxySYc14FVj1797v6eV7R+iWIetYwyfYIqgLn85ACAAAAQB1/Ms0DIe6Cv8nAxr0VNCc/GO81y4OWOidGkW55CPS/RgOlVuuQMcKmYz5mF1SbQRVQbkIhmi3ytiT3M8455QsAAAAAAAAAAYIz8lAAAABAPJuAFojiMvhqLdvl8/qGqmC+DO8w8B/pfwu1e53kQPsygbmXljx+MX/zg3lv/ueks9YuP7XtJWxzHPK2eNpJBg==
```
Screen Shot 2022-04-29 at 4 31 29 PM

The way we render signatures is now incorrect. We render them as if the hint is the 4 bytes from the key, but actually it depends on the type of signature. In the case of signed payload signers the the hint is an XOR of the G key and the payload, so the way the lab renders the signature with the G on the front of the hint is incorrect, in the case of the third sig, it would be `P`. But the lab won't have access to any information to know if it should put a G or a P, so maybe it should stop doing that?

cc @brunomuler

quietbits:
@leighmcculloch Would it be OK to fix both payload and signatures in a follow-up PR?

leighmcculloch:
> Would it be OK to fix both payload and signatures in a follow-up PR?

Yes I think so. These are both minor cosmetic things. The signature hint is maybe something we also need time to brainstorm about.

quietbits:
Follow-up issue https://github.com/stellar/laboratory/issues/649

PR #651, opened 2022-05-05
Changing 'signatures' to 'transactions' on the #txsigner view introduction

ElliotFriend:
I noticed this while guiding a quester through transaction building/signing/submitting. It seems like the word 'transactions' is what is meant here, rather than 'signatures.'

I very well could be wrongly assuming the intent, and if I am feel free to ignore this PR. Just thought I'd submit it in case it's a useful catch. 👍🏻


quietbits (approved):
`transaction` makes more sense to me there. :+1:

PR #657, opened 2022-08-30
Changing the `HelpMark` link for the `homeDomain` option

ElliotFriend:
It was pointing previously to the additional-signing-keys link,
and this seems like a more relevant link to use.

Another possible link could be found on the page about federation,
but that seems less useful for the `setOptions` operation in particular:
https://developers.stellar.org/docs/glossary/federation#looking-up-federation-provider-via-a-home-domain-entry

Signed-off-by: Elliot Voris <elliot@voris.me>

### DIFF
--- a/src/components/OperationPanes/SetOptions.js
+++ b/src/components/OperationPanes/SetOptions.js
@@ -216,7 +216,7 @@ export default function SetOptions(props) {
       label={
         <span>
           Home Domain{" "}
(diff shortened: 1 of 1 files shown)


quietbits (approved):

quietbits:
Thank you for fixing this, @ElliotFriend ! :rocket:

piyalbasu (approved):

PR #662, opened 2022-09-28
add useIsSoroban hook, add banner and use soroban-sdk if soroban in xdrView

acharb:
Screen Shot 2022-09-28 at 9 13 04 AM

### DIFF
--- a/src/helpers/extrapolateFromXdr.js
+++ b/src/helpers/extrapolateFromXdr.js
@@ -9,16 +9,210 @@
 // - string: string values that appear as just plain text
 // - object: typed values always with a type and value `{type: 'code', value: 'Foo();'}`
 
-import { xdr, StrKey, Keypair, Operation } from "stellar-sdk";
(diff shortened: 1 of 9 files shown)


src/helpers/extrapolateFromXdr.js
```diff
@@ -9,195 +9,219 @@
...
-  } else if (object.switch) {
-    parseArm(anchor, object);
+  let xdr, StrKey, Keypair, Operation;
+  if (isSoroban) {
+    xdr = SorobanSdk.xdr;
+    StrKey = SorobanSdk.StrKey;
+    Keypair = SorobanSdk.Keypair;
+    Operation = SorobanSdk.Operation;
   } else {
-    parseNormal(anchor, object);
```
acharb, line 45: just moved all these helper methods w/in `extrapolateFromXdr`, only change was the xdr, StrKey etc var assignments

PR #672, opened 2022-10-20
Updates keypair generator view to include funding link on futurenet

ElliotFriend:
The link to fund with freidnbot underneath the generated public/secret key does not show up when using futurenet. It's a small problem, but has slowed me down enought times I thought I'd contribute a fix.


quietbits (approved):
:rocket:

piyalbasu (approved):

PR #675, opened 2022-11-05
Enable Friendbot and Soroban on custom network

Smephite:
When working with a local instance of the [quickstart](https://github.com/stellar/quickstart) custom network to develop on Soroban it is quite useful to still be able to use laboratory via the `custom` network option.  
Sadly this type of network does not include the UI for friendbot nor the new changes needed to parse Soroban compatible XDRs.  

This PR enables both of these features.


src/actions/accountCreator.js
```diff
@@ -22,10 +23,10 @@ export function updateFriendbotTarget(target) {
 
 export const START_FRIENDBOT_REQUEST = "START_FRIENDBOT_REQUEST";
 export const FINISH_FRIENDBOT_REQUEST = "FINISH_FRIENDBOT_REQUEST";
-export function startFriendbotRequest(target, isSoroban = false) {
-  const friendbotURL = isSoroban
-    ? "https://friendbot-futurenet.stellar.org"
-    : "https://friendbot.stellar.org";
+export function startFriendbotRequest(target, isSoroban = false, isCustom = false) {
+  const friendbotURL = isCustom ? (networkLocalStorageGetValue()?.horizonURL + "/friendbot"): (isSoroban
```
quietbits, line 27: To make sure that the friendbot exists on the custom network, maybe we could fetch the URL to see if we get a response. Only if we do get a response (even if it's a 400), we show the fund the account section. What do you think, @Smephite ?

quietbits (commented):
Thanks for adding this feature, @Smephite ! :rocket: It looks good. I just added a comment about the check that would be nice to have. Let me know what you think. :pray:

src/actions/accountCreator.js
Smephite: That would definitely work.
The question is if we care to check this every single time when showing the fund the account section thus creating quite a lot of duplicate queries or if we want to store a map of all custom networks we ever encountered and storing a true/false whether friendbot was encountered or not in local storage

quietbits: We could set a flag in the store for friendbot for custom network only. Something like `isCustomFriendbotEnabled` (with a better name :sweat_smile:). So this check would happen only once when the network is set, then data is taken from the store.

quietbits:
This should work in Lab v2 https://lab.stellar.org/.

PR #685, opened 2023-03-17
Update Amplitude tracking + to Node v16

quietbits:
- Update Node from v14 to Node v16
- Add `.nvmrc` file for convenience
- Use JS variable instead of `env` for Amplitude tracking


piyalbasu (approved):
🎉 🎉 🎉

PR #703, opened 2023-07-05
upgrade soroban-client for Preview 10

piyalbasu:


aristidesstaffieri (approved):
lgtm but the build seems to fail due to repo permissions, not sure if this is existing.

Issue #710, opened 2023-08-08
Transaction Builder Page UI Crash


SyedMuqeetAqib:
### What version are you using?
0.0.1 in package-lock.json

### What did you do?
on https://laboratory.stellar.org/#txbuilder page is crashing when user tries to enter transaction details, more specifically for me the page is crashing when user tries to build a transaction with "change trust" operation and after adding asset type and asset code when user enters issuer address, the page crashes.

### What did you expect to see?
The page should not crash and it should let user continue tx building process while showing relevant tx status of being successful or error

### What did you see instead?
white screen (crash)

janewang:
Closing as we launched Lab https://lab.stellar.org/

PR #711, opened 2023-08-08
Fix/tx build UI: Fix the break UI change on the transaction builder page

SyedMuqeetAqib:
##Issue
https://github.com/stellar/laboratory/issues/710

## Description

Fix the break UI change on the transaction builder page. When filling in the inputs error title or success title changes conditionally breaking the transaction builder page and the user was unable to process further their transaction through UI. I was trying to execute changeTrust operation specifically where I found that bug.

## Changes Made

In TxBuilderResult.js, wrapped the successTitle and errorTitle headings in the parent div, these lines were causing an error when the success heading or error heading appears conditionally.

## Checklist
- [*] I have tested my changes manually and they work as intended.
- [*] I have mentioned the reviewers using the GitHub reviewer assignment feature and/or "@username" mentions in the description/comments.

## Screenshots (if applicable)

image


quietbits:
This should be fixed in Lab v2 https://lab.stellar.org/

PR #719, opened 2023-09-20
fix unsignedHyper's new prototype

piyalbasu:


src/actions/transactionBuilder.js
```diff
@@ -88,7 +87,7 @@ export function fetchSequence(accountId, horizonBaseUrl) {
       .then((r) =>
         dispatch({
           type: FETCH_SEQUENCE_SUCCESS,
-          sequence: UnsignedHyper.fromString(r.data.sequence).add(1).toString(),
```
piyalbasu, line 91: UnsignedHyper used to extend `long.js` which had built in addition/subtraction methods. This is no longer the case and (for some reason) TS did't warn against this. We can just convert sequence number to Number and do addition manually

aristidesstaffieri (approved):

PR #723, opened 2023-10-28
Update stellar-sdk import in claimantHelpers.js

Row-Bear:
The Create Claimable Balance operation fails to generate an XDR if a predicate is used, instead reporting 
"- Operation #1: o.a is undefined "
I believe that is due to the import of stellar-sdk, which was updated recently for other parts


piyalbasu (approved):

piyalbasu:
@Row-Bear Sorry for the late review - but thank you for catching this! I'm going to add this to the next release

PR #729, opened 2023-11-28
Update Trezor packages

quietbits:


piyalbasu (approved):

aristidesstaffieri (approved):

felixlucid:
As a Trezor Model T user, I can confirm that as of now, the StellarLab is still unable to sign XDR via the Trezor hardware wallet. This prevents the claiming of any airdrops or claimable balances. When attempting to sign, I now receive the error of: Error:"Cannot destructure property 'type' of 't[r]' as it is undefined."

PR #740, opened 2024-02-14
V2 Layout and basic page setup

quietbits:
- Layout and basic scaffolding for pages and navigation. 
- Added helper link components to make sure all links behave the same (work with client-side routing and open external links in new tab/window).
- Using SDS `2.0.0-beta.1`,

> [!NOTE]  
> Routing is not working when the page is refreshed because we're using static build. Ops will create a [Node server deployment](https://github.com/stellar/ops/issues/2846) that should fix this issue.


src/components/MainNav.tsx
```diff
@@ -32,6 +32,10 @@ const secondaryNavLinks: NavLink[] = [
 
 export const MainNav = () => {
   const pathname = usePathname();
+  const searchParams = useSearchParams();
+
+  console.log(">>> pathname: ", pathname);
+  console.log(">>> searchParams: ", searchParams.toString());
```
quietbits, line 38: Leaving these for now to test server setup

aristidesstaffieri (approved):

PR #741, opened 2024-02-20
V2 Basic demo setup for state management

quietbits:
State management setup with a basic demo. I will clean up the demo pieces in the next PR. This is to validate the approach and setup.

Using:
- [Zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) for global state management
- [zustand-querystring](https://github.com/nitedani/zustand-querystring) to create shareable link
  - I went through various options (mostly custom ones), and this one was the best. If something doesn't work with more complex data, we can issue a PR with the fix or fork it.

> [!NOTE]  
> Routing is not working when the page is refreshed because we're using static build. Ops will create a [Node server deployment](https://github.com/stellar/ops/issues/2846) that should fix this issue.


src/store/createStore.ts
```diff
@@ -0,0 +1,83 @@
...
+      nestedValue2: number;
+    };
+    update: (value: string) => void;
+    updateNested: (nestedVal: {
+      nestedValue1: string;
+      nestedValue2: number;
+    }) => void;
+    reset: () => void;
+  };
+}
```
quietbits, line 32: We can extract sections into the variables if that's easier to manage.

src/store/createStore.ts
```diff
@@ -0,0 +1,83 @@
...
+        },
+      })),
+      {
+        url: options.url,
+        // Select what to save in query string
+        select() {
+          return {
+            network: true,
+            account: true,
+          };
```
quietbits, line 78: This part controls what we want to keep in the URL

aristidesstaffieri (approved):
yay zustand + immer ❤️

src/store/createStore.ts
```diff
@@ -0,0 +1,83 @@
...
+        },
+        key: "||",
```
jeesunikim, line 80: do we want this vs. its default `"$"`, @quietbits ?

quietbits: @jeesunikim I thought it was a safer option because we could have `$` somewhere (amount in dollars. for example).

PR #777, opened 2024-03-12
Picker components

quietbits:
- Updated SDS version to the latest
- Added `stellar-sdk`
- Generic `RadioPicker` component used to create more use-case-specific pickers:
  - `OrderPicker`
  - `IncludeFailedPicker`
  - `AssetPicker`
- `PubKeyPicker`

<details>
<summary>Demo video</summary>
<hr />

[lab-pickers-demo.webm](https://github.com/stellar/laboratory/assets/7346473/24deb24d-7e7b-4a55-986f-e47dafa5fce3)
</details>

PR #779, opened 2024-03-13
add <GenerateKeypair/> component

jeesunikim:
**Summary:**
- added `stellar-sdk`
- using `<ExpandBox/>` from [this PR](https://github.com/stellar/laboratory/pull/777/files) once it's merged
- Using `navigator.clipboard.writeText` for copy to clipboard functionality (tooltip might be nice to display 'copied')

https://github.com/stellar/laboratory/assets/3912060/cb3ba589-8770-4601-a67b-8ba3c52430e8

### DIFF
--- a/src/components/GenerateKeypair.tsx
+++ b/src/components/GenerateKeypair.tsx
@@ -27,13 +27,9 @@ export const GenerateKeypair = () => {
           fieldSize="md"
           label="Public Key"
           value={account.keypair.publicKey}
-          rightElement={
-            <Icon.Copy01
-              onClick={() => {
-                navigator.clipboard.writeText(account.keypair.publicKey);
-              }}
-            />
-          }
+          copyButton={{
+            position: "right",
+          }}
         />
       )}
 
--- a/src/components/GenerateKeypair.tsx
+++ b/src/components/GenerateKeypair.tsx
@@ -36,6 +36,7 @@ export const GenerateKeypair = () => {
           }
         />
       )}
+
       {account.keypair.secretKey && (
         <Input
           readOnly
--- a/src/components/GenerateKeypair.tsx
+++ b/src/components/GenerateKeypair.tsx
@@ -1,25 +1,11 @@
-import { Button, Icon, Input } from "@stellar/design-system";
-import { Keypair } from "stellar-sdk";
+import { Icon, Input } from "@stellar/design-system";
 import { useStore } from "@/store/useStore";
 
 export const GenerateKeypair = () => {
   const { account } = useStore();
 
-  const generateKeypair = () => {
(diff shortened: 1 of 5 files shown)


src/store/createStore.ts
```diff
@@ -11,14 +11,14 @@ export interface Store {
...
-      nestedValue2: number;
+    keypair: {
+      publicKey: string;
+      secretKey: string;
     };
     update: (value: string) => void;
-    updateNested: (nestedVal: {
-      nestedValue1: string;
-      nestedValue2: number;
+    updateKeypair: (nestedVal: {
```
jeesunikim, line 19: let me know if this was intended for something else @quietbits

src/components/GenerateKeypair.tsx
```diff
@@ -0,0 +1,58 @@
...
+          fieldSize="md"
+          label="Public Key"
+          value={account.keypair.publicKey}
+          rightElement={
+            <Icon.Copy01
+              onClick={() => {
+                navigator.clipboard.writeText(account.keypair.publicKey);
+              }}
+            />
+          }
```
quietbits, line 36: `Input` component has a built-in copy button (including copy functionality). :sunglasses: You can see an example of it in `NetworkSelector` component. So we can remove any extra dependencies that were installed.

```suggestion
          copyButton={{
            position: "right",
          }}
```

src/store/createStore.ts
quietbits: No, this was just an example I used for testing. We should remove anything we don't need for the account from there.

src/store/createStore.ts
```diff
@@ -11,14 +11,14 @@ export interface Store {
   // Account
   account: {
     value: string;
-    nestedObject: {
-      nestedValue1: string;
-      nestedValue2: number;
+    keypair: {
+      publicKey: string;
+      secretKey: string;
```
quietbits, line 16: I don't think we should save the secret key in the store. We will only need the public key for the Friendbot, right?

jeesunikim: I thought it would be nice to save it so users can use it when they make transactions on `TESTNET/FUTURENET`. What do you think?

src/components/GenerateKeypair.tsx
```diff
@@ -0,0 +1,37 @@
+import { Icon, Input } from "@stellar/design-system";
+import { useStore } from "@/store/useStore";
+
+export const GenerateKeypair = ({ secretKey }: { secretKey: string }) => {
```
quietbits, line 4: Should we make this component not connected to the store? I mean, pass in public and secret keys for it to display, and that's it.

quietbits (approved):

PR #787, opened 2024-03-14
Run `eslint:recommended` when committing

jeesunikim:
**Summary:** 
- Run `eslint:recommended` when committing ([nextjs doc](https://nextjs.org/docs/app/building-your-application/configuring/eslint#additional-configurations))
- It will only run on modified files during the commit via lint-staged
- `import React from "react";` was needed for `React.ReactNode` type
- it was throwing errors on our placeholder values during `next build` (which means it's working :sparkles:) so I added `eslint-disable` for the following files:
-- `FormElements/AssetPicker.tsx`
-- `FormElements/IncludeFailedPicker.tsx`
-- `FormElements/OrderPicker.tsx`
-- `FormElements/PubKeyPicker.tsx`
-- `RadioPicker/index.tsx`
-- `TabView/index.tsx`
-- `constants/routes.ts`

Example: `Link` and `Icon` are't used in a file and when committing that file, it throws an error

Screenshot 2024-03-14 at 4 12 34 PM


src/components/FormElements/AssetPicker.tsx
```diff
@@ -1,4 +1,6 @@
-import { useState } from "react";
+/* eslint-disable */
```
quietbits, line 1: Let's replace these with more specific rules. Inline if possible, or global for a specific rule.

quietbits (approved):
Thanks for adding these! :tada:

PR #789, opened 2024-03-19
Explore Endpoints: Accounts Params view + template rendering setup

quietbits:
- Rendering/template for all Explore Endpoints with a working example for the Accounts endpoints.
- Submit + streaming + JSON response will be done in the follow-up PRs.
- Other endpoints will be added gradually once we validate that this approach works.


src/components/FormElements/AssetPicker.tsx
```diff
@@ -1,271 +1,145 @@
-import React, { useState } from "react";
+import React from "react";
```
quietbits, line 1: I simplified the `AssetPicker` component always to accept and return an object. Then, the view can handle the response as it fits that view (string or object).

The validation was also moved outside of the form elements to make them more reusable.

src/constants/exploreEndpointsPages.ts
```diff
@@ -0,0 +1,366 @@
...
+            isStreaming?: boolean;
+            custom?: AnyObject;
+          }
+        // TODO: remove once all pages are filled
+        | undefined;
+    }[];
+  }[];
+};
+
+export const EXPLORE_ENDPOINTS_PAGES_HORIZON: ExploreEndpointsPagesProps = {
```
quietbits, line 29: Combined the side navigation data with the data needed to render the pages.

src/constants/formComponentTemplate.tsx
```diff
@@ -0,0 +1,155 @@
...
+};
+
+type FormComponentTemplate = {
+  // eslint-disable-next-line no-unused-vars
+  render: (...args: any[]) => JSX.Element;
+  // eslint-disable-next-line no-unused-vars
+  validate: ((...args: any[]) => any) | null;
+};
+
+export const formComponentTemplate = (
```
quietbits, line 43: All the components are rendered from here. We'll add more as we go.

src/validate/index.ts
```diff
@@ -0,0 +1,11 @@
+import { asset } from "./methods/asset";
```
quietbits, line 1: All the validation methods will go here to make them easier to find.

src/app/(sidebar)/explore-endpoints/[[...pages]]/page.tsx
```diff
@@ -16,72 +16,325 @@ import {
...
+                      updateParams({
+                        [f]: isEmptyObject(sanitizeObject(assetObj || {}))
+                          ? undefined
+                          : JSON.stringify(assetObj),
+                      });
+                      const error = component.validate?.(assetObj, isRequired);
+
+                      if (error) {
+                        setFormError({ ...formError, [f]: error });
+                      } else {
```
Comment (jeesunikim) [inline src/app/(sidebar)/explore-endpoints/[[...pages]]/page.tsx:253]: should this be just `else if()`

src/styles/globals.scss
```diff
@@ -265,7 +269,9 @@
   &__nestedItemsWrapper {
     display: grid;
     grid-template-rows: 0fr;
-    transition: grid-template-rows 0.5s ease-out, margin-top 0.5s ease-out;
+    transition:
+      grid-template-rows 100ms ease-out,
```
jeesunikim, line 273: yay i like it with faster animation

src/store/createStore.ts
```diff
@@ -43,6 +81,36 @@ export const createStore = (options: CreateStoreOptions) =>
               state.account.publicKey = "";
             }),
         },
+        // Explore Endpoints
+        exploreEndpoints: {
+          ...initExploreEndpointState,
+          updateNetwork: (network: Network) =>
```
jeesunikim, line 87: I can't tell from the design. Can `exploreEndpoints.network` be different from the top level `network`?

src/store/createStore.ts
```diff
@@ -16,22 +19,57 @@ export interface Store {
     update: (value: string) => void;
     reset: () => void;
   };
+
+  // Explore Endpoints
+  exploreEndpoints: {
+    network: Network | EmptyObj;
```
jeesunikim, line 25: can we have a different network within `exploreEndpoints` than the top level one?

src/store/createStore.ts
```diff
@@ -51,6 +119,10 @@ export const createStore = (options: CreateStoreOptions) =>
           return {
             network: true,
```
jeesunikim, line 120: `exploreEndpoints` below got me thinking. What do you think of returning `network: { id: true}` because currently our params has all the network information. For `TESTNET`, `FUTURENET`, and `MAINNET`, we do not need all the information since they aren't configurable. For example, `/?||=network$id=testnet&label=Testnet&horizonUrl=https:////horizon-testnet.stellar.org&rpcUrl=https:////soroban-testnet.stellar.org&passphrase=Test%20SDF%20Network%20/;%20September%202015;;`

It might be useful for `Custom`, but I also wonder if we can save that info in `localStorage`.

src/validate/index.ts
jeesunikim: very nice! I'll use this once it's merged for Fund account

src/constants/formComponentTemplate.tsx
jeesunikim: Sounds good. I'll add mine to `FormElements`

jeesunikim: is this file supposed to be in `constants`?

quietbits: I wasn't sure where else to put it. :woman_shrugging: It's static data that we render dynamically. I'm open to suggestions. :pray:

src/store/createStore.ts
quietbits: So, I made it this way considering that we can save endpoints in the future. And we'll need to include the network with that endpoint to make sure it functions properly.

It also helps to clear the current endpoint data in the store if the top level network changes.

src/store/createStore.ts
quietbits: They should be in sync most of the time, except when the top level changes. If they are different, we know that the top level network changed, so we need to clear the current endpoint data from the store.

jeesunikim (approved):

src/constants/formComponentTemplate.tsx
jeesunikim: it feels more like a component to me because it imports `components`.

quietbits: :+1: I'll move it to `components` in the next PR.

PR #794, opened 2024-03-25
[Account] create and parse muxed account

jeesunikim:
**Summary:**
- [Figma Design](https://www.figma.com/file/yuikaFOPmJifyAfS90ER6T/Laboratory?type=design&node-id=351-2164&mode=design&t=dje7U5RMjnWeQ7Se-0)
- Added the pages for `create` and `parse`
- Created FormElements `MuxedIdPicker`, `MuxedKeyPicker`, and `PubKeyPicker`
- Added `readOnly` props for `PubKeyPicker` because the muxed account result section displays `PubKeyPicker` that shouldn't be edited
- `helpers/muxedAccount` added
- Deleted `helpers/validatePublicKey` in favor of `validate` method
- `createStore` updated to include muxed account information

**Screenshots:**
- Successful Create Mux Account

create-mux

- Error state for Create Mux Account

create-mux-accountid-error

- Failed error message with Parse Mux Account Page

parse-mux


src/app/(sidebar)/account/fund/page.tsx
```diff
@@ -52,8 +53,8 @@ export default function FundAccount() {
             onChange={(e) => {
               setGeneratedPublicKey(e.target.value);
 
-              const error = validatePublicKey(e.target.value);
```
jeesunikim, line 55: using the `validate` method

src/components/FormElements/PubKeyPicker.tsx
```diff
@@ -7,34 +7,33 @@ interface PubKeyPickerProps extends Omit<InputProps, "fieldSize"> {
   label: string;
   labelSuffix?: string | React.ReactNode;
   placeholder?: string;
+  readOnly?: boolean;
```
jeesunikim, line 10: I added this because muxed account result displays `pubkeypicker`, but with `readOnly`

src/store/createStore.ts
```diff
@@ -15,8 +15,14 @@ export interface Store {
   // Account
   account: {
     publicKey: string;
+    generatedMuxedAccount: MuxedAccount;
+    parsedMuxedAccount: MuxedAccount;
     // eslint-disable-next-line no-unused-vars
-    update: (value: string) => void;
+    updatePublicKey: (value: string) => void;
+    // eslint-disable-next-line no-unused-vars
```
jeesunikim, line 22: going to look into optimizing this next (and making linting better in general)

src/app/(sidebar)/account/muxed-create/page.tsx
```diff
@@ -1,5 +1,194 @@
...
+                size="md"
+                variant={"secondary"}
+                onClick={generateMuxedAccount}
+              >
+                Create
+              </Button>
+            </div>
+          </div>
+
+          <ExpandBox
```
quietbits, line 129: We might need to put the `ExpandBox` inside another container because it's creating extra space at the bottom where it is now. We can do something similar to Create Account.

image

src/components/FormElements/MuxedIdPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+  labelSuffix?: string | React.ReactNode;
+  placeholder?: string;
+  readOnly?: boolean;
+  value: string;
+  error: string | undefined;
+  // eslint-disable-next-line no-unused-vars
+  onChange?: (e: React.ChangeEvent<any>) => void;
+}
+
+export const MuxedIdPicker = ({
```
quietbits, line 17: :thinking: Do we need this as a separate component? We could use an `Input` instead, passing in a placeholder we need here. What do you think?

src/components/FormElements/MuxedKeyPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+  labelSuffix?: string | React.ReactNode;
+  placeholder?: string;
+  readOnly?: boolean;
+  value: string;
+  error: string | undefined;
+  // eslint-disable-next-line no-unused-vars
+  onChange?: (e: React.ChangeEvent<any>) => void;
+}
+
+export const MuxedKeyPicker = ({
```
quietbits, line 17: :thinking: The same question here. It seems so similar to `PubKeyPicker` that we might use that instead.

src/helpers/muxedAccount.ts
```diff
@@ -0,0 +1,62 @@
...
+      return {
+        muxedAddress: muxedAccount.accountId(),
+        error: "",
+      };
+    } catch (e: any) {
+      return {
+        muxedAddress: "",
+        error: `Something went wrong. ${e.toString()}`,
+      };
+    }
```
quietbits, line 27: If we want to return the same object in both cases, we could optimize the code a bit.

```suggestion
    let muxedAddress = "";
    let error = "";
    
    try {
      const muxedAccount = new MuxedAccount(
        new Account(baseAddress, "0"),
        muxedAccountId,
      );
      
      muxedAddress = muxedAccount.accountId();
    } catch (e: any) {
      error = `Something went wrong. ${e.toString()}`;
    }
    
    return { muxedAddress, error }
```

src/helpers/muxedAccount.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+  parse: ({
+    muxedAddress,
+  }: {
+    muxedAddress: string;
+  }): Partial<MuxedAccountFieldType> => {
+    try {
```
quietbits, line 34: We could optimize this similarly to the above code.

src/store/createStore.ts
```diff
@@ -72,7 +78,31 @@ export const createStore = (options: CreateStoreOptions) =>
...
+          generatedMuxedAccount: {
+            id: "",
+            baseAddress: "",
+            muxedAddress: "",
+          },
+          parsedMuxedAccount: {
+            id: "",
+            baseAddress: "",
+            muxedAddress: "",
+          },
```
quietbits, line 90: We could make these optional with initial values `undefined`.

src/app/(sidebar)/account/muxed-create/page.tsx
jeesunikim: eagle eyes!!!

jeesunikim: same thing happened in a different container. Create Account is using `margin-top` inside the div. Not sure how I can achieve this using `gap`. it seems like they don't collapse even when a div is not there.

quietbits: Ooh, I missed the `margin-top` in Create Account. We shouldn't use margins for layout spacing. I'll create a component that will make this easier.

src/app/(sidebar)/account/muxed-create/page.tsx
```diff
@@ -1,5 +1,213 @@
...
+              id="muxed-account-address-result"
+              label="Muxed Account M Address"
+              value={account.generatedMuxedAccount.muxedAddress || ""}
+              error=""
+              readOnly={true}
+              copyButton={{
+                position: "right",
+              }}
+            />
+          </div>
```
quietbits, line 187: How about we move this to its own component with props for G-address, M-address, and ID? It's a static component we can use to create and process muxed accounts. The forms are identical on both screens.

<details>
<summary>Screenshots</summary>

image

image
</details>

src/store/createStore.ts
```diff
@@ -72,7 +84,47 @@ export const createStore = (options: CreateStoreOptions) =>
...
+            id: undefined,
+            baseAddress: undefined,
+            muxedAddress: undefined,
+          },
+          parsedMuxedAccountInput: undefined,
+          parsedMuxedAccount: {
+            id: undefined,
+            baseAddress: undefined,
+            muxedAddress: undefined,
+          },
```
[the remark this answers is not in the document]
jeesunikim: how about an empty object?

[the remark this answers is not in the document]
quietbits: Sure, that works.

src/components/MuxedAccountResult.tsx
```diff
@@ -0,0 +1,47 @@
+import { Input } from "@stellar/design-system";
+import { useStore } from "@/store/useStore";
+
+import { PubKeyPicker } from "@/components/FormElements/PubKeyPicker";
+
+export const MuxedAccountResult = () => {
```
quietbits, line 6: The form doesn't work on the Create Muxed Account page. It might be better to make this component "dumb" and not connect it to the store. It can accept all three params needed and show what is passed. What do you think?

image

jeesunikim: good point!

quietbits (approved):

PR #800, opened 2024-03-27
Box component + tweak ExpandBox

quietbits:
- Box component to use anywhere a container with a row gap is needed (`flex-direction: column`)
- Update `ExpandBox` to handle top offset (when collapsed, it was creating a double gap)

### DIFF
--- a/src/app/(sidebar)/account/create/page.tsx
+++ b/src/app/(sidebar)/account/create/page.tsx
@@ -52,13 +52,14 @@ export default function CreateAccount() {
               Fund account with Friendbot
             </Button>
           </div>
+
+          <ExpandBox isExpanded={Boolean(account.publicKey)} offsetTop="xl">
+            <GenerateKeypair
+              publicKey={account.publicKey}
+              secretKey={secretKey}
+            />
+          </ExpandBox>
         </div>
-        <ExpandBox isExpanded={Boolean(account.publicKey)}>
-          <GenerateKeypair
-            publicKey={account.publicKey}
-            secretKey={secretKey}
-          />
-        </ExpandBox>
       </Card>
     </div>
   );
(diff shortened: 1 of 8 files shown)


src/app/(sidebar)/account/create/page.tsx
```diff
@@ -52,13 +52,14 @@ export default function CreateAccount() {
               Fund account with Friendbot
             </Button>
           </div>
+
+          <ExpandBox isExpanded={Boolean(account.publicKey)} offsetTop="xl">
+            <GenerateKeypair
+              publicKey={account.publicKey}
+              secretKey={secretKey}
+            />
+          </ExpandBox>
```
quietbits, line 61: We can update the Fund Account page and other places if this looks good.

src/components/layout/Box/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const Box = ({
+  gap,
+  children,
+  customValue,
+  addlClassName,
+}: (
+  | { gap: "xs" | "sm" | "md" | "lg" | "xl" | "xxl"; customValue?: undefined }
+  | { gap: "custom"; customValue: string }
+) & { children: React.ReactElement; addlClassName?: string }) => {
```
jeesunikim, line 12: I think `l` is a typo. It shold be `addClassName`?

jeesunikim (approved):
added a comment on typo but looks good. let's merge it!

src/components/layout/Box/index.tsx
quietbits: It's actually "additional classes". We went with that convention in the SDS ([example](https://github.com/stellar/stellar-design-system/blob/sds-v3/%40stellar/design-system/src/components/Link/index.tsx#L27)). Thanks for checking!

PR #807, opened 2024-04-02
Explore Endpoints: pages part 2

quietbits:
Added pages:
- Payments
- Transactions
- Order book
- Trade aggregations
- Trades

Other updates:
- Upgraded SDS to the latest version
- Fixed records rendering in `PrettyJson`

### DIFF
--- a/src/app/(sidebar)/explore-endpoints/[[...pages]]/page.tsx
+++ b/src/app/(sidebar)/explore-endpoints/[[...pages]]/page.tsx
@@ -86,7 +86,7 @@ export default function ExploreEndpoints() {
         const param = p.match(REGEX_TEMPLATE_PATH_PARAM_VALUE)?.[0];
 
         if (param) {
-          return urlPathParamArr.push(param);
+          urlPathParamArr.push(param);
         }
       });
 


src/validate/methods/xdr.ts
```diff
@@ -0,0 +1,7 @@
+export const xdr = (value: string) => {
```
jeesunikim, line 1: should we name this to `validateBase64`? In our current lab, we have two methods: [validateBase64](https://github.com/stellar/laboratory/blob/ed2094d980bdeb967e5c636ea6053e31ab9c8773/src/helpers/validateBase64.ts#L4) and [validateTxXdr](https://github.com/stellar/laboratory/blob/ed2094d980bdeb967e5c636ea6053e31ab9c8773/src/helpers/validateTxXdr.ts#L5)

src/components/FormElements/XdrPicker.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface XdrPickerProps extends Omit<TextareaProps, "fieldSize"> {
+  id: string;
+  fieldSize?: "sm" | "md" | "lg";
+  labelSuffix?: string | React.ReactNode;
+  label: string;
+  value: string;
+  placeholder?: string;
+  error: string | undefined;
+  // eslint-disable-next-line no-unused-vars
```
jeesunikim, line 12: you can remove this! 🎉

quietbits: Will do!

src/constants/exploreEndpointsPages.ts
```diff
@@ -461,22 +489,59 @@ export const EXPLORE_ENDPOINTS_PAGES_HORIZON: ExploreEndpointsPagesProps = {
         {
           route: Routes.EXPLORE_ENDPOINTS_PAYMENTS,
           label: "All Payments",
-          form: undefined,
+          form: {
+            docsUrl:
+              "https://developers.stellar.org/network/horizon/resources/list-all-payments",
+            docsLabel: "payments",
+            requestMethod: "GET",
```
jeesunikim, line 496: i like using enum for some of the repeated constants. just a thought! def don't have to make a change in this pr.

jeesunikim (approved):
it works well!

src/constants/exploreEndpointsPages.ts
quietbits: We're using "GET" as a string here because it's coming from the type (`requestMethod: "GET" | "POST";`). I find it more convenient to use it this way in this case and also easier to validate.

PR #808, opened 2024-04-04
adding tests for create and fund account

jeesunikim:
**Summary:**
- Added tests for: `create account`, `fund account`, `create muxed account`, `parse muxed account`

### DIFF
--- a/src/app/(sidebar)/account/fund/page.tsx
+++ b/src/app/(sidebar)/account/fund/page.tsx
@@ -60,7 +60,7 @@ export default function FundAccount() {
             error={inlineErrorMessage}
           />
 
-          <div className="Account__CTA">
+          <div className="Account__CTA" data-testid="fundAccount-buttons">
             <Button
               disabled={!generatedPublicKey || Boolean(inlineErrorMessage)}
               size="md"
--- a/src/app/(sidebar)/account/fund/page.tsx
+++ b/src/app/(sidebar)/account/fund/page.tsx
@@ -2,8 +2,11 @@
 
 import { useEffect, useState } from "react";
 import { Alert, Card, Input, Text, Button } from "@stellar/design-system";
+import Link from "next/link";
+import { Routes } from "@/constants/routes";
 
 import { shortenStellarAddress } from "@/helpers/shortenStellarAddress";
+import { useIsTestingNetwork } from "@/hooks/useIsTestingNetwork";
 import { useFriendBot } from "@/query/useFriendBot";
 import { useStore } from "@/store/useStore";
 
@@ -18,6 +21,8 @@ export default function FundAccount() {
   const [generatedPublicKey, setGeneratedPublicKey] = useState<string>("");
   const [inlineErrorMessage, setInlineErrorMessage] = useState<string>("");
 
+  const IS_TESTING_NETWORK = useIsTestingNetwork();
+
   const { error, isError, isLoading, isSuccess, refetch, isFetchedAfterMount } =
     useFriendBot({
       network: network.id,
@@ -30,6 +35,15 @@ export default function FundAccount() {
     }
   }, [isError, isSuccess]);
 
+  if (!IS_TESTING_NETWORK) {
+    return (
+      <div className="Account">
+        <h2>Not Found</h2>
+        <p>Could not find requested resource</p>
+        <Link href={Routes.ROOT}>Return Home</Link>
+      </div>
+    );
+  }
   return (
     <div className="Account">
       <Card>
--- a/src/app/(sidebar)/account/layout.tsx
+++ b/src/app/(sidebar)/account/layout.tsx
@@ -4,33 +4,50 @@ import React from "react";
 
 import { LayoutSidebarContent } from "@/components/layout/LayoutSidebarContent";
 import { Routes } from "@/constants/routes";
+import { useIsTestingNetwork } from "@/hooks/useIsTestingNetwork";
+
+// Includes navigation items that should ONLY display
+// in TESTNET or FUTURENET
+const DEFAULT_ACCOUNT_NAV_ITEMS = [
+  {
+    route: Routes.ACCOUNT_CREATE,
+    label: "Create Account",
+  },
+  {
+    route: Routes.ACCOUNT_FUND,
+    label: "Fund Account",
+    isTestnetOnly: true,
+  },
+  {
+    route: Routes.ACCOUNT_CREATE_MUXED,
+    label: "Create Muxed Account",
+  },
+  {
+    route: Routes.ACCOUNT_PARSE_MUXED,
+    label: "Parse Muxed Account",
+  },
+];
 
 export default function AccountTemplate({
   children,
 }: {
   children: React.ReactNode;
 }) {
+  const IS_TESTING_NETWORK = useIsTestingNetwork();
+
+  // Filtered navigation items for MAINNET
+  const filteredNavItems = DEFAULT_ACCOUNT_NAV_ITEMS.filter(
+    ({ isTestnetOnly }) => !isTestnetOnly,
+  );
+
+  const navItems = IS_TESTING_NETWORK
+    ? DEFAULT_ACCOUNT_NAV_ITEMS
+    : filteredNavItems;
+
   return (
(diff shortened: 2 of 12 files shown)


tests/fundAccountPage.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      .getByText("Get lumens");
+
+    // Type in an invalid string in 'Public Key' input field
+    await publicKeyInput.fill(publicKey);
+
+    await expect(publicKeyInput).toHaveAttribute("aria-invalid", "false");
+    await expect(getLumenButton).toBeEnabled();
+
+    // Mock the friendbot api call
+    await page.route("**/*", async (route, request) => {
```
jeesunikim, line 97: mocking friendbot api call

tests/fundAccountPage.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    // Type in an invalid string in 'Public Key' input field
+    await publicKeyInput.fill(
+      "GBX6W44ISK6XTBDDJ5ND6KTJHLYEYHMR4SDG635NRARYVJ3G2YXGDT6Y",
+    );
+
+    await expect(publicKeyInput).toHaveAttribute("aria-invalid", "false");
+    await expect(getLumenButton).toBeEnabled();
+
+    await getLumenButton.click();
```
jeesunikim, line 141: I could also mock the api call to return an error vs. clicking the button twice

tests/createAccountPage.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      page.locator("input[id='generate-keypair-secretkey']"),
+    ).toHaveValue(/^S/);
+  });
+
+  test("Test 'Fund account' button", async ({ page }) => {
+    await page
+      .getByRole("button", { name: "Fund account with Friendbot" })
+      .click();
+
+    await page.goto("http://localhost:3000/account/fund");
```
quietbits, line 37: The Fund account button should take us to the next page, right? I think we want to check if that happened using [`toHaveURL` method](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-url).

tests/createMuxedAccountPage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const createButton = page.getByRole("button").getByText("Create");
+
+    await expect(publicKeyInput).toHaveAttribute("aria-invalid", "false");
+    await expect(muxedAccountIdInput).toHaveAttribute("aria-invalid", "false");
+    await expect(createButton).toBeEnabled();
+
+    await createButton.click();
+
+    await expect(page.getByTestId("createAccount-success")).toBeVisible();
```
quietbits, line 61: We could also check that the correct muxed account address was created.

tests/fundAccountPage.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    // Select 'Futurenet' in the network dropdown list
+    await expect(submitButton).toHaveText("Switch to Futurenet");
+    await expect(submitButton).toBeEnabled();
+
+    // Click 'Switch to Futurenet' button
+    await submitButton.click();
+
+    await expect(page.getByTestId("networkSelector-button")).toHaveText(
+      "Futurenet",
+    );
```
quietbits, line 43: We could simplify the network selection by setting a URL. We test Network selection separately, so we can assume that it's working here.

```suggestion
    await page.goto("http://localhost:3000/account/muxed-create?||=network$id=futurenet&label=Futurenet&horizonUrl=https:////horizon-futurenet.stellar.org&rpcUrl=https:////rpc-futurenet.stellar.org&passphrase=Test%20SDF%20Future%20Network%20/;%20October%202022;;")
`

File: tests/fundAccountPage.test.ts [hunk 6]
```diff
@@ -0,0 +1,148 @@
...
+    await expect(page.locator("h1")).toHaveText(
+      "Friendbot: fund a futurenet network account",
+    );
+  });
+
```
quietbits, line 48: Should we also test that friendbot is disabled for mainnet?

tests/fundAccountPage.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  }) => {
+    // Get a new public key
+    const publicKey =
+      "GDVOT2ALMUF3G54RBHNJUEV6LOAZCQQCARHEVNUPKGMVPWFC4PFN33QR";
+    const publicKeyInput = page.locator("#generate-keypair-publickey");
+    const getLumenButton = page
+      .getByTestId("fundAccount-buttons")
+      .getByText("Get lumens");
+
+    // Type in an invalid string in 'Public Key' input field
```
quietbits, line 90:
```suggestion
    // Type in an valid string in 'Public Key' input field
```

tests/fundAccountPage.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  test("Gets an error when submitting 'Get lumens' button twice with a valid public key", async ({
+    page,
+  }) => {
+    const publicKeyInput = page.locator("#generate-keypair-publickey");
+    const getLumenButton = page
+      .getByTestId("fundAccount-buttons")
+      .getByText("Get lumens");
+
+    // Type in an invalid string in 'Public Key' input field
```
quietbits, line 133:
```suggestion
    // Type in an valid string in 'Public Key' input field
```

tests/fundAccountPage.test.ts
quietbits: Yes, let's mock an error response to control this better.

tests/fundAccountPage.test.ts
quietbits: Nice! Could we use full route string here `*/**/?addr=GDVOT2ALMUF3G54RBHNJUEV6LOAZCQQCARHEVNUPKGMVPWFC4PFN33QR` and remove the `if` check?

tests/createMuxedAccountPage.test.ts
jeesunikim: I can use mainnet muxed account value since testnet one will reset.

src/app/(sidebar)/account/fund/page.tsx
```diff
@@ -30,6 +35,15 @@ export default function FundAccount() {
     }
   }, [isError, isSuccess]);
 
+  if (!IS_TESTING_NETWORK) {
```
jeesunikim, line 38: I also tried out with adding a [middleware.ts](https://nextjs.org/docs/app/building-your-application/routing/middleware) to reroute to `404` if `nextUrl` includes `mainnet` in params. However, it doesn't always work in all cases. ex. user directly types in `/account/fund` in a browser which doesn't populate params until after user lands on the page.

another method i tried is `router.push` on the client side but that produced a glitch which wasn't a good ux.

so I came up with this.

quietbits:
Not related to tests, the generated public address wasn't cleared when I switched from Testnet to Mainnet.

src/app/(sidebar)/account/fund/page.tsx
jeesunikim: I'll update the copy in the next pr when I make the overall changes following the updates

jeesunikim:
> Not related to tests, the generated public address wasn't cleared when I switched from Testnet to Mainnet.

added this feedback in [the new issue ticket](https://github.com/stellar/laboratory/issues/819) created by Jane

src/app/(sidebar)/account/layout.tsx
```diff
@@ -4,33 +4,50 @@ import React from "react";
...
+// in TESTNET or FUTURENET
+const DEFAULT_ACCOUNT_NAV_ITEMS = [
+  {
+    route: Routes.ACCOUNT_CREATE,
+    label: "Create Account",
+  },
+  {
+    route: Routes.ACCOUNT_FUND,
+    label: "Fund Account",
+    isTestnetOnly: true,
```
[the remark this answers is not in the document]
quietbits: We won't need this extra logic for the new update where we're not hiding the nav, right? Just a note for the next PR. :nerd_face:

quietbits (approved):

PR #809, opened 2024-04-04
Explore Endpoints: pages part 3

quietbits:
Added pages:
- Liquidity pools
- Paths


src/components/formComponentTemplate.tsx
```diff
@@ -421,6 +509,69 @@ export const formComponentTemplate = (
...
+        ),
+        validate: validate.amount,
+      };
+    case "source_asset":
+      return {
+        render: (templ: TemplateRenderAssetProps) => (
+          <AssetPicker
+            key={id}
+            assetInput={custom?.assetInput}
+            id="destination_asset"
```
jeesunikim, line 548: I think this is supposed to be either `{id}` or `source_asset`

src/components/formComponentTemplate.tsx
```diff
@@ -224,6 +232,69 @@ export const formComponentTemplate = (
...
+        ),
+        validate: validate.amount,
+      };
+    case "destination_asset":
+      return {
+        render: (templ: TemplateRenderAssetProps) => (
+          <AssetPicker
+            key={id}
+            assetInput={custom?.assetInput}
+            id="destination_asset"
```
jeesunikim, line 271: can this be replaced with `id={id}`?

src/components/formComponentTemplate.tsx
quietbits: Yeah, I'll update to use `id` in components where possible to avoid typos like this.

src/validate/methods/amount.ts
```diff
@@ -0,0 +1,11 @@
+export const amount = (value: string) => {
```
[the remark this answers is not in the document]
quietbits: I think it's fine if it's working. We're not sure if this was done intentionally. :woman_shrugging: Horizon seems to accept leading zeroes.

image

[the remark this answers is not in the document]
jeesunikim: I'll add to the `nice to have` list since horizon removes the leading zero i think frontend should do the same

jeesunikim (approved):
👏

PR #826, opened 2024-04-12
[Sign TX] Input Overview View

jeesunikim:
**Summary:**
- Only the view parts for the `default import` and `overview` view once tx is successfully imported


src/components/FormElements/PubKeyPicker.tsx
```diff
@@ -21,6 +21,7 @@ export const PubKeyPicker = ({
   placeholder = "Ex: GCEXAMPLE5HWNK4AYSTEQ4UWDKHTCKADVS2AHF3UI2ZMO3DPUSM6Q4UG",
   value,
   error,
+  readOnly,
```
jeesunikim, line 24: Ideally I'd like to have a copy by default on click functionality. I didn't want to add too much to the original component though. Should I just create its own component? @quietbits

src/components/FormElements/XdrPicker.tsx
```diff
@@ -5,11 +5,13 @@ interface XdrPickerProps extends Omit<TextareaProps, "fieldSize"> {
...
   fieldSize?: "sm" | "md" | "lg";
   labelSuffix?: string | React.ReactNode;
-  label: string;
+  label: string | React.ReactNode;
   value: string;
   placeholder?: string;
-  error: string | undefined;
-  onChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => void;
+  error?: string | undefined;
+  note?: string | React.ReactNode;
```
jeesunikim, line 12: using note as a success message

src/validate/methods/xdr.ts
```diff
@@ -1,7 +1,37 @@
-export const xdr = (value: string) => {
+import { trim } from "@/helpers/trim";
+import { xdr as stellarXDR } from "@stellar/stellar-sdk";
+
+const validateBase64 = (value: string) => {
```
jeesunikim, line 4: I can add this to its own file! but I think so far, this is just used within checking xdr

src/app/(sidebar)/transaction/sign/page.tsx
```diff
@@ -1,5 +1,198 @@
...
+            one signature if there are multiple source accounts or signing keys.
+          </Text>
+        </Alert>
+      </>
+    );
+  };
+
+  const renderOverviewView = () => {
+    if (!tx) {
+      return;
```
quietbits, line 122: React might not be happy if we return `undefined`

```suggestion
      return null;
```

src/app/(sidebar)/transaction/sign/page.tsx
```diff
@@ -1,5 +1,198 @@
...
+      mergedFields = [...REQUIRED_FIELDS, ...TX_FIELDS(tx)];
+    }
+
+    return (
+      <>
+        <Card>
+          <div className="SignTx__FieldViewer">
+            {mergedFields.map((field) => {
+              const className =
+                field.value.toString().length >= MIN_LENGTH_FOR_FULL_WIDTH_FIELD
```
quietbits, line 154: Are we always guaranteed to have `field.value`? Just making sure we don't need to chain these like optional with `?`.

src/components/FormElements/XdrPicker.tsx
quietbits: :+1: We might need to add the `success` param in SDS.

src/constants/signTransactionPage.ts
```diff
@@ -0,0 +1,62 @@
...
+    label: "Fee source account",
+    value: feeBumpTx.feeSource,
+  },
+  {
+    label: "Transaction Fee (stroops)",
+    value: feeBumpTx.fee,
+  },
+  {
+    label: "Number of existing signatures",
+    value: feeBumpTx.signatures.length,
```
quietbits, line 14: I think we might need to make these optional in case the provided XDR doesn't have the values. What do you think?

```suggestion
    value: feeBumpTx.signatures?.length,
```

src/validate/methods/xdr.ts
quietbits: If it's only used here right now, let's keep it here. :+1:

src/app/(sidebar)/transaction/sign/page.tsx
jeesunikim: yes!

src/components/FormElements/XdrPicker.tsx
jeesunikim: done!

Issue #819, opened 2024-04-09
[Create Account] Fund account with Friend button behavior


janewang:
### What problem does your feature solve?

<img width="724" alt="Screenshot 2024-04-09 at 10 24 01 AM" src="https://github.com/stellar/laboratory/assets/734336/27d8c14b-7051-478a-b58f-cb5428b8e35d">

On the Create Account screen:
- [x] If a user creates an account, the [Fund account with Friendbot] button should fund the account on the screen
- [x] If a user has not created an account, the [Fund account with Friendbot] button should not be clickable and be in a disabled state
- [x] If a user is on mainnet, this button should not be available
- [x] If a user is on a custom work, this button should not be available unless they have inputted Horizon endpint
- [x] (added by Jeesun) clear the the generated public address wasn't cleared when switching network ([thanks iveta](https://github.com/stellar/laboratory/pull/808))

PR #827, opened 2024-04-15
[Create Account] Fund account with Friend button behavior

jeesunikim:
Addresses the following from the issue #819:
- [x] If a user creates an account, the [Fund account with Friendbot] button should fund the account on the screen
- [x] If a user has not created an account, the [Fund account with Friendbot] button should not be clickable and be in a disabled state
- [x] If a user is on mainnet, this button should not be available
- [x] If a user is on a custom work, this button should not be available unless they have inputted Horizon endpint
- [x] (added by Jeesun) clear the the generated public address wasn't cleared when switching network ([thanks iveta](https://github.com/stellar/laboratory/pull/808))

### DIFF
--- a/src/app/(sidebar)/account/create/page.tsx
+++ b/src/app/(sidebar)/account/create/page.tsx
@@ -1,18 +1,19 @@
 "use client";
 
 import { useCallback, useEffect, useState } from "react";
-import { Alert, Card, Text, Button } from "@stellar/design-system";
+import { Card, Text, Button } from "@stellar/design-system";
 import { Keypair } from "@stellar/stellar-sdk";
 
 import { useStore } from "@/store/useStore";
 import { useFriendBot } from "@/query/useFriendBot";
 import { useQueryClient } from "@tanstack/react-query";
 
-import { shortenStellarAddress } from "@/helpers/shortenStellarAddress";
 import { useIsTestingNetwork } from "@/hooks/useIsTestingNetwork";
 
 import { GenerateKeypair } from "@/components/GenerateKeypair";
 import { ExpandBox } from "@/components/ExpandBox";
+import { SuccessMsg } from "@/components/FriendBot/SuccessMsg";
+import { ErrorMsg } from "@/components/FriendBot/ErrorMsg";
 
 import "../styles.scss";
 
@@ -62,6 +63,8 @@ export default function CreateAccount() {
   }, [account.registeredNetwork, network.id]);
 
   const generateKeypair = () => {
+    resetStates();
+
     const keypair = Keypair.random();
 
     if (IS_TESTING_NETWORK) {
@@ -98,6 +101,7 @@ export default function CreateAccount() {
                 size="md"
                 disabled={!account.publicKey || isLoading}
                 variant="tertiary"
+                isLoading={isLoading}
                 onClick={() => refetch()}
                 data-testid="fundAccount-button"
               >
@@ -119,34 +123,22 @@ export default function CreateAccount() {
         </div>
       </Card>
 
-      {showAlert && isFetchedAfterMount && isSuccess && account.publicKey && (
-        <Alert
-          placement="inline"
-          variant="success"
-          actionLabel="View on stellar.expert"
-          actionLink={`https://stellar.expert/explorer/${network.id}/account/${account.publicKey}`}
-          onClose={() => {
-            setShowAlert(false);
-          }}
-          title={`Successfully funded ${shortenStellarAddress(account.publicKey)} on 
-          ${network.id}`}
-        >
-          {""}
-        </Alert>
-      )}
+      <SuccessMsg
+        isVisible={Boolean(
+          showAlert && isFetchedAfterMount && isSuccess && account.publicKey,
+        )}
+        onClose={() => {
+          setShowAlert(false);
+        }}
+      />
 
-      {showAlert && isFetchedAfterMount && isError && (
-        <Alert
-          placement="inline"
-          variant="error"
-          onClose={() => {
-            setShowAlert(false);
-          }}
-          title={error?.message}
-        >
-          {""}
-        </Alert>
-      )}
+      <ErrorMsg
+        isVisible={Boolean(showAlert && isFetchedAfterMount && isError)}
+        errorMsg={error?.message}
+        onClose={() => {
+          setShowAlert(false);
+        }}
+      />
     </div>
   );
 }
--- a/src/app/(sidebar)/account/create/page.tsx
+++ b/src/app/(sidebar)/account/create/page.tsx
@@ -132,6 +132,7 @@ export default function CreateAccount() {
       </Card>
 
       <SuccessMsg
+        publicKey={account.publicKey!}
         isVisible={Boolean(showAlert && isSuccess && account.publicKey)}
         onClose={() => {
           setShowAlert(false);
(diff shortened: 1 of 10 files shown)


src/app/(sidebar)/account/create/page.tsx
```diff
@@ -1,29 +1,75 @@
...
+  );
+
+  const resetStates = useCallback(() => {
+    account.reset();
+    resetQuery();
+  }, [resetQuery]);
+
+  const { error, isError, isLoading, isSuccess, refetch, isFetchedAfterMount } =
+    useFriendBot({
+      network,
```
jeesunikim, line 44: passing `network` so `useQuery` can call `network.horizonUrl/friendbot` dynamically vs. direct inputs

src/app/(sidebar)/account/layout.tsx
```diff
@@ -16,7 +16,7 @@ export default function AccountTemplate({
         navItems: [
           {
             route: Routes.ACCOUNT_CREATE,
-            label: "Create Account",
+            label: "Create Account Keypair",
```
jeesunikim, line 19: following the design update

src/app/(sidebar)/account/create/page.tsx
```diff
@@ -46,27 +92,61 @@ export default function CreateAccount() {
...
 
-          <ExpandBox isExpanded={Boolean(account.publicKey)} offsetTop="xl">
-            <div className="Account__result">
-              <GenerateKeypair
-                publicKey={account.publicKey}
-                secretKey={secretKey}
-              />
-            </div>
-          </ExpandBox>
+          {Boolean(account.publicKey) && (
```
jeesunikim, line 109: sometimes empty grey box appeared when switching network so adding an extra boolean `Boolean(account.publicKey)`

src/store/createStore.ts
```diff
@@ -13,12 +13,14 @@ export interface Store {
 
   // Account
   account: {
-    publicKey: string;
+    publicKey: string | undefined;
+    secretKey: string | undefined;
+    registeredNetwork: Network | EmptyObj;
```
jeesunikim, line 18: added this to compare the current switched network to the previous network that the account was created in

src/app/(sidebar)/account/create/page.tsx
```diff
@@ -46,27 +92,61 @@ export default function CreateAccount() {
...
-            {IS_TESTING_NETWORK ? (
+
+            {IS_TESTING_NETWORK || IS_CUSTOM_NETWORK_WITH_HORIZON ? (
               <Button
                 size="md"
+                disabled={!account.publicKey || isLoading}
                 variant="tertiary"
-                onClick={() => router.push(Routes.ACCOUNT_FUND)}
+                onClick={() => refetch()}
+                data-testid="fundAccount-button"
```
quietbits, line 102: We should have a loading state on the button when funding is in progress.

src/app/(sidebar)/account/create/page.tsx
```diff
@@ -46,27 +92,61 @@ export default function CreateAccount() {
             <Button size="md" variant="secondary" onClick={generateKeypair}>
```
quietbits, line 92: The previous state wasn't reset when clicking "Generate keypair` after successfully funding an account.

Should we also reset the state when user leaves this page? It feels a bit off to see the generated keypair when returning to the page.

src/app/(sidebar)/account/create/page.tsx
```diff
@@ -46,27 +92,61 @@ export default function CreateAccount() {
...
+          )}
         </div>
       </Card>
+
+      {showAlert && isFetchedAfterMount && isSuccess && account.publicKey && (
+        <Alert
+          placement="inline"
+          variant="success"
+          actionLabel="View on stellar.expert"
+          actionLink={`https://stellar.expert/explorer/${network.id}/account/${account.publicKey}`}
```
quietbits, line 127: Should we hide the StellarExpert link for the `futurenet` and `custom` networks?

src/app/(sidebar)/account/fund/page.tsx
```diff
@@ -104,10 +113,10 @@ export default function FundAccount() {
         </div>
       </Card>
 
-      {showAlert && isFetchedAfterMount && isSuccess && (
+      {showAlert && isFetchedAfterMount && isSuccess && account.publicKey && (
```
quietbits, line 116: We could create a Friendbot success and error message component to be shared between the Create and Fund pages. That way we would handle all the logic only once. What do you think?

src/query/useFriendBot.ts
```diff
@@ -1,24 +1,26 @@
...
     queryKey: ["friendBot"],
     queryFn: async () => {
+      if (!network.horizonUrl) {
+        throw new Error(`Please use a network that supports Horizon`);
+      }
+
       try {
-        const response = await fetch(`${friendbotURL}/?addr=${publicKey}`);
+        const response = await fetch(
+          `${network.horizonUrl}/friendbot?addr=${publicKey}`,
```
quietbits, line 22: I know we don't do it in the current Lab, but should we show a better error message when we have more info? For example, when funding an account that is already funded.

We get that info in the response
image

src/store/createStore.ts
```diff
@@ -127,13 +135,16 @@ export const createStore = (options: CreateStoreOptions) =>
...
               };
             }),
-          updatePublicKey: (value: string) =>
+          updateKeypair: (publicKey: string, secretKey?: string) =>
             set((state) => {
-              state.account.publicKey = value;
+              const defaultSecretKey = secretKey || "";
+              state.account.publicKey = publicKey;
+              state.account.secretKey = defaultSecretKey;
+              state.account.registeredNetwork = state.network;
```
quietbits, line 143: We can simplify the secret key assignment inline.

```suggestion
              state.account.publicKey = publicKey;
              state.account.secretKey = secretKey || "";
              state.account.registeredNetwork = state.network;
```

src/app/(sidebar)/account/create/page.tsx
jeesunikim: i think that makes sense if they save the keys to localStorage. the current lab currently saves the state. i personally prefer what you suggest since the code would be much simpler. maybe i'll create a task to add this once we enable `save keypairs` functionality

jeesunikim: will add resetting before generating keypair

tests/fundAccountPage.test.ts
```diff
@@ -118,6 +118,7 @@ test.describe("[futurenet/testnet] Fund Account Page", () => {
 
     // Success <Alert/> is visible
     const alertBox = page.getByText(/Successfully funded/);
+    console.log("alertBox: ", alertBox);
```
quietbits, line 121: `console.log`

src/query/useFriendBot.ts
```diff
@@ -1,37 +1,41 @@
...
```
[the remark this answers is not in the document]
jeesunikim: i used this approach for the custom network (see [slack](https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1712695564373659))

src/app/(sidebar)/account/create/page.tsx
```diff
@@ -46,27 +95,51 @@ export default function CreateAccount() {
             <Button size="md" variant="secondary" onClick={generateKeypair}>
```
quietbits, line 95: Sorry, one more thing here. When funding is in progress, let's disable the "Generate keypair" button. I like how you did that on the Fund page. :raised_hands:

PR #832, opened 2024-04-22
[Build Transaction] Params input and validation + encode JSON to XDR

quietbits:
## Production

- Using Option 2 from below until Stellar XDR is updated to support string for number (stopgap solution). 
- `lossless-json` package is used to stringify JSON because native `JSON.stringify` doesn't work with `BigInt`.
- Everything should work (validation, keep values and validation on page refresh, etc). If something doesn't work, it's a bug.

<details>
<summary>XDR example is rendered in Operations tab</summary>

image
</details>

## Proof of concept

Using Stellar XDR to build transactions from JSON instead of transaction builder from Stellar SDK. This allows us to decode/encode to/from JSON/XDR easily.

This example uses only transaction params and no options, but the approach would be the same. Temporary validation and XDR output are displayed at the bottom of the page.

<details>
<summary>Screenshot</summary>

image
</details>

Everything works, but there is a challenge with the number/int type that is required in XDR for fields such as sequence number, base fee, and Memo ID. Inputs in the form must be strings, so at some point, we need to convert the string to number. There are a couple of ways we can handle this on the client.

1. Convert string to a number when saving the value in the store JSON (this is used directly to convert to XDR when needed). This makes the validation UI a little strange because we can only show something that is a number as it is typed in the field and not any other character. See the example below (I'm typing `aaa` when the error is shown). We can still show the error message to let the user know what is wrong.

<details>
<summary>Screen recording</summary>

[lab-fee-validation.webm](https://github.com/stellar/laboratory/assets/7346473/e0908db4-d05e-4671-9965-a65fb5c87027)
</details>

2. Create a helper that converts all needed number params before we encode them to XDR. This way, the validation is more straightforward: users can input anything and will see that value in the input field. The downside is that we need to manually write checks for every field that needs to be converted. For the params, it's not an issue because those fields are static and always required. It will be more work to implement this for operations.

<details>
<summary>Code example</summary>

image
</details>

3. The best solution for the client side would be for the Stellar XDR to accept numbers and strings for fields that must be number/int type and do the conversion when validating/building the XDR. I don't know if this is possible in Rust, nor that it's a good idea for the XDR.

>[!NOTE]
>Validation on page refresh doesn't work, so the XDR encoding might break. Make a change in any field to trigger validation. I didn't add page load validation to keep the updates smaller and more relevant to working with the Stellar XDR package.

If we're happy with how this works, we'll need to clean up the code in this PR and finish validation before merging it.


.eslintrc.json
```diff
@@ -6,9 +6,10 @@
...
     "prettier"
   ],
-  "rules":{
+  "rules": {
     // @TODO: ideally, these should be removed
     "@typescript-eslint/no-explicit-any": "off",
     "import/named": "off"
-  }
+  },
+  "ignorePatterns": ["./temp/*"]
```
quietbits, line 14: Adding these only while we're testing the Stellar XDR package locally in the repo. These ESlint updates will be removed once we use the npm package.

package.json
```diff
@@ -25,9 +25,13 @@
     "@tanstack/react-query": "^5.28.8",
     "@tanstack/react-query-devtools": "^5.28.8",
     "@typescript-eslint/eslint-plugin": "^7.5.0",
+    "bignumber.js": "^9.1.2",
+    "bindings-js": "file:./src/temp/stellar-xdr-web",
```
quietbits, line 29: Temporarily including this locally in the repo. When the package is published, we'll clean everything up.

src/app/(sidebar)/transaction/build/page.tsx
```diff
@@ -1,45 +1,350 @@
...
+                  : value;
+            }
+
+            break;
+          default:
+            val = value;
+        }
+
+        return { ...res, [key]: val };
+      }, {});
```
quietbits, line 308: This code won't be needed once Stellar XDR supports string numbers.

package.json
```diff
@@ -25,9 +25,12 @@
     "@tanstack/react-query": "^5.28.8",
     "@tanstack/react-query-devtools": "^5.28.8",
     "@typescript-eslint/eslint-plugin": "^7.5.0",
+    "bindings-js": "file:./src/temp/stellar-xdr-web",
     "dompurify": "^3.0.11",
     "html-react-parser": "^5.1.10",
     "immer": "^10.0.4",
+    "lodash": "^4.17.21",
```
jeesunikim, line 32: should we use lodash for some of the helpers that we had previously created, for example: `isEmptyObject` & `isEmpty`

jeesunikim (approved):
I checked the encoded XDR in old lab view XDR and they work!

src/app/(sidebar)/transaction/build/page.tsx
jeesunikim: should we add `@TODO` comment

quietbits: :+1: There is a `TODO:` on line 260.

PR #839, opened 2024-04-29
[sign tx] add signing with secrets and ValidationResponseCard

jeesunikim:
Sign Transaction
- Sign Transaction's `Import` and `Overview` view
- Signing with secret keys enabled
- We're not saving `secret keys` in store

note: waiting on design for the validation response's error state


src/app/(sidebar)/transaction/sign/components/Import.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  <Icon.AlertCircle />
+                </span>
+              </Text>
+            }
+            value={txXdr || ""}
+            error={txErrMsg}
+            success={
+              <Text size="xs" as="span" addlClassName="success-message">
+                {txSuccessMsg}
+              </Text>
```
quietbits, line 79: We might not need this extra styling anymore. The latest SDS version should handle it for ijnputs.

src/app/(sidebar)/transaction/sign/components/Import.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <div className="SignTx__xdr">
+          <XdrPicker
+            id="sign-transaction-xdr"
+            label={
+              <Text size="xs" as="span" weight="medium">
+                Import a transaction envelope in XDR format
+                <span className="SignTx__icon">
+                  <Icon.AlertCircle />
+                </span>
+              </Text>
```
quietbits, line 72: We don't need to style the label and use SDS default. There is a new `infoText` to add the info bubble.

src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const { xdr, message } = transactionSigner.secretKeys({
+      txXdr,
+      signers,
+      networkPassphrase,
+    });
+
+    if (xdr) {
+      updateSignedTx(xdr);
+    }
+    setSignedTxSuccessMsg(message);
```
quietbits, line 66: Could we get an error at this step when signing the transaction? :thinking:

src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -0,0 +1,286 @@
...
+              resetSign();
+            }}
+          >
+            Clear and import new
+          </Button>
+        </div>
+
+        <Card>
+          <div className="SignTx__FieldViewer">
+            {mergedFields!.map((field) => {
```
quietbits, line 115: Just to be safe, let's use `?` instead of "!" in case we don't get `mergedFields`.

src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+export const Overview = () => {
+  const { network, transaction } = useStore();
+  const { sign, updateSignActiveView, updateSignedTx, resetSign } = transaction;
+
+  const [secretInputs, setSecretInputs] = useState<string[]>([""]);
+  const [signedTxSuccessMsg, setSignedTxSuccessMsg] = useState<string>("");
+
+  // @TODO bip path
+  const [bipPath, setBipPath] = useState<string>("");
```
quietbits, line 32: Values that we want to persist when the page is refreshed, we should save in the store.

src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -0,0 +1,286 @@
...
+                  onClick={() =>
+                    signTxWithSecretKeys(
+                      sign.importXdr,
+                      secretInputs,
+                      network.passphrase,
+                    )
+                  }
+                >
+                  Sign with secret key
+                </Button>
```
quietbits, line 209: We should enable the sign button only when signatures are valid.

image

src/app/(sidebar)/transaction/sign/components/Import.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+      // change to 'overview' view when successfully imported
+      updateSignActiveView("overview");
+    } catch (e) {
+      setTxErrMsg("Unable to import a transaction envelope");
+      updateSignActiveView("import");
+    }
+  };
+
+  return (
```
quietbits, line 55: We're not persisting entered data when the page is refreshed. Is that intentional?

src/app/(sidebar)/transaction/sign/components/Overview.tsx
jeesunikim: i haven't started working on bipPath! I will do that in the next pr

src/app/(sidebar)/transaction/sign/components/Import.tsx
jeesunikim: that's how it is in the current lab

src/app/(sidebar)/transaction/sign/components/Import.tsx
jeesunikim: I don't think `infoText` works on [TextArea](https://design-system-pr208.previews.kube001.services.stellar-ops.com/textarea)

jeesunikim: oh I had to enable both label and infoText

src/app/(sidebar)/transaction/sign/components/Import.tsx
quietbits: The current Lab's state is broken, so it's not persisting the data after the page refresh. It would be a better UX if it did. What do you think?

jeesunikim: task created here https://github.com/stellar/laboratory/issues/845

src/validate/methods/secretKey.ts
```diff
@@ -0,0 +1,15 @@
...
+    if (!StrKey.isValidEd25519SecretSeed(value)) {
+      return "Invalid secret key.";
+    }
+  } else {
+    if (!value.match(/^[0-9a-f]{2,128}$/gi) || value.length % 2 == 1) {
+      return `Invalid hex value. Please provide up to 64 bytes in hexadecimal format.`;
+    }
+  }
+
+  return false;
```
jeesunikim, line 14: we are using `return false` for `validate` methods because we're using for `error = validate.secretKey` but when I just look at the validate method codes, I expect the validated return value to be `true` instead of `false`

quietbits: We are returning a string (error message) if it is not valid, so it made sense to return `false` or an empty string when it is valid. I chose to use `false` to make it easier to check response type. I agree that it is a bit confusing when looking at the method. Should we rename it to something other than `validate`?

jeesunikim: how about the following?
```
validators/
  └── methods/
      ├── getAmountValidationError.ts
      ├── getAssetValidationError.ts
      ├── getPositiveIntegerValidationError.ts
```

quietbits (approved):

PR #843, opened 2024-05-08
[Build Transaction] Added operations 

quietbits:
- Added operation description and docs URL
- Some refactoring to improve operation field validation and XDR formatting
- Added the following operations:
  - Payment
  - Manage Sell Offer
  - Manage Buy Offer
  - Create Passive Sell Offer

### DIFF
--- a/src/components/formComponentTemplateTxnOps.tsx
+++ b/src/components/formComponentTemplateTxnOps.tsx
@@ -3,9 +3,10 @@ import { JSX } from "react";
 import { PubKeyPicker } from "@/components/FormElements/PubKeyPicker";
 import { TextPicker } from "@/components/FormElements/TextPicker";
 import { AssetPicker } from "@/components/FormElements/AssetPicker";
+import { PositiveIntPicker } from "@/components/FormElements/PositiveIntPicker";
 
 import { validate } from "@/validate";
-import { AssetObjectValue } from "@/types/types";
+import { AnyObject, AssetObjectValue, JsonAsset } from "@/types/types";
 
 // Types
 type TemplateRenderProps = {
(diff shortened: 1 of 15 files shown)


src/components/formComponentTemplateTxnOps.tsx
```diff
@@ -0,0 +1,216 @@
+import { JSX } from "react";
```
quietbits, line 1: I separated endpoints and operation components because the same param required a different type of component. The logic is more clear now, even though there will be some duplication in code.

src/helpers/xdr/fraction.ts
```diff
@@ -0,0 +1,42 @@
+// This content is copied from js-stellar-base (src/util/continued_fraction.js)
```
quietbits, line 1: I put all the XDR helpers from `js-stellar-base` in this folder to make it easier to extract them later if needed.

jeesunikim:
Question: for 'Payment operation, we save native asset related value in its fields, but we don't do the same thing for `Alphanumeric 4` or `Alphanumeric 12`. Why is that?

payment-native
payment-alpha

jeesunikim (approved):

quietbits:
> Question: for 'Payment operation, we save native asset related value in its fields, but we don't do the same thing for `Alphanumeric 4` or `Alphanumeric 12`. Why is that? This is the behavior in our current lab.
> 
> payment-native payment-alpha

Thanks for the question, @jeesunikim ! :raised_hands: The values for non-native asset inputs are cleared when the type changes because the values can't be used interchangeably. For example, an Alphanumeric 4 asset code is not a valid Alphanumeric 12 asset code, and vice versa.

PR #847, opened 2024-05-17
[Build Transaction] More operations pt. 4

quietbits:
Added operations:
- Clawback
- Clawback Claimable Balance
- Path Payment Strict Send
- Path Payment Strict Receive


jeesunikim:
Screenshot 2024-05-20 at 4 51 41 PM
what does this copy icon supposed to copy? it's currently copying the destination asset's `issuer account ID`. Is that expected?

src/components/formComponentTemplateTxnOps.tsx
```diff
@@ -84,17 +96,24 @@ export const formComponentTemplateTxnOps = ({
...
             id={id}
-            label="Asset"
+            label={custom?.label ?? "Asset"}
             labelSuffix={!templ.isRequired ? "optional" : undefined}
             value={assetPickerValue(templ.value)}
             error={templ.error}
-            includeNative
+            note={custom?.note}
+            includeNative={
+              typeof custom?.includeNative === "undefined"
```
jeesunikim, line 113: if `includeNative` is `undefined`, native is included?

quietbits: Yes, by default, it's enabled because we show it more often than hide it. If it's set to `false` in custom config, then use that value to hide the native asset option.

quietbits:
> Screenshot 2024-05-20 at 4 51 41 PM
> what does this copy icon supposed to copy? it's currently copying the destination asset's `issuer account ID`. Is that expected?
> 
> Oh I see it copies the operation method!

Yes, it duplicates the operation.

jeesunikim (approved):

PR #858, opened 2024-06-04
[Build Transaction] Set Trust Line Flags + Revoke Sponsorship operations

quietbits:


src/components/FormElements/RevokeSponsorshipPicker.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  const getInputId = (type: string, inputId: string) =>
+    `${id}-${type}-${inputId}`;
+
+  const renderInputs = () => {
+    if (!value?.type) {
+      return null;
+    }
+
+    switch (value.type) {
```
quietbits, line 96: I tried to make this piece more DRY, but it was harder to understand. So, I left it this way for clarity.

jeesunikim (approved):

PR #859, opened 2024-06-05
[Sign Tx] Add a sign with hardware wallets (trezor, ledger)

jeesunikim:
**Summary:**
- My ledger malfunctioned so I couldn't test the ledger part. The replacement is on its way

**Examples:**

- `Sign with Trezor` Example with a secret key and trezor signature:

https://github.com/stellar/laboratory/assets/3912060/135e72cc-836e-4693-8256-3a93a8eab122

- `Sign with Ledger` Error State

error-state


src/store/createStore.ts
```diff
@@ -322,10 +330,23 @@ export const createStore = (options: CreateStoreOptions) =>
             set((state) => {
               state.transaction.sign.signedTx = tx;
             }),
+          updateBipPath: (bipPath: string) =>
+            set((state) => {
+              state.transaction.sign.bipPath = bipPath;
```
quietbits, line 335: We are saving the BIP path in the store, but it does not persist when the page is refreshed. Should we persist it if it's already in the store?

src/styles/globals.scss
```diff
@@ -511,4 +511,22 @@
       gap: pxToRem(8px);
     }
   }
+
+  .Input__container {
+    .Select__container {
+      border: 0;
+      // border-left: 1px solid var(--Input-color-border);
```
quietbits, line 518: Commented out code

quietbits (approved):

PR #875, opened 2024-06-12
Link "Sign in Transaction Signer" button to "Sign Transaction" page

jeesunikim:
Link `Sign in Transaction Signer` button from "Build Transaction" page to "Sign Transaction" page

Screenshot 2024-06-12 at 11 49 59 AM

### DIFF
--- a/src/store/createStore.ts
+++ b/src/store/createStore.ts
@@ -166,7 +166,7 @@ const initTransactionState = {
     },
   },
   sign: {
-    activeView: "import",
+    activeView: SignTxActiveView.import,
     importTx: undefined,
     importXdr: "",
     signedTx: "",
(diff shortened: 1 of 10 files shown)


src/store/createStore.ts
```diff
@@ -39,6 +39,11 @@ type TransactionBuildParamsObj = {
   [K in keyof TransactionBuildParams]?: TransactionBuildParams[K];
 };
 
+export enum SignTxActiveView {
+  import = "import",
+  overview = "overview",
+}
+
```
quietbits, line 46: We could use a string union instead of an enum in cases like this. That would give us better hints in code editors.

quietbits (approved):

PR #879, opened 2024-06-13
[Submit Tx] Add a Submit Tx Page

jeesunikim:
**Summary:**
- Created `useIsXdrInit` hook to reuse `await StellarXdr.init();`
- `<ErrorResponse/>` is based on our current horizon_api [code](https://github.com/stellar/js-stellar-sdk/blob/7dc0a8cfee29c0260c260d2a2552bb394d08c929/src/horizon/horizon_api.ts#L5). There are a few edge cases to cover
- Using `useMutation` for async submit function call


src/app/(sidebar)/transaction/submit/page.tsx
```diff
@@ -1,5 +1,201 @@
...
+    submitTx.mutate(
+      { transaction, server },
+      {
+        onSuccess: (res) => setTxResponse(res),
+        onError: (res) => setTxErr(res),
+      },
+    );
+  };
+
+  const xdrDecodeJson = () => {
```
jeesunikim, line 54: we're reusing this code so shall put it in `/helpers`

src/app/(sidebar)/transaction/submit/page.tsx
```diff
@@ -1,5 +1,201 @@
...
+                  <div>Result Meta XDR:</div>
+                  <div>{txResponse.result_meta_xdr}</div>
+                </div>
+                <div>
+                  <div>Fee Meta XDR:</div>
+                  <div>{txResponse.fee_meta_xdr}</div>
+                </div>
+              </Box>
+            }
+            // @TODO: the current lab doesn't support displaying stellar.expert
```
jeesunikim, line 177: i'll remove this once I confirm with @charles-sdf tomorrow

src/store/createStore.ts
```diff
@@ -93,6 +93,9 @@ export interface Store {
       bipPath: string;
       hardWalletSigs: xdr.DecoratedSignature[] | [];
     };
+    submit: {
+      xdrBlob: string;
+    };
```
quietbits, line 98: I think we should use the `xdr` store we already have. Lab v1 shows XDR in both places when added on either page, so that would be consistent with that behavior. Also, XDR can get very long, and we might run out of URL space if we need to keep 2 XDRs in the URL (for View XDR and Submit Txn).

src/app/(sidebar)/transaction/submit/page.tsx
```diff
@@ -1,5 +1,201 @@
...
+import { ErrorResponse } from "./components/ErrorResponse";
+
 export default function SubmitTransaction() {
-  return <div>Submit Transaction</div>;
+  const { network, transaction } = useStore();
+  const { updateXdrBlob, submit } = transaction;
+
+  const { xdrBlob } = submit;
+
+  const isXdrInit = useIsXdrInit();
```
quietbits, line 28: :clap:

src/app/(sidebar)/transaction/submit/components/ErrorResponse.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface ErrorProps {
+  error: NetworkError & {
+    response: {
+      data: {
+        extras: {
+          result_codes: string;
+          result_xdr: string;
+        };
+      };
```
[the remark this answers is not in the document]
jeesunikim: that shouldn't hurt! The current lab doesn't make them optional ([code](https://github.com/stellar/laboratory/blob/693aecb45bd9c92ace104a5dc0804292fbeb2e5c/src/components/TxSubmitterResult.tsx#L188-L200)) but I think it's okay to make it optional.

src/app/(sidebar)/transaction/submit/page.tsx
```diff
@@ -1,5 +1,201 @@
...
+import { useIsXdrInit } from "@/hooks/useIsXdrInit";
+
+import { TransactionResponse, useSubmitTx } from "@/query/useSubmitTx";
+
+import { Box } from "@/components/layout/Box";
+import { PrettyJson } from "@/components/PrettyJson";
+import { Tabs } from "@/components/Tabs";
+import { XdrPicker } from "@/components/FormElements/XdrPicker";
+import { ValidationResponseCard } from "@/components/ValidationResponseCard";
+import { ErrorResponse } from "./components/ErrorResponse";
```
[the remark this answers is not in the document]
jeesunikim: it's the `/component` subfolder within `/submit` folder

src/store/createStore.ts
jeesunikim: good point!

src/app/(sidebar)/transaction/submit/components/ErrorResponse.tsx
```diff
@@ -0,0 +1,78 @@
...
+    extras = (
+      <Box gap="xs">
+        <div>
+          <div>Destination account:</div>
+          <div>{error.accountId}</div>
+        </div>
+        <div>
+          <div>Operation index:</div>
+          {error.operationIndex}
+        </div>
```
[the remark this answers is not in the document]
jeesunikim: Screenshot 2024-06-14 at 5 01 54 PM
I made it work with `<Box/>`, but I had to add an extra empty `div` which makes me think i should just create a css style.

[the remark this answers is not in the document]
quietbits: Yeah, I think we should have a custom component for this because we have a similar layout in other places. We can still use the `Box` as a container; the label element can be a simple `div`, but the value should have a class to set the padding/margin from the left. :+1:

src/components/TxResponse/styles.scss
```diff
@@ -0,0 +1,7 @@
+@use "../../styles/utils.scss" as *;
+
+.TxResponse {
+  &__value {
+    margin-left: calc(var(--sds-gap-sm));
```
quietbits, line 5:
```suggestion
    margin-left: var(--sds-gap-sm);
```

src/app/(sidebar)/transaction/submit/page.tsx
```diff
@@ -1,5 +1,163 @@
...
+                />
+                <TxResponse label="Result XDR:" value={txResponse.result_xdr} />
+                <TxResponse
+                  label="Result Meta XDR:"
+                  value={txResponse.result_meta_xdr}
+                />
+                <TxResponse
+                  label="Fee Meta XDR:"
+                  value={txResponse.fee_meta_xdr}
+                />
```
quietbits, line 150: Much better! :tada:

quietbits (approved):

PR #895, opened 2024-07-01
[Transactions] Saved Transactions

quietbits:
- Saved transactions page
- Updated Saved endpoints layout
- Enabled prefetch for navigation links for smoother page transition
  - NOTE: Without this, I lost search params when navigating to pages locally (only in Chrome browser). Adding this fixed it, but we need to watch for anything else we might need to optimize. It looks like search params were lost because of slow page load (timing issue).

<hr />
Saved transactions

image

Updated Saved endpoints layout

image

### DIFF
--- a/src/app/(sidebar)/transaction/build/components/TransactionXdr.tsx
+++ b/src/app/(sidebar)/transaction/build/components/TransactionXdr.tsx
@@ -10,6 +10,7 @@ import { useRouter } from "next/navigation";
 import { SdsLink } from "@/components/SdsLink";
 import { ValidationResponseCard } from "@/components/ValidationResponseCard";
 import { Box } from "@/components/layout/Box";
+import { ViewInXdrButton } from "@/components/ViewInXdrButton";
 
 import { isEmptyObject } from "@/helpers/isEmptyObject";
 import { xdrUtils } from "@/helpers/xdr/utils";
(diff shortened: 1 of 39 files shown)


src/app/(sidebar)/transaction/build/components/TransactionXdr.tsx
```diff
@@ -547,20 +548,13 @@ export const TransactionXdr = () => {
                   updateSignImportXdr(txnXdr.xdr);
                   updateSignActiveView("overview");
 
-                  router.push("/transaction/sign");
+                  router.push(Routes.SIGN_TRANSACTION);
```
quietbits, line 551: We should use `Routes` instead of strings to make future updates easier.

src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -284,6 +285,7 @@ export const Overview = () => {
                 onUpdate={onUpdateSecretInputs}
                 validate={validate.secretKey}
                 placeholder="Secret key (starting with S) or hash preimage (in hex)"
+                autocomplete="off"
```
quietbits, line 288: We need to make sure we never save private keys in the browser with autocomplete.

src/app/(sidebar)/transaction/saved/page.tsx
```diff
@@ -0,0 +1,200 @@
...
+          }
+          rightElement={
+            <InputSideElement
+              variant="button"
+              placement="right"
+              onClick={() => {
+                setCurrentTxnTimestamp(txn.timestamp);
+              }}
+              icon={<Icon.Edit05 />}
+            ></InputSideElement>
```
jeesunikim, line 79:
```suggestion
            />
```

jeesunikim (approved):

PR #919, opened 2024-07-18
Rename validate methods function

jeesunikim:
rename validate functions to describe its behavior
- our validate functions return an error message. Previous naming was confusing because I'd expect it to return `true` if it's valid, but its functions return `true` if there's an error aka it returns an error message.


fnando (approved):
Loved the more explicit naming.

PR #932, opened 2024-07-29
UI tweaks

quietbits:
- Copy and Save action tooltip
- Fixed text spacing in `Alert`
- Removed underline from links in the main nav
- Use a custom icon in Alert on the Saved Transactions page

Details with screenshots in the [SDS PR](https://github.com/stellar/stellar-design-system/pull/251).

### DIFF
--- a/src/components/NetworkSelector/styles.scss
+++ b/src/components/NetworkSelector/styles.scss
@@ -49,6 +49,7 @@
 
   // Floater
   &__floater {
+    z-index: 2;
     top: calc(100% + 0.5rem);
     right: 0;
     left: auto;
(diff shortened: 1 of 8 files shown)


src/components/NetworkSelector/styles.scss
```diff
@@ -125,6 +126,23 @@
...
+          &:nth-child(2) {
+            z-index: 3;
+          }
+          &:nth-child(3) {
+            z-index: 2;
+          }
+          &:nth-child(4) {
+            z-index: 1;
+          }
+        }
```
quietbits, line 145: This is a hacky temporary fix for the copy tooltip overlap in input if it is followed by another input with a copy button. A proper fix will need to be applied in the SDS, but it might not be a simple one ([issue](https://github.com/stellar/stellar-design-system/issues/254)). Added this to save time right now.

Before

[lab-input-copy-tooltip-overlap-before.webm](https://github.com/user-attachments/assets/a6bd995e-aff8-4c62-9a78-6e53565028e0)

After

[lab-input-copy-tooltip-overlap-after.webm](https://github.com/user-attachments/assets/a69eb1cb-40c6-4bf4-93d6-ed217a860cb7)

jeesunikim (approved):

PR #934, opened 2024-07-30
Simulate Transaction page

quietbits:
>[!NOTE]
>The Simulate transaction uses the same logic as the Submit transaction, where `xdr` is used globally (also used in the XDR to JSON page). All these pages share this param. Because the XDR can be very long, we should be mindful of how it is saved in the URL and shouldn't save it multiple times. If the current logic is not good UX, we can adjust it, but we should still use only a single XDR value in the URL.

<details>
<summary>Screenshots</summary>
<hr />

Success
image

Error
image

Warning
image
</details>


src/query/useSimulateTx.ts
```diff
@@ -0,0 +1,62 @@
...
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          jsonrpc: "2.0",
+          id: 1,
+          method: "simulateTransaction",
+          params: {
+            transaction: transactionXdr,
+            ...(instructionLeeway
```
jeesunikim, line 28: oh I might do it this way for rpc endpoints instead of my current approach to `instructionLeeway: Number(params.resourceConfig) || undefined`. TY!

jeesunikim (approved):

src/query/useSimulateTx.ts
quietbits: :+1: This only works if the server is not expecting that param when optional. Sometimes we still need to pass the param without value. Just something to keep in mind.

PR #937, opened 2024-07-30
Amplitude tracking

quietbits:
- Amplitude API key will be used as an `env` variable (we'll need assistance from Ops to set it up).
- The tracking won't work in production (PR preview or beta website) until that work is done.
- It can be tested locally by setting a `NEXT_PUBLIC_AMPLITUDE_API_KEY` `env` variable.


src/metrics/tracking.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  try {
+    amplitude.track(type, properties);
+  } catch (e) {
+    // Do nothing
+  }
+};
+
+export enum TrackingEvent {
```
quietbits, line 41: @janewang These are some basic events we can start with, but we can always tweak them and add more.

janewang: @quietbits Looks great as a starting point, thank you

janewang (approved):

quietbits:
Closing this PR for now as it's getting way behind the updates.

PR #953, opened 2024-08-13
XDR: fix large number parsing in JSON

quietbits:
Native JSON doesn't parse large numbers correctly, so we need to use a library for that.

### DIFF
--- a/src/helpers/parseToLosslessJson.ts
+++ b/src/helpers/parseToLosslessJson.ts
@@ -1,0 +1,12 @@
+import { parse, isNumber } from "lossless-json";
+import { AnyObject } from "@/types/types";
+
+export const parseToLosslessJson = (stringObj: string) => {
+  return parse(stringObj, null, (value) => {
+    if (isNumber(value)) {
+      return BigInt(value);
+    }
+
+    return value;
+  }) as AnyObject;
+};
--- a/src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
+++ b/src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
@@ -13,6 +13,7 @@ import {
   Textarea,
 } from "@stellar/design-system";
 import { useQueryClient } from "@tanstack/react-query";
+import { stringify } from "lossless-json";
 
 import { SdsLink } from "@/components/SdsLink";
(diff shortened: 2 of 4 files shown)


jeesunikim (approved):

src/helpers/parseToLosslessJson.ts
```diff
@@ -0,0 +1,12 @@
+import { parse, isNumber } from "lossless-json";
+import { AnyObject } from "@/types/types";
+
+export const parseToLosslessJson = (stringObj: string) => {
```
Shaptic, line 4: I thought you could accomplish this via `JSON.stringify`'s `replacer` parameter ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)) like so:

```js
let js = { "bigNum": 10n }
let json = JSON.stringify(js, (key, val) => typeof val === 'bigint' ? val.toString() : val);
console.log(json);
// {"bigNum":"10"}
```
iirc either @piyalbasu or @aristidesstaffieri had to do this in Freighter somewhere

Or is the issue then converting back into a JSON object, since then you'd have to describe somewhere that `"bigNum"` should encode to a `BigInt`?

Shaptic: er this comment is in the wrong place but you know what I mean

aristidesstaffieri: yup, examples is here - https://github.com/stellar/freighter/blob/74aaf2b54d888a8b3d61568fbbc524f5c07f17df/extension/src/popup/components/signAuthEntry/AuthEntry/index.tsx#L36

quietbits: Yes, it's basically the same thing. This library provides more checks for all numbers. Since we use it to stringify as well, I thought it's good to use it here as well instead of checking all the cases manually.

Shaptic: Ah got it - I thought you were introducing the dependency just for this functionality; should've read the PR closer :see_no_evil:

quietbits: No worries at all, thank you, @Shaptic ! It's good to have these things in check. :+1:

PR #954, opened 2024-08-14
XDR: update labels

quietbits:


janewang (approved):
Thank you for the quick fix

PR #956, opened 2024-08-16
Endpoints: handle JSON response links

quietbits:
Feature parity with Lab v1 + added some extra cases that we have routes for.

We will always open Endpoint JSON response links in a new tab to preserve the context from the query in the initial tab.

### DIFF
--- a/src/components/PrettyJson/index.tsx
+++ b/src/components/PrettyJson/index.tsx
@@ -6,7 +6,23 @@ import { AnyObject } from "@/types/types";
 
 import "./styles.scss";
 
-export const PrettyJson = ({ json }: { json: AnyObject }) => {
+export type CustomKeyValueLinkMap = {
+  [key: string]: {
+    text?: string;
+    action: (value: string, key?: string) => void;
+    condition?: (val: string) => boolean;
+  };
+};
+
+type PrettyJsonProps = {
+  json: AnyObject;
+  customKeyValueLinkMap?: CustomKeyValueLinkMap;
+};
+
+export const PrettyJson = ({
+  json,
+  customKeyValueLinkMap,
+}: PrettyJsonProps) => {
   if (typeof json !== "object") {
     return null;
   }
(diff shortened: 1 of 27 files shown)


MonsieurNicolas:
cool thank you!

I just tried it with transactions (from "all ledgers"), when clicking on the xdr for the transaction frame xdr, I am taken as expected to the decoded page, but when hitting "back" I need to resubmit the query, so context is loss (that endpoint's result changes with every ledger) so I can't open both the transaction and its result for example.

I tried to open in a different tab (ctrl-click) but it doesn't work, so there doesn't seem to be a workaround.

quietbits:
> cool thank you!
> 
> I just tried it with transactions (from "all ledgers"), when clicking on the xdr for the transaction frame xdr, I am taken as expected to the decoded page, but when hitting "back" I need to resubmit the query, so context is loss (that endpoint's result changes with every ledger) so I can't open both the transaction and its result for example.
> 
> I tried to open in a different tab (ctrl-click) but it doesn't work, so there doesn't seem to be a workaround.

Thanks for checking, @MonsieurNicolas ! 

The context is lost because we clear it when a link is clicked. In the XDR case, we could keep the previous state in Endpoints. Then, it would be there when a back button is clicked.

But we can't do that when clicking on another link that goes to the Endpoints page (account, for example) because it's the same parent page (the API response will be updated, errors cleared, etc). We would need to do some state caching to make that work, which would be complex.

We could make them proper links (instead of button links they are right now) to open them in a new tab. Would that be acceptable?

src/components/PrettyJson/index.tsx
```diff
@@ -44,6 +60,26 @@ export const PrettyJson = ({ json }: { json: AnyObject }) => {
...
+
+      if (custom) {
+        if (custom.condition && !custom.condition(item)) {
+          return render(item, key);
+        }
+
+        const href = custom.getHref(item, key);
+
+        return (
+          <SdsLink href={href || item} {...(href ? { target: "_blank" } : {})}>
```
jeesunikim, line 74: thoughts on adding an underline for links to make it clearer that it's linkable?

jeesunikim (approved):

src/components/PrettyJson/index.tsx
quietbits: @charles-sdf what do you think?

charles-sdf: @jeesunikim @quietbits yep im good with that! let's add em

PR #959, opened 2024-08-20
Add getLedgerEntries

jeesunikim:
TO DO:
- [x] styling
- [ ] asset dropdown

### DIFF
--- a/src/components/FormElements/XdrLedgerKeyPicker.tsx
+++ b/src/components/FormElements/XdrLedgerKeyPicker.tsx
@@ -3,8 +3,8 @@ import { ReactElement, useEffect, useState } from "react";
 import { Button, Card, Icon, Select } from "@stellar/design-system";
 
 import { arrayItem } from "@/helpers/arrayItem";
-import { isEmptyObject } from "@/helpers/isEmptyObject";
-import { sanitizeObject } from "@/helpers/sanitizeObject";
+// import { isEmptyObject } from "@/helpers/isEmptyObject";
+// import { sanitizeObject } from "@/helpers/sanitizeObject";
 import * as StellarXdr from "@/helpers/StellarXdr";
 
 import { Box } from "@/components/layout/Box";
@@ -18,7 +18,7 @@ import { validate } from "@/validate";
 
 import {
   AnyObject,
-  AssetObjectValue,
+  // AssetObjectValue,
   LedgerKeyFieldsType,
   LedgerKeyType,
 } from "@/types/types";
@@ -42,7 +42,7 @@ const ledgerKeyFields: {
   id: LedgerKeyType;
   label: string;
   templates: string;
-  addlFields?: AnyObject;
+  custom?: AnyObject;
 }[] = [
   {
     id: "account",
@@ -53,7 +53,9 @@ const ledgerKeyFields: {
     id: "trustline",
     label: "Trustline",
     templates: "account_id,asset",
-    addlFields: {
+    custom: {
+      assetInput: "alphanumeric",
+      includeLiquidityPoolShares: true,
       asset: {
         type: "",
         code: "",
@@ -74,7 +76,7 @@ const ledgerKeyFields: {
   {
     id: "claimable_balance",
     label: "Claimable Balance",
-    templates: "claimable_balance_id",
+    templates: "balance_id",
   },
   {
     id: "liquidity_pool",
@@ -226,8 +228,8 @@ export const XdrLedgerKeyPicker = ({
         return accr;
       }, {} as AnyObject);
 
-      if (selectedLedgerKey.addlFields) {
-        formLedgerKeyJson[selectedLedgerKey.id] = selectedLedgerKey.addlFields;
+      if (selectedLedgerKey.custom) {
+        formLedgerKeyJson[selectedLedgerKey.id] = selectedLedgerKey.custom;
       }
 
       const formLedgerKeyJsonString = JSON.stringify(formLedgerKeyJson);
@@ -245,10 +247,13 @@ export const XdrLedgerKeyPicker = ({
       const json = JSON.parse(ledgerKeyJsonString);
       const obj = json[selectedLedgerKey!.id];
 
-      const component = formComponentTemplateEndpoints(template);
+      const component = formComponentTemplateEndpoints(
+        template,
+        selectedLedgerKey?.custom,
+      );
 
       if (component) {
-        const handleChange = (key: any, val: any) => {
+        const handleInputChange = (key: any, val: any) => {
           const error = component.validate?.(val);
 
           if (error) {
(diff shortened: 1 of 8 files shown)


src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -0,0 +1,406 @@
...
+        onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
+          const selectedVal = e.target.value;
+          setLedgerKeyJsonString("");
+
+          if (selectedVal) {
+            const selectedLedgerKey = getKeyType(selectedVal);
+            selectLedgerKey(selectedLedgerKey!);
+          } else {
+            selectLedgerKey(null);
+          }
```
Shaptic, line 329: Just a suggestion in passing: I'm a sucker for one-liners lol but feel free to ignore

```suggestion
          selectLedgerKey(selectedVal ? getKeyType(selectedVal)! : null);
```

src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -0,0 +1,406 @@
...
+  },
+  {
+    id: "liquidity_pool",
+    label: "Liquidity Pool",
+    templates: "liquidity_pool_id",
+  },
+  {
+    id: "contract_data",
+    label: "Contract Data",
+    templates: "contract,key,durability",
```
Shaptic, line 90: I don't see code to render the `key` field as JSON (via [`scValToNative`](https://stellar.github.io/js-stellar-sdk/global.html#scValToNative)), will that be a part of this? It'd be crazy useful.

Shaptic (commented):
Just a lil' drive by review heh

src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -0,0 +1,406 @@
+import { ReactElement, useEffect, useState } from "react";
+
+import { Button, Icon } from "@stellar/design-system";
+
+import { Select } from "@stellar/design-system";
```
quietbits, line 5: We can combine these imports

src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -0,0 +1,406 @@
...
+  },
+];
+
+export const XdrLedgerKeyPicker = ({
+  id,
+  value,
+  onChange,
+}: XdrLedgerKeyPicker) => {
+  const [selectedLedgerKey, selectLedgerKey] =
+    useState<LedgerKeyFieldsType | null>(null);
```
quietbits, line 115: Nit: To keep consistent naming convention.

```suggestion
  const [selectedLedgerKey, setSelectedLedgerKey] =
    useState<LedgerKeyFieldsType | null>(null);
```

src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -0,0 +1,406 @@
...
+      }, {} as AnyObject);
+
+      if (selectedLedgerKey.addlFields) {
+        formLedgerKeyJson[selectedLedgerKey.id] = selectedLedgerKey.addlFields;
+      }
+
+      const formLedgerKeyJsonString = JSON.stringify(formLedgerKeyJson);
+      setLedgerKeyJsonString(formLedgerKeyJsonString);
+    }
+  }, [ledgerKeyJsonString, selectedLedgerKey]);
```
quietbits, line 239: We're doing a lot in `useEffect` in a single component. This might affect performance if we're not careful. Maybe we can simplify these?

I think it's not good to have a `useEffect` dependency that can go from `undefined` to a value. It would be better to use a variable. For example:

```typescript
// This way, we always check a string
const decodedJsonString = xdrJsonDecoded?.jsonString || "";
```

src/components/FormElements/XdrLedgerKeyPicker.tsx
jeesunikim: ty!

janewang:
Screenshot 2024-08-21 at 11 23 05 AM

@jeesunikim I am seeing an issue with claimable balances. The error is "message":
"cannot unmarshal key value at index 0". I think this is a type error.

janewang:
Trustline:

For Trustline, I cannot submit the form below. See video below

https://github.com/user-attachments/assets/23697d38-c3ba-429a-9a5c-583153ccf846

I think Trustline accepts 4 different types based on the switch statement here: https://github.com/stellar/stellar-xdr/blob/curr/Stellar-ledger-entries.x#L248 On the UI, I only saw two types to be selected: native vs issued. 

```
union TrustLineAsset switch (AssetType type)
{
case ASSET_TYPE_NATIVE: // Not credit
    void;

case ASSET_TYPE_CREDIT_ALPHANUM4:
    AlphaNum4 alphaNum4;

case ASSET_TYPE_CREDIT_ALPHANUM12:
    AlphaNum12 alphaNum12;

case ASSET_TYPE_POOL_SHARE:
    PoolID liquidityPoolID;

    // add other asset types here in the future
};
```

jeesunikim:
@janewang claimable balance issue has been resolved. I'm currently working on enabling 'asset dropdown' for `Trustline`.

janewang:
@jeesunikim Btw I'm having a lot of issues testing on mainnet due to the content security policy issue. 

Screenshot 2024-08-21 at 12 52 48 PM

janewang:
contract code

https://github.com/user-attachments/assets/8c14d787-46ee-45c8-a8f0-ec0759f72636

When requesting contract code, I cannot submit form using this wasm hash `f5c9668827b4783e1f87a7f6b7406f6c426b72e82f114654a724713e4e6c0de4` on [testnet](https://stellar.expert/explorer/testnet/contract/CBHNQTKJD76Q55TINIT3PPP3BKLIKIQEXPTQ32GUUU7I3CHBD5JECZLW). 

Also please note that the field title is incorrect. The contract is tied to a `WASM Hash`, not a `transaction hash`.

janewang:
contract data

https://github.com/user-attachments/assets/6648580d-58e6-4cd0-a38d-5f25d0046bd3

I'm testing using the same contract on [testnet](https://stellar.expert/explorer/testnet/contract/CBHNQTKJD76Q55TINIT3PPP3BKLIKIQEXPTQ32GUUU7I3CHBD5JECZLW), which only has [one data entry](https://stellar.expert/explorer/testnet/contract/CBHNQTKJD76Q55TINIT3PPP3BKLIKIQEXPTQ32GUUU7I3CHBD5JECZLW/storage). But I cannot request the data from this contract. The form doesn't submit.

janewang:
TTL: 

I think something is not quite right with this TTL form per data structure.

Screenshot 2024-08-21 at 1 34 43 PM


I think the definition for [TTLEntry](https://github.com/stellar/stellar-xdr/blob/curr/Stellar-ledger-entries.x#L542) has two required fields: 

```
struct TTLEntry {
    // Hash of the LedgerKey that is associated with this TTLEntry
    Hash keyHash;
    uint32 liveUntilLedgerSeq;
};
```

janewang:
Config Setting fairly complex. 

Screenshot 2024-08-21 at 1 40 23 PM

In this switch statement, [in this line](https://github.com/stellar/stellar-xdr/blob/curr/Stellar-ledger-entries.x#L583) we see that data structure expected is a `ConfigSettingEntry`.

In [this file here](https://github.com/stellar/stellar-xdr/blob/curr/Stellar-contract-config-setting.x#L257), we see that the ID expected is a switch statement so this means we need a dropdown list for each of the option in the switch statement here: https://github.com/stellar/stellar-xdr/blob/curr/Stellar-contract-config-setting.x#L241-L254

In this different IDs, some of the cases are simple like `CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES`, but we also have fairly complex data required for `CONFIG_SETTING_CONTRACT_COMPUTE_V0` for example.

janewang:
Data: 



Screenshot 2024-08-21 at 3 02 29 PM

For `Data` ledger key, according to this [data structure](https://github.com/stellar/stellar-xdr/blob/curr/Stellar-ledger-entries.x#L355-L368), it may require a data value input field.

```
/* DataEntry
    Data can be attached to accounts.
*/
struct DataEntry
{
    AccountID accountID; // account this data belongs to
    string64 dataName;
    DataValue dataValue;

    // reserved for future use
    union switch (int v)
    {
    case 0:
        void;
    }
    ext;
};
```

leighmcculloch:
There are LedgerKey's and LedgerEntry's, and for the input to the getLedgerEntries we only need to produce LedgerKey's which are much simpler.

We shouldn't need TTLEntry, or DataEntry, that I see mentioned above. The entry structures is what the endpoint will return.

As input we just need to produce the LedgerKey type here:

https://github.com/stellar/stellar-xdr/blob/4ec28d95dd84b109253e22b151314478d6f00522/Stellar-ledger-entries.x#L600-L663

leighmcculloch:
Can we generate the form for the getLedgerEntries endpoint off the JSON schema of the `LedgerKey` XDR type?

PR #965, opened 2024-08-21
Set more exciting text around Horizon exploration

Shaptic:
Just tryna make it a little more hyphy :sunglasses:


Shaptic:
Oh wait I'll wait for the upstream v2 merge

PR #984, opened 2024-08-24
Implement GetLedgerEntries

jeesunikim:
**Testing Note:**
- Account: Use any public key
- Trustline:
-- For pool share, use `67260c4c1807b262ff851b0a3fe141194936bb0215b2f77447f1df11998eabb9`
-- pool share xdr: `AAAAAQAAAACm45wpNUH4LHLWH4qc39cu1tmRsEY1/2+qC4lklkkmCQAAAANnJgxMGAeyYv+FGwo/4UEZSTa7AhWy93RH8d8RmY6ruQ==`
- Offer
- Data
- Claimable Balance
-- `balance_id`: `0000000010a8f6991f79df306f22a2032f6007ad594dd30f966b21556f7d75658ec1c4e9`
- Liquidity Pool
-- `67260c4c1807b262ff851b0a3fe141194936bb0215b2f77447f1df11998eabb9`
- Contract Data
-- xdr example: `AAAABgAAAAHXkotywnA8z+r365/0701QSlWouXn8m0UOoshCtNHOYQAAABAAAAABAAAAAgAAAA8AAAAHQmFsYW5jZQAAAAASAAAAAcwD/nT9D7Dc2LxRdab+2vEUF8B+XoN7mQW21oxPT8ALAAAAAQ==`
- Contract Code
-- wasm hash: `f5c9668827b4783e1f87a7f6b7406f6c426b72e82f114654a724713e4e6c0de4`
-- xdr: `AAAAB/XJZogntHg+H4en9rdAb2xCa3LoLxFGVKckcT5ObA3k`
- TTL
- Config Setting

### DIFF
--- a/src/components/FormElements/AssetPicker.tsx
+++ b/src/components/FormElements/AssetPicker.tsx
@@ -53,7 +53,7 @@ export const AssetPicker = ({
   fitContent,
   includeNative = true,
   includeLiquidityPoolShares,
-  includePoolShare,
+  includeSingleLiquidityPoolShare,
 }: AssetPickerProps) => {
   let options: AssetObject[] = [];
 
@@ -117,7 +117,7 @@ export const AssetPicker = ({
       });
     }
 
-    if (includePoolShare) {
+    if (includeSingleLiquidityPoolShare) {
       options.push({
         id: "pool_share",
         label: "Liquidity Pool Share",
--- a/src/components/FormElements/AssetPicker.tsx
+++ b/src/components/FormElements/AssetPicker.tsx
@@ -38,7 +38,7 @@ type AssetPickerProps = {
   fitContent?: boolean;
   includeNative?: boolean;
   includeLiquidityPoolShares?: boolean;
-  includePoolShare?: boolean;
+  includeSingleLiquidityPoolShare?: boolean;
 };
 
 export const AssetPicker = ({
--- a/src/components/FormElements/AssetPicker.tsx
+++ b/src/components/FormElements/AssetPicker.tsx
@@ -5,6 +5,7 @@ import { ExpandBox } from "@/components/ExpandBox";
 import { RadioPicker } from "@/components/RadioPicker";
 import { PubKeyPicker } from "@/components/FormElements/PubKeyPicker";
 import { LiquidityPoolShares } from "@/components/FormElements/LiquidityPoolShares";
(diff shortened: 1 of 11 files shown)


src/components/FormElements/AssetPicker.tsx
```diff
@@ -5,29 +5,40 @@ import { ExpandBox } from "@/components/ExpandBox";
...
+      | AssetObjectValue
+      | AssetPoolShareObjectValue
+      | AssetSinglePoolShareValue
+      | undefined,
   ) => void;
   assetInput: "issued" | "alphanumeric";
   fitContent?: boolean;
   includeNative?: boolean;
   includeLiquidityPoolShares?: boolean;
+  includePoolShare?: boolean;
```
quietbits, line 41: Maybe rename it something like this to differentiate it from the above `includeLiquidityPoolShares`?
```suggestion
  includeSingleLiquidityPoolShare?: boolean;
```

src/components/FormElements/ConfigSettingIdPicker.tsx
```diff
@@ -0,0 +1,105 @@
...
+          const val = e.target.value as ConfigSettingIdType;
+
+          if (val) {
+            onChange(val);
+          } else {
+            onChange(undefined);
+          }
+        }}
+      >
+        <option value="">Select config setting id</option>
```
quietbits, line 95: To match other options
```suggestion
        <option value="">Select Config Setting ID</option>
```

src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -0,0 +1,474 @@
...
+        error: `Unable to decode JSON as LedgerKey: ${e}`,
+      };
+    }
+  };
+
+  const reset = () => {
+    setFormError({});
+    setLedgerKeyJsonString("");
+    onChange("");
+  };
```
quietbits, line 171: Should we reset error as well?

```suggestion
  const reset = () => {
    setFormError({});
    setLedgerKeyJsonString("");
    setLedgerKeyXdrError("")
    onChange("");
  };
```

quietbits (approved):
:tada:

It looks like we're not validating submit for required fields (the Submit button is always enabled). If adding that validation is not too complicated, maybe we can do that in a follow-up PR.

PR #987, opened 2024-08-27
Add a transaction cli-sign page

elizabethengelman:
This is a work in progress that I'd love some eyes on to help me think through the approach. For now I am just focusing on allowing the CLI to open Lab with the tx XDR already imported, and ready to sign.

- To sign a tx on the CLI side, we have easy access to the `rpc-url` and the `network-password`.
- To sign a tx in Lab, we need a Network object:
    
    ```
     Network = {
        id: NetworkType;
        label: string;
        horizonUrl: string;
        rpcUrl: string;
        passphrase: string;
    }
    ```

- Instead of adding additional network information to CLI, my current approach is to pass the tx xdr, and the configured network_passphrase from the CLI to Lab via query string params. We should be able to look up the Network by the passphrase based on the [NetworkOptions](https://github.com/stellar/laboratory/blob/d8837f5b3a9838c706d8521c766e9f715514713e/src/constants/settings.ts#L14).
- I first tried to just pass the network passphrase in the query string like this: `http://localhost:3000/transaction/sign?$=network$passphrase=Test SDF Network /; September 2015;&transaction$sign$activeView=overview&importXdr=AAAAAgAAAAC...=;;`
- But, if we don’t pass the network id in the `/transaction/sign` query string, Lab will use a default network that is not necessarily the one that the user configured with their CLI command. This behavior makes sense to me, but I wasn’t sure of the best way to get around the auto network selection, and instead use the network passphrases passed in to lookup which network to use.
    - The network selector functionality happens before we get to the `transaction/sign` page. So by the time we get to the sign tx page, the query string & store have been updated to include the newly chosen network. So we no longer have the network passphrase information the user sent from the CLI to look up the correct network.
    - example
        - when we visit `/transaction/sign?$=network$passphrase=Test SDF Network /; September 2015;&transaction$sign$activeView=overview&importXdr=AAAAAgAAAAC...=;;` 
        - the network selector will update it to `/transaction/sign?$=network$id=futurenet&label=Futurenet&horizonUrl=https:////horizon-futurenet.stellar.org&rpcUrl=https:////rpc-futurenet.stellar.org&passphrase=Test%20SDF%20Future%20Network%20/;%20October%202022;&transaction$sign$activeView=overview&importXdr=AAAAAgAAAAC...0=;;`
- A couple ideas I had to get around this:
    - Figure out what that assumed network id should be in the CLI, based on the network passphrase the user has configured and pass that along in the Lab request query string. But I wonder if it makes sense to do this in Lab instead since we already have the full network settings to do the look up in [NetworkOptions](https://github.com/stellar/laboratory/blob/d8837f5b3a9838c706d8521c766e9f715514713e/src/constants/settings.ts#L14)
    - Add another query param when sending the request to `/transaction/sign` that wouldn’t get overwritten in the NetworkSelector, but this felt like it wasn’t the right call. We already have the info we need.
    - Have an intermediary page `/transaction/cli-sign` that has a manual Network selector, and then redirects to `/transaction/sign. 

The last option is what I am currently doing in this PR, though it doesn't seem ideal. Curious if anyone has any thoughts on how to handle this. I haven't used zustand-querystring before, so I may be missing some option with that lib to handle this better. 

Sidenote: I also noticed that some of the weird URL escaping that zustand-querystring may be fixed in more recent release, but that seems like something we can address in follow up PRs.


src/components/NetworkByPasswordProvider.tsx
```diff
@@ -0,0 +1,42 @@
+"use client";
+
+import { NetworkOptions } from "@/constants/settings";
+import { useStore } from "@/store/useStore";
+import { useSearchParams } from "next/navigation";
+import { useEffect } from "react";
+
+// Component to set Network and Xdr if only password is given in query string
+export const NetworkByPasswordProvider = ({
```
elizabethengelman, line 9: I'd love some feedback/other ideas on the name of this component.

src/app/layout.tsx
```diff
@@ -1,12 +1,13 @@
...
+import React from "react";
 
 import { LayoutMain } from "@/components/layout/LayoutMain";
 import { QueryProvider } from "@/query/QueryProvider";
 import { StoreProvider } from "@/store/StoreProvider";
 
-import "@stellar/design-system/build/styles.min.css";
+import { NetworkByPasswordProvider } from "@/components/NetworkByPasswordProvider";
 import "@/styles/globals.scss";
+import "@stellar/design-system/build/styles.min.css";
```
quietbits, line 10: Let's keep the Design System import before the `global.css` to make sure the global overrides work.

src/app/(sidebar)/transaction/cli-sign/page.tsx
```diff
@@ -0,0 +1,18 @@
...
+export default function CliSignTransaction() {
+  const { network, transaction } = useStore();
+  const { sign } = transaction;
+  const router = useRouter();
+  useEffect(() => {
+    router.push(Routes.SIGN_TRANSACTION);
+  }, [sign.importXdr, network, router]);
+
+  return <Loader />;
+}
```
quietbits, line 18: I like this idea of using the route to do the redirect. :raised_hands: Do you think we could move `NetworkByPasswordProvider` logic here and remove it completely from `src/app/layout.tsx`. That way, everything related to CLI sign transaction would be handled in one place.

elizabethengelman: Ah, great idea! I'll get an update pushed shortly

Ifropc:
Functionally and code-vise LGTM!

src/app/(sidebar)/transaction/cli-sign/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+
+    if (xdr) {
+      transaction.updateSignActiveView("overview");
+      transaction.updateSignImportXdr(xdr);
+    }
+
+    router.push(Routes.SIGN_TRANSACTION);
+  }, [sign.importXdr, network, router, searchParams, updateIsDynamicNetworkSelect, selectNetwork, transaction]);
```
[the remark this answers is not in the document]
elizabethengelman: Good call! I think my IDE was being a little too generous with the deps... I'll update this shortly.

PR #1014, opened 2024-09-11
Show a message on top of RPC pages when the RPC URL is not input

jeesunikim:
https://github.com/user-attachments/assets/1dfad774-49d1-4b2c-a60b-0d0ad34736d8


src/components/NetworkSelector/index.tsx
```diff
@@ -194,6 +197,10 @@ export const NetworkSelector = () => {
       const latestData = getData();
 
       selectNetwork(latestData);
+
+      // also update the network setting for endpoints
+      updateNetwork(latestData);
```
jeesunikim, line 202: added this to update its global endpoints network setting so that users would see the change right away vs. refresh the page to see the updated endpoint

src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
```diff
@@ -902,6 +910,14 @@ export default function Endpoints() {
       <Card>
         <form className="PageBody" onSubmit={handleSubmit}>
           {renderEndpointUrl()}
+
+          {/* display a missing request url banner if requestUrl is empty */}
+          {!requestUrl ? (
+            <Banner icon={<Icon.AlertTriangle />} variant="warning">
+              Set a {endpointNetwork.label} RPC URL in order to submit
+            </Banner>
```
Comment (quietbits) [inline src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:918]: We should use `Notification` or `Alert` for inline messages. `Banner` should be used only on the top level (top or bottom of the page) and not inline (for example, the network reset message).

`Notification` and `Alert` don't have a yellow background. Ideally, we shouldn't customize SDS components and use them as is. But we can check with Design how important it is to make this background yellow and update it locally in CSS if needed.

Comment (sdfcharles) [reply src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:918]: could we add yellow bg to `notification` and `alert` components in SDS instead of locally? i think it's important in this case to have the yellow. we'll also need this for other cases as well.

Comment (quietbits) [reply src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:918]: Yes, we can update SDS to add the filled background option. :+1:

Comment (jeesunikim) [reply src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:918]: @sdfcharles, are you sure you want a background for the `warning` variant on `Alert`? None of the `Alert` variants including `error` variant has a specific background color. For `<notification/>`, did you mean to add yellow background by default for the 'warning' variant?

Comment (jeesunikim) [reply src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:918]: https://github.com/stellar/stellar-design-system/pull/258

quietbits (approved):

PR #1062, opened 2024-09-25
Build Transaction: fix Set Trust Line Flags flag selection

quietbits:


ElliotFriend (approved):
not entirely sure about the code, since I'm unfamiliar with how the new lab is constructed. **BUT**, in the preview it's looking like everything is working the way it should be! 👍🏻

jeesunikim (approved):

PR #1069, opened 2024-09-27
Build Transaction: set default value for offer_id in Manage Sell and Buy Offer operations

quietbits:
- Match the behavior of Stellar SDK, where by default, `offer_id` is set to `0` in Manage Sell/Buy Offer operations.
- Not adding `optional` to the field because it's required in the API, which might create confusion.


janewang:
This makes sense to set offer ID to default value, thank you for making this change!

jeesunikim (approved):

PR #1082, opened 2024-10-03
Support RPC's new flag to enable response in JSON

jeesunikim:
**getLedgerEntries before:**
getLedgerEntries-before

**getLedgerEntries after:**
getLedgerEntries-after

**getTransactions before:**
gettransaction-before

**getTransactions after:**
gettransactions


quietbits (approved):

src/components/formComponentTemplateEndpoints.tsx
```diff
@@ -906,6 +906,31 @@ export const formComponentTemplateEndpoints = (
...
+        }) => (
+          <Select
+            key={id}
+            id={id}
+            fieldSize="md"
+            label="XDR Format"
+            value={templ.value || ""}
+            onChange={templ.onChange}
+          >
+            <option id="base64" value="base64" selected>
```
quietbits, line 924: Nit: we don't need `selected` attribute here. It will be selected by default as the first option on the list.

quietbits (approved):

PR #1084, opened 2024-10-07
Share transactions

quietbits:
image

image


jeesunikim (approved):
ty!

janewang:
Hi @quietbits I tested sharing transaction and I think the XDR is not being carried to the new tab. Please let me know if you see the same. 

Uploading share-tx-not-quite-working.mov…

PR #1095, opened 2024-10-09
 Improve XDR Possible Type UX

jeesunikim:
Screenshot 2024-10-09 at 4 50 28 PM

xdr used is `AAAAAAAAAADJrq4b4AopDZibkeBWpDxuWKUcY4FUUNQdIEF3Nm9dkQ==` (ledger key xdr)

### DIFF
--- a/src/components/XdrTypeSelect/index.tsx
+++ b/src/components/XdrTypeSelect/index.tsx
@@ -102,6 +102,16 @@ export const XdrTypeSelect = ({ error }: XdrTypeSelectProps) => {
               <OptionItem key={`popular-${p}`} option={p} />
             ),
           )}
+          {guessedTypes ? (
+            <>
+              <OptionItem sectionTitle="Possible Types" />
+              {guessedTypes.map((o) => (
+                <>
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
👏🏻

src/components/XdrTypeSelect/index.tsx
```diff
@@ -102,6 +102,16 @@ export const XdrTypeSelect = ({ error }: XdrTypeSelectProps) => {
               <OptionItem key={`popular-${p}`} option={p} />
             ),
           )}
+          {guessedTypes ? (
+            <>
+              <OptionItem sectionTitle="Possible Types" />
+              {guessedTypes.map((o) => (
+                <>
+                  <OptionItem key={`guessed-${o}`} option={o} />
+                </>
```
quietbits, line 111: I don't think we need to wrap it in a fragment here.

```suggestion
                <OptionItem key={`guessed-${o}`} option={o} />
```

quietbits (approved):

PR #1097, opened 2024-10-10
Sign transaction updates

quietbits:
Updated sign transaction logic:
- Have arrays of decorated signatures for every signature type (secret key, hardware wallet, extension wallet).
  - Secret key and hardware wallet already provide signatures.
  - Extension wallet provides signed transaction XDR. We use it to extract the last signature. If, in the future, extension wallets can return signatures, it will be an easy update.
- Every signature type signs the imported transaction (without any newly added signatures).
- If anything changes in the signature type's array, that signature type is removed. For example, if a transaction is signed with two secret key signatures, then one signature is removed or updated, the secret key signature will be removed from the transaction (other types are not affected).
- On every change in signature type arrays, the transaction is signed by going through all signature type arrays and adding signatures that are currently in the arrays. This way, we're guaranteed always to have the most up-to-date list of signatures. The performance will be fine because we can add 20 signatures to a transaction at most.
- This approach should scale well if we need to add more signature types in the future.
- The current UI allows only a single hardware wallet or extension wallet signature. We can support adding multiple if needed.


quietbits:
@stellar/security-team, could you please do a sanity check on this Lab PR? :pray:

aristidesstaffieri (approved):
lgtm. I don't think I have the context to know much about the impact of the MultiPicker changes but the signature helpers and functionality look good to me.

src/hooks/useSignWithExtensionWallet.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  useEffect(() => {
+    if (isEnabled) {
+      reset();
+      signTx();
+    }
+  }, [isEnabled, signTx]);
+
+  useEffect(() => {
+    reset();
```
jeesunikim, line 119: is `isClear` supposed to be inside of this `useEffect`? I don't see it being used

jeesunikim:
Screenshot 2024-10-22 at 5 44 17 PM
should we remove the old `transaction signed!` message if any of the signature inputs are updated

jeesunikim (approved):

quietbits:
> Screenshot 2024-10-22 at 5 44 17 PM should we remove the old `transaction signed!` message box if any of the signature inputs get updated (without pressing the button)

I'm not sure. The user won't successfully submit the same transaction with different signatures, right (sequence number will be wrong, for example)? It may be good to have a reminder that it's already been submitted. What do you think @janewang @sdfcharles ? :pray:

PR #1100, opened 2024-10-15
Sign Transaction: add signature

quietbits:
> [!NOTE]
> Based on https://github.com/stellar/laboratory/pull/1097

image

image


src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -627,6 +788,82 @@ export const Overview = () => {
...
+                          e.target.value,
+                          "publicKey",
+                          idx,
+                        )
+                      }
+                    />
+
+                    <TextPicker
+                      id={`${idx}-tx-sig-b64sig`}
+                      placeholder="B64 encoded signature"
```
leighmcculloch, line 817: What does b64 mean? I thought it might mean base64, but in the screenshot looks like it is hex.

If it's base64, I think we should use the full term:
```suggestion
                      placeholder="Base64 encoded signature"
```

If it's hex, I think we should use a descriptive term:
```suggestion
                      placeholder="Hex encoded 64-byte ed25519 signature"
```

leighmcculloch (commented):
Nice. One comment inline about "b64".

src/app/(sidebar)/transaction/sign/components/Overview.tsx
quietbits: It is hex. Thanks, @leighmcculloch !

jeesunikim (approved):

PR #1105, opened 2024-10-18
Build Transaction: time bounds timestamp human readable date

quietbits:
image

image

### DIFF
--- a/src/helpers/formatEpochToDate.ts
+++ b/src/helpers/formatEpochToDate.ts
@@ -9,7 +9,8 @@ export const formatEpochToDate = (epoch: number) => {
     minute: "numeric",
     second: "numeric",
     hourCycle: "h24",
-    timeZoneName: "longOffset",
+    timeZone: "utc",
+    timeZoneName: "short",
   });
 
   return dateTimeFormatter.format(date);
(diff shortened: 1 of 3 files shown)


janewang:
I think it's better if it's UTC. 

> By international convention, UTC is equivalent to GMT, however their measurements differ: GMT is measured from midday, whereas UTC is measured from midnight. UTC is the legal basis of time in the world.

quietbits:
> I think it's better if it's UTC.
> 
> > By international convention, UTC is equivalent to GMT, however their measurements differ: GMT is measured from midday, whereas UTC is measured from midnight. UTC is the legal basis of time in the world.

@janewang Sure, we can use the UTC timezone. The design showed GMT.

src/helpers/formatEpochToDate.ts
```diff
@@ -0,0 +1,17 @@
...
+  const dateTimeFormatter = new Intl.DateTimeFormat("en-US", {
+    weekday: "short",
+    month: "short",
+    day: "numeric",
+    year: "numeric",
+    hour: "numeric",
+    minute: "numeric",
+    second: "numeric",
+    hourCycle: "h24",
+    timeZone: "utc",
```
jeesunikim, line 12: assuming it doesn't matter whether this is capitalized or not ([mdn doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat))

jeesunikim (approved):

src/helpers/formatEpochToDate.ts
quietbits: Yes, both seem to be correct values. Thanks for checking! :raised_hands:

PR #1115, opened 2024-10-23
Fix generated links

quietbits:
Fixes generated links to view endpoints and XDR in JSON response. The issue was when there were other params in the URL (for example, transaction or XDR).


jeesunikim (approved):
much easier to understand! ty!

PR #1119, opened 2024-10-29
Mask secret key inputs + some cleanup

quietbits:
- Use `LabelHeading` component instead of `Label` in places where it's not associated with an input (`htmlFor` is empty).
- Small Eslint fixes.
- Mask secret key inputs on the Sign Transaction page.

image

### DIFF
--- a/src/app/(sidebar)/transaction/sign/components/Overview.tsx
+++ b/src/app/(sidebar)/transaction/sign/components/Overview.tsx
@@ -1,14 +1,7 @@
 "use client";
 
 import { useEffect, useState } from "react";
-import {
-  Card,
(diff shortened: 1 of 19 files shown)


src/app/(sidebar)/transaction/sign/components/Overview.tsx
```diff
@@ -684,6 +678,7 @@ export const Overview = () => {
                 autocomplete="off"
                 useAutoAdd
                 note="Paste a secret key to add an additional signer"
+                isPassword
```
jeesunikim, line 681: nice

jeesunikim (approved):

PR #1138, opened 2024-11-06
Account: Saved Keypairs

quietbits:
image

image

image

### DIFF
--- a/src/app/(sidebar)/account/saved/page.tsx
+++ b/src/app/(sidebar)/account/saved/page.tsx
@@ -1,0 +1,193 @@
+"use client";
+
+import { useCallback, useEffect, useState } from "react";
+import { Alert, Text, Card, Input, Icon, Button } from "@stellar/design-system";
+
+import { Box } from "@/components/layout/Box";
+import { InputSideElement } from "@/components/InputSideElement";
+import { SaveKeypairModal } from "@/components/SaveKeypairModal";
+
+import { useStore } from "@/store/useStore";
+import { localStorageSavedKeypairs } from "@/helpers/localStorageSavedKeypairs";
+import { NetworkOptions } from "@/constants/settings";
+import { useIsTestingNetwork } from "@/hooks/useIsTestingNetwork";
+
+import { NetworkType, SavedKeypair } from "@/types/types";
+
+export default function SavedKeypairs() {
+  const { network, selectNetwork, updateIsDynamicNetworkSelect } = useStore();
+
+  const [savedKeypairs, setSavedKeypairs] = useState<SavedKeypair[]>([]);
+  const [currentKeypairTimestamp, setCurrentKeypairTimestamp] = useState<
+    number | undefined
+  >();
+
+  const IS_TESTING_NETWORK = useIsTestingNetwork();
+
+  const updateSavedKeypairs = useCallback(() => {
+    const keypairs = localStorageSavedKeypairs
+      .get()
+      .filter((s) => s.network.id === network.id);
+    setSavedKeypairs(keypairs);
+  }, [network.id]);
+
+  useEffect(() => {
+    updateSavedKeypairs();
+  }, [updateSavedKeypairs]);
+
+  const SavedKeypair = ({ keypair }: { keypair: SavedKeypair }) => {
+    return (
+      <Box gap="sm">
+        <Input
+          id={`saved-kp-${keypair.timestamp}-name`}
+          fieldSize="md"
+          value={keypair.name}
+          readOnly
+          leftElement="Name"
+          rightElement={
+            <InputSideElement
+              variant="button"
+              placement="right"
+              onClick={() => {
+                setCurrentKeypairTimestamp(keypair.timestamp);
+              }}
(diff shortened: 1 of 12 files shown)


janewang:
@quietbits On the Save Keypair page, should we follow the same pattern as on the Create Account Keypair page  and hide the secret key as ********. Wdyt?

quietbits:
> @quietbits On the Save Keypair page, should we follow the same pattern as on the Create Account Keypair page and hide the secret key as ********. What do you think?

@janewang , yes, we can mask the secret key. :+1: 

> 
> Separately, when my mouse hovers over the public key, my cursor turns to this 🚫 . I think it should be the same as the Create Account Keypair page.

The input fields are disabled, that's why you see the 🚫 icon on hover. We can change that to read-only, but the design has the disabled style. It's an easy update.


image

janewang:
> The input fields are disabled, that's why you see the 🚫 icon on hover. We can change that to read-only, but the design has the disabled style. It's an easy update.

I think the design should be consistent in saved keypairs and in create keypairs. Thank you for the update!

jeesunikim:
I think we should reset the `generate keypair` inputs when switching from `mainnet` to `futurenet`. its public key input stays intact without private key input.

https://github.com/user-attachments/assets/c54029ef-2c8a-44f8-bda6-be4106133778

jeesunikim:
`Fund Account` and `Saved Keypairs` styling on mainnet should match

saved-keypairs-styling

friendbot-styling

src/app/(sidebar)/account/saved/page.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+import { useStore } from "@/store/useStore";
+import { localStorageSavedKeypairs } from "@/helpers/localStorageSavedKeypairs";
+import { NetworkOptions } from "@/constants/settings";
+import { useIsTestingNetwork } from "@/hooks/useIsTestingNetwork";
+
+import { NetworkType, SavedKeypair } from "@/types/types";
+import { arrayItem } from "@/helpers/arrayItem";
+
+export default function SavedKeypairs() {
```
jeesunikim, line 19: curious - when should we use `export default function` vs. `export const`. in vibrant days, we preferred using `const` for exporting vs. default ([product convention](https://github.com/stellar/product-conventions/blob/master/STYLEGUIDE.md#all-files-should-only-export-named-variables-not-defaults)).

jeesunikim: (I am not suggesting that we follow the convention btw, I'm just curious)

src/app/(sidebar)/account/saved/page.tsx
```diff
@@ -0,0 +1,238 @@
...
+          placement="inline"
+          actionLabel={`Switch to ${otherNetworkLabel}`}
+          onAction={() => {
+            const newNetworkId =
+              network.id === "testnet" ? "futurenet" : "testnet";
+
+            getNetworkById(newNetworkId);
+          }}
+        >
+          {`You must switch your network to ${otherNetworkLabel} in order to see those saved
```
jeesunikim, line 138: can we just do the following? 
`You must switch your network to {otherNetworkLabel} in order to see those saved keypairs. This feature is only available on Futurenet and Testnet for security reasons.`

src/components/SavedItemTimestampAndDelete.tsx
```diff
@@ -0,0 +1,30 @@
...
+  timestamp: number;
+  onDelete: () => void;
+}) => {
+  return (
+    <>
+      <Box gap="sm" direction="row" align="center" justify="end">
+        <Text
+          as="div"
+          size="xs"
+        >{`Last saved ${formatTimestamp(timestamp)}`}</Text>
```
jeesunikim, line 18: can we do `Last saved {formatTimestamp(timestamp)}` instsead?

quietbits:
> I think we should reset the `generate keypair` inputs when switching from `mainnet` to `futurenet`. its public key input stays intact without private key input.
> 
>  mainnet-to-futurenet.mov

Good idea, we should reset the keypair when the network changes. [Issue created](https://github.com/stellar/laboratory/issues/1143).

src/app/(sidebar)/account/saved/page.tsx
quietbits: In this case ([page.tsx](https://nextjs.org/docs/14/app/api-reference/file-conventions/page)), we need to follow the Next.js convention and any other framework level file ([examples](https://nextjs.org/docs/14/app/api-reference/file-conventions)). I agree that we should use named exports when possible. :+1:

src/app/(sidebar)/account/saved/page.tsx
quietbits: Yes, that works, too. In cases like this, I prefer formatting it as a string because it's more clear and consistent in the code editor what it is. This is just my preference, nothing else. :blush: If you feel strongly about it, I can update it.

image

vs

image

jeesunikim (approved):

PR #1141, opened 2024-11-07
add copy button for the json payload for RPC

jeesunikim:
**Summary:**
- Add a copy button for the json payload for RPC on **saved requests** and **rpc method**
- Added `e.preventDefault();` for `Copy JSON` button. Without it, it'd submit the form

saved-requests-rpc

rpc-method


quietbits (approved):
We could create a shared component to use in both places.

PR #1147, opened 2024-11-08
Saved Keypairs: show balance or fund account button

quietbits:
image


src/query/useFriendBot.ts
```diff
@@ -20,7 +20,7 @@ export const useFriendBot = ({
       : "https://friendbot.stellar.org";
 
   const query = useQuery({
-    queryKey: ["friendBot", key],
+    queryKey: ["friendBot", publicKey, key],
```
quietbits, line 23: We need this to be unique to fetch account info on the Saved Keypairs page.

jeesunikim: thank you

jeesunikim (approved):

PR #1159, opened 2024-11-14
Upgrade Stellar SDK to v13.0.0-rc.2

quietbits:


Shaptic (approved):
Poked around in the endpoints that changed, everything seems to be in order!

Shaptic:
I tagged off the stable build since everything seemed to check out, so please use that, instead :)

jeesunikim (approved):

PR #1172, opened 2024-11-21
Sign Transaction UI Test

jeesunikim:
Sign Transaction UI Test
- [x] flow for the XDR with one operation that also checks its final signed xdr with two signatures signed from secret keys and signature
- [x] flow for the XDR with 3 operations that also checks its final signed xdr with two signatures signed from secret keys and signature
- [x] flow for the non transaction enveloper XDR

For `Ledgers` and `Wallet Extension`, it checks its happy path - making sure the dropdown works and trezor to open the new link. Ledger opens an extension so i skipped the new tab flow for ledger. Wallet Extension checks to make sure the modal pops up and it lists 6 wallets that it supports.

### DIFF
--- a/tests/signTransactionPage.test.ts
+++ b/tests/signTransactionPage.test.ts
@@ -18,6 +18,9 @@ test.describe("Sign Transaction Page", () => {
     const secretKeysView = page.getByTestId("sign-tx-secretkeys");
     const hardwareView = page.getByTestId("sign-tx-hardware");
     const walletExtView = page.getByTestId("sign-tx-wallet-ext");
+    const sigExtView = page.getByTestId("sign-tx-signature");
+    const validationView = page.getByTestId("sign-tx-validation-card");
+    const signedXdr = page.getByTestId("validation-card-response");
 
     // Import Screen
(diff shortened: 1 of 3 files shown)


tests/signTransactionPage.test.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    await expect(
+      page.getByRole("heading", { name: "Connect a Wallet" }),
+    ).toBeVisible();
+
+    // Wallet Extension to display 6 wallets
+    await expect(page.getByRole("listitem")).toHaveCount(6);
+
+    // Exit out of the wallet extension modal
+    await page.click("body", { position: { x: 10, y: 10 } });
```
quietbits, line 168: This test will break if the close button position changes, but style changes shouldn't affect the tests. We can't use a class selector? Something like `page.locator(".wallets-header__button")`.

quietbits (approved):

tests/signTransactionPage.test.ts
jeesunikim: it's more of a clicking outside of the modal

PR #1173, opened 2024-11-21
Build Transaction UI Tests

quietbits:
Sorry for the long PR, but it didn't make sense to split the operations into multiple PRs.

I'm testing the happy path for all of the operations and validation tests only where the fields have not already been tested. For example, we can test a public key input once because the same component will be used everywhere where the input is a public key.

### DIFF
--- a/tests/buildTransaction.test.ts
+++ b/tests/buildTransaction.test.ts
@@ -813,6 +813,234 @@ test.describe("Build Transaction Page", () => {
         ).toBeVisible();
       });
     });
+
+    // Account Merge
+    test.describe("Account Merge", () => {
+      test("Happy path", async ({ page }) => {
+        const { operation_0 } = await selectOperationType({
+          page,
+          opType: "account_merge",
+        });
+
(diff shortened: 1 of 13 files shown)


tests/buildTransaction.test.ts
```diff
@@ -0,0 +1,1470 @@
...
+      );
+
+      await lowerTimeBoundsInput.fill("1729270000");
+      await expect(lowerTimeBoundsInput.locator("+ div")).toHaveText(
+        "Fri, Oct 18, 2024, 16:46:40 UTC",
+      );
+
+      await upperTimeBoundsInput.fill("1729517340");
+      await expect(upperTimeBoundsInput.locator("+ div")).toHaveText(
+        /Mon, Oct 21, 2024, 13:29:00 UTC/,
```
jeesunikim, line 122: how come this one uses '/' regex

quietbits: Because the UI also includes the "Set to 5 min from now" text at the end of that div. I thought it would be confusing to add that text to the test. Regex makes it cleaner.

image

jeesunikim (approved):

PR #1190, opened 2024-12-09
Submit Transaction UI Test

jeesunikim:
Covers:
- Valid XDR with a Failed 400 Submission Flow
-- Submit via Horizon
-- Submit via RPC
- Invalid XDR Flow:
-- display an error with a random text
-- display an error with a valid XDR that is not TX Envelope
- Simulate Transaction Flow
- Save Transaction Flow
- Valid XDR with a Successful Submission Flow
-- Submit via Horizon
-- Submit via RPC


tests/submitTransactionPage.test.ts
```diff
@@ -0,0 +1,611 @@
...
+      expect(methods).toEqual(["via RPC", "via Horizon"]);
+
+      await submitMethodsDropdown.getByText("via RPC").click();
+
+      const submitTxBtn = page.getByRole("button", {
+        name: "Submit transaction",
+      });
+
+      // Mock the rpc submission api call
+      await page.route("**", async (route) => {
```
quietbits, line 240: Should we use `https://soroban-testnet.stellar.org/` instead here?

quietbits (approved):

tests/submitTransactionPage.test.ts
jeesunikim: yes! it was returning an error with it when I had the response in place but now that I've removed it, I will put it back

PR #1191, opened 2024-12-09
Fee Bump UI Tests

quietbits:


tests/feeBumpPage.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    test("Sign in Transaction Signer", async ({ page }) => {
+      await page.getByLabel("Base Fee").fill(BASE_FEE);
+      await page.getByText("Sign in Transaction Signer").click();
+
+      await expect(page.locator("h1")).toHaveText("Transaction Overview");
+      await expect(page.getByLabel("Transaction Envelope XDR")).toHaveText(
+        MOCK_XDR,
+      );
```
quietbits, line 61: Here, we're just making sure that the correct page loads. There is no need to repeat the tests we've done on that page separately. The same for the XDR viewer page below.

jeesunikim (approved):

PR #1219, opened 2025-01-07
Contract Explorer: Contract version history

quietbits:
Smart contract with multiple versions on Mainnet: `CCBVCCNPIFMCXW7S3GBM6IOQBD5TEUCSQ6WWJGB5VIXZCRVJJQHQQE23`

image

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -23,8 +23,8 @@ export default defineConfig({
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
-  /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  /* Retry */
+  retries: process.env.CI ? 2 : 1,
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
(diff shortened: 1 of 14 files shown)


playwright.config.ts
```diff
@@ -23,8 +23,8 @@ export default defineConfig({
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
-  /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  /* Retry */
+  retries: process.env.CI ? 2 : 1,
```
quietbits, line 27: This makes running tests locally more consistent on my machine.

src/app/(sidebar)/smart-contracts/contract-explorer/components/VersionHistory.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  if (!versionHistoryData) {
+    return (
+      <Text as="div" size="sm">
+        No version history
+      </Text>
+    );
+  }
+
+  type TableHeader = {
```
quietbits, line 55: We'll move the `Table` to its own file when implementing another data table. That way, we'll have a better idea of the features we need to support.

jeesunikim (approved):

PR #1226, opened 2025-01-17
[Build Tx] Add a Soroban Operation - Extend TTL

jeesunikim:
**Add Soroban Operation - Extend TTL**

includes:
- displaying only one operation when soroban operation is selected
- add a `fetch minimum resource fee` button that simulates tx and calculates the resource fee
- it follows how cli is operating using `key-xdr` ([cli doc](url)) (as recommended by dima)
- I used auth soroban example to invoke a contract and use stellat js sdk to get the key xdr ([example](https://github.com/ElliotFriend/ye-olde-guestbook/blob/main/src/lib/server/getLedgerEntries.ts#L45))
- On sign transaction, method dropdown to auto select `rpc` if it's a soroban tx
- separate the soroban logic from classic in `Operation`
- `createStore` to include a separate soroban object that stores `xdr` and `params`
- added tests
- Updated `saved` page

### DIFF
--- a/src/helpers/sorobanUtils.ts
+++ b/src/helpers/sorobanUtils.ts
@@ -10,7 +10,7 @@ import {
 } from "@stellar/stellar-sdk";
 
 import { TransactionBuildParams } from "@/store/createStore";
-import { SorobanOpType } from "@/types/types";
+import { SorobanOpType, TxnOperation } from "@/types/types";
 
 export const isSorobanOperationType = (operationType: string) => {
   return [
@@ -87,12 +87,12 @@ export const getSorobanDataResult = ({
 export const buildSorobanTx = ({
   sorobanData,
   params,
-  sorobanParams,
+  sorobanOp,
   networkPassphrase,
 }: {
   sorobanData: xdr.SorobanTransactionData;
   params: TransactionBuildParams;
-  sorobanParams: any;
+  sorobanOp: TxnOperation;
   networkPassphrase: string;
 }) => {
   // decrement seq number by 1 because TransactionBuilder.build()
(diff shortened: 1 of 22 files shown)


src/store/createStore.ts
```diff
@@ -109,13 +109,20 @@ export interface Store {
   // Transaction
   transaction: {
     build: {
-      params: TransactionBuildParams;
+      // @TODO wrap it within classic object
```
jeesunikim, line 112: @quietbits what do you think of having a separate classic object?

jeesunikim: done!

src/helpers/sorobanUtils.ts
```diff
@@ -0,0 +1,174 @@
...
+  //   ReadableByteStreamController,
+}: {
+  readOnlyXdrLedgerKey?: xdr.LedgerKey[];
+  readWriteXdrLedgerKey?: xdr.LedgerKey[];
+  resourceFee: string;
+}) => {
+  // one of the two must be provided
+  if (!readOnlyXdrLedgerKey && !readWriteXdrLedgerKey) {
+    return;
+  }
```
quietbits, line 167: Should we check for length here since we're setting default values as empty arrays?

tests/buildTransaction.test.ts
```diff
@@ -1299,6 +1299,142 @@ test.describe("Build Transaction Page", () => {
...
+        await testInputError({
+          page,
+          isSorobanOp: true,
+          label: "Resource Fee (in stroops)",
+          value: "aaa",
+          errorMessage: "Expected a whole number.",
+        });
+      });
+
+      test("Check rendering betweem classic and soroban", async ({ page }) => {
```
quietbits, line 1397:
```suggestion
      test("Check rendering between classic and soroban", async ({ page }) => {
```

tests/buildTransaction.test.ts
```diff
@@ -1299,6 +1299,142 @@ test.describe("Build Transaction Page", () => {
...
+        });
+
+        const classicOperation = page.getByTestId(
+          "build-transaction-operation-0",
+        );
+
+        await expect(classicOperation).toBeVisible();
+
+        // Soroban Operation only allows one operation
+        // Add Operation button should not be visible
```
quietbits, line 1434: Looks like this comment needs to be updated.

jeesunikim:
I'll create smaller new PRs tomorrow

PR #1227, opened 2025-01-17
Contract Explorer: Contract storage

quietbits:


src/query/external/useSEContractVersionHistory.ts
```diff
@@ -38,7 +40,9 @@ export const useSEContracVersionHistory = ({
         throw `Something went wrong. ${e}`;
       }
     },
-    enabled: Boolean(networkId && contractId),
+    enabled: Boolean(isActive && networkId && contractId),
+    // Keep data for 30 seconds
```
jeesunikim, line 44: is this keep data for 30 seconds or don't call another api request for 30 seconds?

src/query/external/useSEContracStorage.ts
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        allRecords = [
+          ...allRecords,
+          ...(responseJson?._embedded?.records || []),
+        ];
+      };
+
+      try {
+        while (allRecords.length < totalEntriesCount) {
```
jeesunikim, line 63: a blend smart contract `CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP` has `5987 entries`. Upon initial fetch, there is some loading time. What do you think of batch fetch data call instead fetch data all entries?

blend

jeesunikim (approved):

src/query/external/useSEContractVersionHistory.ts
quietbits: Yes, that's precisely what it means. I thought adding a little "cache" time for cases when users quickly go through tabs was good. This way, we won't be making too many API calls.

src/query/external/useSEContracStorage.ts
quietbits: Unfortunately, I don't think we can batch the calls right now. :disappointed: If we're doing sorting, filters, and search on the client, we need to have all the data to make it work as expected. If this becomes an issue, we might look into some clever things to do with batching, but that will be quite complex.

PR #1233, opened 2025-01-29
 Contract Explorer: Contract storage filters

quietbits:
1. The filter logic is "or", meaning that selecting more filters includes more records. We could implement "and" logic (narrowing down selection), making the filters dynamic. Adding more data to store locally might slow the UI, though.
2. We need an "Apply" button to apply the selected filters. Because we can have thousands of records, we can't make updates on the fly as the filters are selected. For now, I added a basic button, so we have "Apply" and "Clear filter" buttons.
3. What should be the logic for selecting `value` filters? I'm using object keys from decoded `value` XDR, and we might end up with numbers. For example, `CAS3FL6TLZKDGGSISDBWGGPXT3NRR4DYTZD7YOD3HMYO6LTJUVGRVEAM` (Mainnet). `key` filters are much more predictable than `value`, which can be pretty much anything.
4. How should we handle very long lists of filters? For example, `CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP` (Mainnet) has a very long list of `value` filters.
5. We need to keep checking how we handle thousands of records. On my machine, `CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP` (Mainnet) with 6k+ records at this time, seems OK. It takes a few seconds on the initial load, and then the loader is displayed when sorting. But the UI doesn't feel sluggish; it just takes longer to load the result.
6. Should we display the filtered results count?
7. I will add tests in a follow-up PR once we're all set with the UI.

image

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
@@ -9,6 +9,7 @@ import { useSEContractStorage } from "@/query/external/useSEContracStorage";
 import { formatEpochToDate } from "@/helpers/formatEpochToDate";
 import { formatNumber } from "@/helpers/formatNumber";
 import { capitalizeString } from "@/helpers/capitalizeString";
+import { decodeScVal } from "@/helpers/decodeScVal";
 
 import { useIsXdrInit } from "@/hooks/useIsXdrInit";
 
@@ -64,13 +65,74 @@ export const ContractStorage = ({
     );
   }
 
+  const parsedKeyValueData = () => {
+    return storageData.map((i) => ({
+      ...i,
+      keyJson: i.key ? decodeScVal(i.key) : undefined,
+      valueJson: i.value ? decodeScVal(i.value) : undefined,
+    }));
+  };
+
+  const parsedData = parsedKeyValueData();
+
+  const getKeyValueFilters = () => {
+    return parsedData.reduce(
+      (
+        res: {
+          key: string[];
+          value: string[];
+        },
+        cur,
+      ) => {
+        // Key
+        if (cur.keyJson && Array.isArray(cur.keyJson)) {
+          const keyFilter = cur.keyJson[0];
+
+          if (!res.key.includes(keyFilter)) {
+            res.key = [...res.key, keyFilter];
+          }
+        }
+
+        // Value
+        if (cur.valueJson && typeof cur.valueJson === "object") {
+          // Excluding keys that start with _ because on the UI structure is
+          // different. For example, for Instance type.
+          const valueFilters = Object.keys(cur.valueJson).filter(
+            (f) => !f.startsWith("_"),
+          );
+
+          valueFilters.forEach((v) => {
+            if (!res.value.includes(v)) {
+              res.value = [...res.value, v];
+            }
+          });
+        }
+
+        return res;
+      },
+      { key: [], value: [] },
+    );
+  };
+
+  const keyValueFilters = getKeyValueFilters();
+
   return (
     <DataTable
       tableId="contract-storage"
-      tableData={storageData}
+      tableData={parsedData}
       tableHeaders={[
-        { id: "key", value: "Key", isSortable: false },
-        { id: "value", value: "Value", isSortable: false },
+        {
+          id: "key",
+          value: "Key",
+          isSortable: false,
+          filter: keyValueFilters.key,
+        },
+        {
+          id: "value",
+          value: "Value",
+          isSortable: false,
+          filter: keyValueFilters.value,
+        },
         { id: "durability", value: "Durability", isSortable: true },
         { id: "ttl", value: "TTL", isSortable: true },
         { id: "updated", value: "Updated", isSortable: true },
@@ -81,22 +143,14 @@ export const ContractStorage = ({
         {
           value: (
             <div className="CodeBox">
-              <ScValPrettyJson
-                xdrString={vh.key}
-                json={vh.keyJson}
-                isReady={isXdrInit}
-              />
+              <ScValPrettyJson xdrString={vh.key} isReady={isXdrInit} />
             </div>
           ),
         },
         {
           value: (
             <div className="CodeBox">
-              <ScValPrettyJson
-                xdrString={vh.value}
-                json={vh.valueJson}
-                isReady={isXdrInit}
-              />
+              <ScValPrettyJson xdrString={vh.value} isReady={isXdrInit} />
             </div>
           ),
         },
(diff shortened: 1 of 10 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
```diff
@@ -60,18 +65,81 @@ export const ContractStorage = ({
     );
   }
 
+  const parsedKeyValueData = () => {
+    return storageData.map((i) => ({
+      ...i,
+      keyJson: i.key ? decodeScVal(i.key) : undefined,
+      valueJson: i.value ? decodeScVal(i.value) : undefined,
```
quietbits, line 72: We need to decode `key` and `value` XDR to get filter values.

janewang:
> Should we display the filtered results count? 
No, I think there is no need. Once they filter down the list, maybe the user can download

jeesunikim:
I'm getting a warning for missing `key` prop in console.
console

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
```diff
@@ -60,18 +65,81 @@ export const ContractStorage = ({
     );
   }
 
+  const parsedKeyValueData = () => {
+    return storageData.map((i) => ({
```
jeesunikim, line 69: what do you think of adding type - like `i: ContractStorageResponseItem`?

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
```diff
@@ -60,18 +65,81 @@ export const ContractStorage = ({
...
+      ...i,
+      keyJson: i.key ? decodeScVal(i.key) : undefined,
```
jeesunikim, line 71: will this be a standard data type? should we create a type for this? and use that type on line `85`?

src/helpers/processContractStorageData.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (["asc", "desc"].includes(sortByDir)) {
+      // Asc
+      sortedData = sortedData.sort((a: any, b: any) =>
+        a[sortById] > b[sortById] ? 1 : -1,
+      );
+
+      // Desc
+      if (sortByDir === "desc") {
+        sortedData = sortedData.reverse();
+      }
```
jeesunikim, line 31: for descending, instead of doing two operations by sorting via ascending then descending via `.reverse()`, can we just do it in one go? also should we include a case when there might be an equality between the value?

src/components/ScValPrettyJson/index.tsx
```diff
@@ -31,7 +33,9 @@ export const ScValPrettyJson = ({
 
   const parseJson = () => {
     try {
-      return parse(StellarXdr.decode("ScVal", xdrString)) as AnyObject;
+      return xdrString
+        ? (parse(StellarXdr.decode("ScVal", xdrString)) as AnyObject)
```
jeesunikim, line 37: 👁️ I should adopt this `parse` from `lossless json` in `getLedgerEntries`.

jeesunikim (approved):
added some questions but looks great. low key worried about big contract data entries getting larger and larger 😅

src/helpers/processContractStorageData.ts
quietbits: That's a good point. :+1: I'll update to use just one operation for both cases.

I don't think we need to do anything special with the equal cases because we don't have a secondary param to sort by if they are equal. Right now, it doesn't matter which item comes first.

src/components/ScValPrettyJson/index.tsx
quietbits: Yes, we should use this instead of the native `JSON.parse` when dealing with large numbers.

sdfcharles:
@quietbits great work! some thoughts:

`1. The filter logic is "or", meaning that selecting more filters includes more records. We could implement "and" logic (narrowing down selection), making the filters dynamic. Adding more data to store locally might slow the UI, though.`

CleanShot 2025-02-03 at 11 34 54
Yeah I think we should implement the 'and' logic here as well. In this case, the user has selected two filters but the combination of the two filters doesn't match any records. There should be copy here that "No results. Clear filters and try again."

`2. We need an "Apply" button to apply the selected filters. Because we can have thousands of records, we can't make updates on the fly as the filters are selected. For now, I added a basic button, so we have "Apply" and "Clear filter" buttons.`

Could we have the **Clear Filter** and **Apply** buttons be fixed at the bottom of the menu and the filters be scrollable? This will be clearer to the user to apply their filters instead of scrolling all the way to the bottom of a potentially long list to apply filters.

`3. What should be the logic for selecting value filters? I'm using object keys from decoded value XDR, and we might end up with numbers. For example, CAS3FL6TLZKDGGSISDBWGGPXT3NRR4DYTZD7YOD3HMYO6LTJUVGRVEAM (Mainnet). key filters are much more predictable than value, which can be pretty much anything.`

CleanShot 2025-02-03 at 11 29 20
When we show numbers in the values filter, some suggestions:
1. if its a G address or C address, truncate the string in the pill e.g. CD25...G5JY
2. wrap the pills when they go past the width of the container
3. handle any other number with a truncated pill e.g. 12345678...

`4. How should we handle very long lists of filters? For example, CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP (Mainnet) has a very long list of value filters.`

Can we set  a max-height of 320px on the filter menus and make them scrollable?

`5. We need to keep checking how we handle thousands of records. On my machine, CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP (Mainnet) with 6k+ records at this time, seems OK. It takes a few seconds on the initial load, and then the loader is displayed when sorting. But the UI doesn't feel sluggish; it just takes longer to load the result.`

CleanShot 2025-02-03 at 11 26 52
Agreed the UI loading state feels good here. Wondering if it's more of a browsabliity issue – one suggestion would be to make the page number box editable so users can access middle pages easier, in addition to clicking through one page as a time.

`6. Should we display the filtered results count?`
CleanShot 2025-02-03 at 11 38 22
Yeah, let's add that here.

jeesunikim:
Screenshot 2025-02-03 at 12 38 05 PM
`stellar.expert` link for contract address leads to `https://stellar.expert/explorer/public/account/` instead of `https://stellar.expert/explorer/public/contract/`. Example, the contract account `CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA` is linked to [public/account](https://stellar.expert/explorer/public/account/CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA) which is incorrect

PR #1241, opened 2025-02-03
[Build Operations] Update `Saved` & `Submit` Page to include Soroban Op

jeesunikim:
- `Saved` to properly save Soroban Operation
- `Submit` to update its submit button to `via RPC` if its XDR includes soroban operation
- Update `getContractIdError` to use `StrKey.isValidContract`
- Update tests


tests/mock/smartContracts.ts
```diff
@@ -1,7 +1,7 @@
 import { SAVED_ACCOUNT_1 } from "./localStorage";
 
 export const MOCK_CONTRACT_ID =
-  "C000NO6F7FRDHSOFQBT2L2UWYIZ2PU76JKVRYAQTG3KZSQLYAOKIF2WB";
+  "CAQP53Z2GMZ6WVOKJWXMCVDLZYJ7GYVMWPAMWACPLEZRF2UEZW3B636S";
```
jeesunikim, line 4: `StrKey.isValidContract` threw an error for the original contract ID so I updated it

src/app/(sidebar)/transaction/submit/page.tsx
```diff
@@ -67,6 +67,30 @@ const SUBMIT_OPTIONS = [
   },
 ];
 
+// traverse the xdr to json string and check if
+// it contains a soroban operation
+const isSorobanXdr = (xdrJsonString: string) => {
```
jeesunikim, line 72: I couldn't find other way. Do you have a suggestion other than traversing through the `xdrJsonString`?

quietbits: I can't think of anything else we could try.

quietbits (approved):

PR #1244, opened 2025-02-05
[Invoke Contract] Create a dropdown & retrieve function methods

jeesunikim:
**Summary:**
- Create an `Invoke Contract Function` option
- Add a `fetchContractFunctionMethods` helper function that will retrieve function methods using `client.spec`

Next PR will include rendering function arguments

invoke

### DIFF
--- a/src/components/FormElements/FetchContractMethodPickerWithQuery.tsx
+++ b/src/components/FormElements/FetchContractMethodPickerWithQuery.tsx
@@ -67,6 +67,8 @@ export const FetchContractMethodPickerWithQuery = ({
   };
 
   const handleFetchContractMethods = async () => {
+    setFetchError("");
+
     const contractMethods: ContractFunctionMethods =
       await fetchContractFunctionMethods({
         contractId: value,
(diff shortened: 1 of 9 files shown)


src/components/FormElements/FetchContractMethodPickerWithQuery.tsx
```diff
@@ -0,0 +1,123 @@
...
+      if (validatedContractIdError) {
+        setContractIdError(validatedContractIdError);
+      } else {
+        setContractIdError("");
+      }
+    }
+  };
+
+  const handleFetchContractMethods = async () => {
+    const contractMethods: ContractFunctionMethods =
```
quietbits, line 70: We should clear the error here as well. I got an error with one RPC URL, then changed to another that worked, but the error wasn't cleared.

quietbits (commented):
"Fetch minimum resource fee from RPC" didn't work for me. I'm not sure if it's still work in progress.

image

quietbits (approved):

PR #1250, opened 2025-02-13
Generate invoke host function inputs from JSON Spec

jeesunikim:
**Summary:**
- we are creating our own form based on JsonSchema
- For the [library](https://rjsf-team.github.io/react-jsonschema-form/) that Jane shared, if it works well with what we currently have, then we can use it. If not, then we can use it as an example and test our implementation.

**This PR includes:**

**This PR does not include:**


package.json
```diff
@@ -22,13 +22,17 @@
     "@creit.tech/stellar-wallets-kit": "^1.2.5",
     "@ledgerhq/hw-app-str": "^7.0.4",
     "@ledgerhq/hw-transport-webusb": "^6.29.4",
+    "@rjsf/core": "^5.24.3",
```
jeesunikim, line 25: included for `feature-invoke-host-function` branch so that we can use it as an example and test our implementation

src/components/JsonSchemaForm/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+          }
+          // render a button to add an item to the array
+          return (
+            <Box gap="sm" key={label}>
+              <LabelHeading size="md">{label}</LabelHeading>
+              <div>
+                <Button
+                  variant="secondary"
+                  size="md"
+                  onClick={() => {}}
```
jeesunikim, line 133: its onClick to add implementation will be added in the next PR

src/components/JsonSchemaForm/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+                handleChange(label, e.target.value);
+              }}
+            />
+          );
+
+        case "I32":
+        case "I64":
+        case "I128":
+        case "I256":
+          return null;
```
quietbits, line 115: Curious: why are we not rendering these types?

src/components/JsonSchemaForm/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  return `tuple<${typeLabels.join(",")}>`;
+};
+
+const renderTitle = (
+  name: string,
+  labels: string[],
+  fields: [string, any][],
+) => {
+  if (labels) {
```
quietbits, line 194: To make sure the array is not empty.

```suggestion
  if (labels.length) {
```

quietbits (approved):

src/components/JsonSchemaForm/index.tsx
jeesunikim: ah I was going to create a component that allows both positive and negative number. I'll do that next!

PR #1251, opened 2025-02-14
Contract Explorer: Contract spec

quietbits:
- Using [`@monaco-editor/react`](https://github.com/suren-atoyan/monaco-react) for code editor (React wrapper for [Monaco Editor](https://microsoft.github.io/monaco-editor/) that is used in VS Code).
  - NOTE: I relaxed CSP rules for `style-src` to make styles work for the Monaco Editor. The files are pulled from a CDN. I couldn't figure out how to configure Next.js to use files locally.
- Fetching wasm data using Stellar SDK from RPC.
- Small refactoring to improve code reuse.

image


janewang:
Screenshot 2025-03-05 at 9 07 15 AM
I got an axios error. See above.

PR #1267, opened 2025-03-06
Contract Spec: add download Wasm button + handle network error for better UX

quietbits:
Download WASM button

image

Warning when there is a network error

image

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractSpec.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractSpec.tsx
@@ -13,6 +13,7 @@ import { useIsXdrInit } from "@/hooks/useIsXdrInit";
 import * as StellarXdr from "@/helpers/StellarXdr";
 import { prettifyJsonString } from "@/helpers/prettifyJsonString";
 import { delayedAction } from "@/helpers/delayedAction";
+import { downloadFile } from "@/helpers/downloadFile";
 
 import { NetworkType } from "@/types/types";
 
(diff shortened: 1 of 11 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractSpec.tsx
```diff
@@ -37,6 +47,47 @@ export const ContractSpec = ({
...
+
+  useEffect(() => {
+    if (wasmBinaryError) {
+      // Automatically clear error message after 5 sec
+      delayedAction({
+        action: resetWasmBlob,
+        delay: 5000,
+      });
+    }
+  }, [resetWasmBlob, wasmBinaryError]);
```
quietbits, line 89: I couldn't find a way to clear this on specific events (unmount, inActive, etc.), so I opted to clear the message after 5 seconds. The UX seemed okay to me.

src/query/external/useSEContractInfo.ts
```diff
@@ -15,13 +16,12 @@ export const useSEContractInfo = ({
   const query = useQuery<ContractInfoApiResponse>({
     queryKey: ["useSEContractInfo", networkId, contractId],
     queryFn: async () => {
-      // Not supported networks
-      if (["futurenet", "custom"].includes(networkId)) {
+      const network = getStellarExpertNetwork(networkId);
```
quietbits, line 19: A small refactor to use a helper since we have this check in multiple places.

janewang:
Hi @quietbits What do you think having a download button also for contract spec?
Screenshot 2025-03-06 at 10 24 00 AM

Also I realized there is not a really good docs on contract spec in dev docs. I will write one by EOW.

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractSpec.tsx
```diff
@@ -37,6 +47,47 @@ export const ContractSpec = ({
...
+    refetch: fetchWasmBinary,
+  } = useSEContractWasmBinary({ wasmHash, networkId });
+
+  const resetWasmBlob = useCallback(() => {
+    queryClient.resetQueries({
+      queryKey: ["useSEContractWasmBinary", networkId, wasmHash],
+    });
+  }, [networkId, queryClient, wasmHash]);
+
+  useEffect(() => {
```
jeesunikim, line 64: can we do this onClick instead of lifecycle?

quietbits: The click triggers React Query to download the wasm binary, so we need to do something after this step happens. The other option would be to put this inside the React Query, but that would limit the usage of that query to just this.

I'm not a huge fan of this approach, but I couldn't come up with something better. I'm open to suggestions! :pray:

jeesunikim (approved):

PR #1279, opened 2025-03-17
Smart Contract List

quietbits:
- Created Smart Contract List page.
- Using StellarExpert API to get the smart contract list.
  - Using API for sorting by "Created Timestamp" prop.
- Updated `DataTable` component to support external actions.

image


janewang:
@jeesunikim could you please review this one?

jeesunikim:
should default be the latest? I see 2024 ones as the first ones
Screenshot 2025-03-19 at 11 16 05 AM

quietbits:
> should default be the latest? I see 2024 ones as the first ones. After changing it to the latest, it keeps defaulting back to the oldest whenever I switch tabs Screenshot 2025-03-19 at 11 16 05 AM

@jeesunikim , I'm following the StellarExpert's API defaults. If we don't pass the order param, this is what we get. We can force it to show the latest first if we want. @janewang, any preference?

We don't persist states between the tabs, so it will reset once you return.

janewang:
Yes, I'd prefer to show the latest first

PR #1281, opened 2025-03-18
[Invoke contract] convert nativeToScVal and use `prepareTransaction` to submit

jeesunikim:
- success case (happy path)
-- [stellar.expert link](https://stellar.expert/explorer/testnet/tx/a4efe2b07a4e093d6e7e1e0aefa05254aed0635b63a9014d8f12804dedb3c628)

https://github.com/user-attachments/assets/e17c417f-6545-4db7-9f81-d6677ae0aadc

- fail case on mainnet using the contract `CA6PUJLBYKZKUEKLZJMKBZLEKP2OTHANDEOWSFF44FTSYLKQPIICCJBE` ([link](https://stellar.expert/explorer/public/contract/CA6PUJLBYKZKUEKLZJMKBZLEKP2OTHANDEOWSFF44FTSYLKQPIICCJBE))
-- this one doesn't provide `debuginfo`

https://github.com/user-attachments/assets/a52068e4-a877-4070-995e-a04d0960283e

I need to test further with contracts to get the scenario with `debuginfo`


janewang:
@jeesunikim This is super neat!

src/components/JsonSchemaForm/index.tsx
```diff
@@ -370,44 +478,79 @@ const getTupleLabel = (items: any[]): string => {
...
 const renderTitle = (
   name: string,
+  description: string,
   labels: string[],
   fields: [string, any][],
 ) => {
-  if (labels.length) {
-    const mappedLabels = labels.map((label) => {
-      const field = fields.find(([key]) => key === label);
+  // if (labels.length) {
```
quietbits, line 485: Do we need to keep this for now?

src/query/useRpcPrepareTx.ts
```diff
@@ -0,0 +1,56 @@
...
+      transactionXdr,
+      networkPassphrase,
+      headers,
+    }: PrepareRpcTxProps) => {
+      try {
+        const transaction = TransactionBuilder.fromXDR(
+          transactionXdr,
+          networkPassphrase,
+        );
+        console.log("transaction :", transaction);
```
quietbits, line 37: `console.log`

quietbits (approved):

src/components/FormElements/ResourceFeePickerWithQuery.tsx
```diff
@@ -1,24 +1,27 @@
...
+  buildTxWithSorobanData,
   getContractDataXDR,
   getSorobanTxData,
 } from "@/helpers/sorobanUtils";
 
 import { InputSideElement } from "@/components/InputSideElement";
 import { PositiveIntPicker } from "@/components/FormElements/PositiveIntPicker";
 
+// We add a bogus fee to simualte to fetch the min resource fee from the RPC
+const BOGUS_RESOURCE_FEE = "100";
```
Shaptic, line 19: could use the SDK constant `sdk.BASE_FEE` here

src/helpers/getScValsFromSpec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      if (inputFieldType === xdr.ScSpecType.scSpecTypeI128().name) {
+        return parsedIntValue.toI128();
+      }
+
+      if (inputFieldType === xdr.ScSpecType.scSpecTypeU128().name) {
+        return parsedIntValue.toU128();
+      }
+
+      return nativeToScVal(parsedIntValue, { type: inputFieldType });
```
Shaptic, line 50: I think you can use `return parsedIntValue.toScVal()` here?

jeesunikim: `parseIntValue.toScVal` returns `u64` even for the required `u128` type. I think it's because the saved value type is in `string` type so it returns the smallest possible XDR integer type when I use `parsedIntValue.toScVal()`, which is `u64` in this case even though the contract requires `u128` in the case of swap func in contract `CA6PUJLBYKZKUEKLZJMKBZLEKP2OTHANDEOWSFF44FTSYLKQPIICCJBE`

Shaptic: Ahhh yes that makes sense - you can pass an optional `{ type: "" }` but that'd be the same amount of code to compare the input field type.

PR #1285, opened 2025-03-21
Smart Contract: build verified badge with tooltip

quietbits:
image

image


jeesunikim:
should we have an `i` button?

jeesunikim (approved):

janewang:
I think it's fine! Thanks @quietbits

PR #1297, opened 2025-04-02
Smart Contracts: build info

quietbits:
> [!NOTE]  
> I couldn't find a smart contract on testnet or mainnet to test this with. I developed using an off-chain smart contract.

No validation screen

image

Validated screen (example using off-chain smart contract)

image

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/BuildInfo.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/BuildInfo.tsx
@@ -1,0 +1,103 @@
+import { Alert, Link, Text } from "@stellar/design-system";
+import { Box } from "@/components/layout/Box";
+import { getNetworkHeaders } from "@/helpers/getNetworkHeaders";
+import { renderWasmStatus } from "@/helpers/renderWasmStatus";
+import { useWasmGitHubAttestation } from "@/query/useWasmGitHubAttestation";
+import { useStore } from "@/store/useStore";
+
+export const BuildInfo = ({
+  wasmHash,
+  sourceCodeLink,
+  rpcUrl,
+  isActive,
+}: {
+  wasmHash: string;
+  sourceCodeLink: string;
+  rpcUrl: string;
+  isActive: boolean;
+}) => {
+  const { network } = useStore();
+
+  const {
+    data: wasmData,
+    error: wasmError,
+    isLoading: isWasmLoading,
+    isFetching: isWasmFetching,
+  } = useWasmGitHubAttestation({
+    wasmHash,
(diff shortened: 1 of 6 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/BuildInfo.tsx
janewang, line 1: This path probably will change once the SEP is merged. Just an FYI

src/query/useWasmGitHubAttestation.ts
```diff
@@ -0,0 +1,158 @@
...
+        const rpcServer = new StellarRpc.Server(rpcUrl, {
+          headers: isEmptyObject(headers) ? undefined : { ...headers },
+          allowHttp: new URL(rpcUrl).hostname === "localhost",
+        });
+
+        const wasm = await rpcServer.getContractWasmByHash(wasmHash, "hex");
+        const sourceRepo = await extractSourceRepo(wasm);
+
+        const att = await fetch(
+          `https://api.github.com/repos/${sourceRepo}/attestations/sha256:${wasmHash}`,
```
janewang, line 36: I think we should display this attestation path either as an URL so the user could click out to see, or display the attestation path as an URL and display the JSON too.

quietbits: That's a good idea. :+1: I added an attestation link that the user can open to view the JSON response.

jeesunikim:
valid-contract

- if the build is verified, should `build info` appear blend contract id `CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP` doesn't load build info.

janewang:
Screenshot 2025-04-08 at 10 19 09 AM

@quietbits For the `Build unverified` tag, I believe it should match the build info. In this case, it looks like this contract is `Build Verified`. Maybe we should pull this locally, instead of pulling what's provided by OL's API.

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
```diff
@@ -264,7 +274,7 @@ export const ContractInfo = ({
       },
     };
 
-    const badge = type ? item[type] : item.unverified;
+    const badge = hasWasmData ? item.verified : item.unverified;
```
jeesunikim, line 277: nice!

quietbits: That's correct. If no data is returned from that call, it means we couldn't fetch the repo data, or the validation checks failed.

PR #1301, opened 2025-04-07
[Invoke contract] Enable nested array of items

jeesunikim:
https://github.com/user-attachments/assets/692ebf39-c7fe-4769-b6d7-9c00f37d05fc

- create `setNestedValueWithArr` function to use a dynamic path to set a value for an array of nested objects. For example: `requests.1.request_type` path will target `requests[1].request_type`.
- added a recursion component for `array` schema type
- added `isRequired` check for validating type


src/helpers/dereferenceSchema.ts
```diff
@@ -1,112 +1,102 @@
...
+
+    return resolvedProps;
+  };
 
   return {
     name: methodName,
-    description: methodSchema.description,
-    properties,
+    description: methodSchemaObj.description ?? "",
+    properties: dereferenceSchemaProps(argsProperties),
```
jeesunikim, line 97: for array type, this will populate its nested properties

src/helpers/dereferenceSchema.ts
```diff
@@ -1,112 +1,102 @@
...
+      Object.entries(funcArgs.properties).forEach(([key, value]) => {
+        // `value` can be either JSONSchema7Definition or false
+        // parse the value if it's an object
+        if (typeof value === "object" && value !== null) {
+          if ("$ref" in value) {
+            resolvedProps[key] = resolveRef(value, fullSchema);
+          }
+          if ("items" in value) {
+            resolvedProps[key] = {
+              ...resolveRef(value.items, fullSchema),
```
jeesunikim, line 83: if it's an array, recursively evaluate and return

src/components/JsonSchema/JsonSchemaForm.tsx
```diff
@@ -0,0 +1,665 @@
...
+  );
+};
+
+const renderTitle = (name: string, description: string) => (
+  <>
+    <Text size="lg" as="h2">
+      {name}
+    </Text>
+    {description ? (
+      <Text size="sm" as="h3">
```
quietbits, line 539: Nit: We shouldn't use a heading element for text. This should be `div` or `p`.

src/components/JsonSchema/JsonSchemaForm.tsx
```diff
@@ -0,0 +1,665 @@
...
+          args: scVals,
+          resource_fee: BASE_FEE, // bogus resource fee for simulation purpose
+        },
+      },
+      networkPassphrase,
+    });
+
+    return builtXdr.toXDR();
+  } catch (e: any) {
+    console.error("e", e);
```
quietbits, line 572: Do we need to keep this `console.log`?

quietbits (approved):
Looks great! :tada:

PR #1302, opened 2025-04-08
Contract Explorer: show GitHub repo README.md in Source Code tab

quietbits:
- Using GitHub repo from StellarExpert's API
- Fetching `README.md` file from that repo
  - File name is case sensitive, but the GitHub official naming convention is "README.md", so we'll check that.

image

### DIFF
--- a/src/query/useGitHubReadmeText.ts
+++ b/src/query/useGitHubReadmeText.ts
@@ -1,0 +1,40 @@
+import { useQuery } from "@tanstack/react-query";
+
+/**
+ * Fetch readme file from a GitHub repo.
+ */
+export const useGitHubReadmeText = ({
+  repo,
+  commit,
+  isActive,
+}: {
+  repo: string;
+  commit: string;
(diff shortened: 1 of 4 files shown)


src/query/useGitHubReadmeText.ts
```diff
@@ -0,0 +1,40 @@
...
+        return null;
+      }
+
+      try {
+        const response = await fetch(
+          `https://raw.githubusercontent.com/${repo}/${commit}/README.md`,
+        );
+
+        if (response.status !== 200) {
+          return null;
```
jeesunikim, line 28: should we return a message here?

jeesunikim (approved):

src/query/useGitHubReadmeText.ts
quietbits: We handle this on the UI. If there is no valid response, we show a message "Couldn't find README.md in this repository".

PR #1321, opened 2025-04-10
Horizon and RPC network health check when website loads

quietbits:
- Add a `standalone` build mode to use with containers.
- Remove `force-dynamic` to test if CSP still works (this might no longer be necessary).
- Show a top banner if the Horizon and RPC networks are not ready when the website loads (needed for Quickstart).

image


fnando:
The message could be better when there's no RPC/horizon running yet.

CleanShot 2025-04-14 at 12 21 59@2x

fnando (approved):

PR #1339, opened 2025-04-17
Upgrade Next.js + React + ESLint + other minor upgrades

quietbits:
Major upgrades include:
- Next.js `v14` > `v15`
- React `v18` > `v19`
- ESLint

We'll handle these upgrades in follow-up PRs once we are certain everything works fine, simply to minimize potential issues.
- Zustand - major upgrade with potential breaking changes
- Stellar Design System - some breaking changes
- Stellar JS SDK - getting some warnings when running the app or tests (same for both `v13.1.0` and `v13.2.0`), so we'll hold off on updating it now.

<details>
<summary>Warnings example</summary>
<hr />

image
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "@next/eslint-plugin-next": "^15.3.1",
     "@playwright/test": "^1.52.0",
     "@stellar/tsconfig": "^1.0.2",
-    "@types/dompurify": "^3.2.0",
     "@types/json-schema": "^7.0.15",
     "@types/lodash": "^4.17.16",
     "@types/node": "^22.14.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1899,13 +1899,6 @@
   dependencies:
     "@types/node" "*"
 
-"@types/dompurify@^3.2.0":
-  version "3.2.0"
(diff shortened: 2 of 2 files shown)


src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
```diff
@@ -186,6 +186,7 @@ export default function Endpoints() {
                   const parsed = JSON.parse(item);
 
                   return parsed;
+                  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Comment (jeesunikim) [inline src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:189]: interesting. why is this necessary?

jeesunikim (approved):
thank you for making my code clear! just one Q

src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
Comment (quietbits) [reply src/app/(sidebar)/endpoints/[[...pages]]/page.tsx:189]: New ESLint rules are stricter about unused vars. In this case, we need to pass in a variable to `catch()`, but I didn't want to disable this rule for other cases (turning this rule off in the eslintsrc file). I thought that disabling it just for that case would be better.

PR #1361, opened 2025-04-25
reset wallet kit when network changes

jeesunikim:
**Summary:**
- reset previously saved wallet key whenever network changes
- use `<Icon/>` instead of `<IconButton/>` within `<button/>`
- remove `walletKit` from `account` store

### DIFF
--- a/src/components/WalletKit/ConnectedModal.tsx
+++ b/src/components/WalletKit/ConnectedModal.tsx
@@ -43,10 +43,19 @@ export const ConnectedModal = ({
         </Box>
       </Modal.Body>
       <Modal.Footer itemAlignment="stack">
-        <Button size="md" variant="tertiary">
-          <CopyText textToCopy={publicKey}>Copy address</CopyText>
-          <Icon.Copy01 />
-        </Button>
+        <CopyText textToCopy={publicKey} variant="headless">
+          <Button
+            isFullWidth
+            size="md"
+            variant="tertiary"
+            icon={<Icon.Copy01 />}
+            iconPosition="right"
+            onClick={(e) => e.preventDefault()}
+          >
+            Copy address
+          </Button>
+        </CopyText>
+
         <Button size="md" variant="tertiary" onClick={onDisconnect}>
           Disconnect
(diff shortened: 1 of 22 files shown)


src/components/WalletKit/ConnectedModal.tsx
```diff
@@ -46,21 +45,11 @@ export const ConnectedModal = ({
...
         <Button size="md" variant="tertiary">
           <CopyText textToCopy={publicKey}>Copy address</CopyText>
-          <IconButton
-            altText="Default"
-            customColor="var(--sds-clr-gray-09)"
-            customSize="14px"
-            icon={<Icon.Copy01 />}
-          />
+          <Icon.Copy01 />
         </Button>
```
quietbits, line 49: We can add an icon through the `Button` component's props: `icon` (pass the icon) and `iconPosition` (`left` or `right`, defaults to `right`). Same for the button below.

It might be better if we wrapped the `Button` component to make the whole `Button` area clickable.

```typescript
<CopyText textToCopy={tx.txHash}>
  <Button
    size="sm"
    variant="tertiary"
    icon={<Icon.Copy01 />}
    iconPosition="left"
    onClick={(e) => e.preventDefault()}
  >Copy address</Button>
</CopyText>
```

src/components/WalletKit/ConnectedModal.tsx
```diff
@@ -44,23 +43,22 @@ export const ConnectedModal = ({
...
-          />
-        </Button>
+            iconPosition="right"
+            onClick={(e) => e.preventDefault()}
+          >
+            Copy address
+          </Button>
+        </CopyText>
+
         <Button size="md" variant="tertiary" onClick={onDisconnect}>
```
quietbits, line 59: We can use icon props for this button as well.

```typescript
 <Button
  size="md"
  variant="tertiary"
  onClick={onDisconnect}
  icon={<Icon.LogOut01 />}
  iconPosition="right"
>Disconnect</Button>
```

quietbits (approved):

PR #1369, opened 2025-04-29
Replace book icon with info icon

quietbits:
- Replace book icon with info icon
- Upgrade to the latest SDS


leighmcculloch (approved):

src/app/(sidebar)/transaction/build/components/ClassicOperation.tsx
```diff
@@ -1,4 +1,4 @@
-import { ChangeEvent, Fragment, useState } from "react";
+import React, { ChangeEvent, Fragment, useState } from "react";
```
jeesunikim, line 1: do we need this?

quietbits: My IDE automatically added it because it prefers having it imported instead of using global. But we don't need it there, so I'll remove it.

PR #1371, opened 2025-04-29
Create Account: show banner on Mainnet

quietbits:
image


jeesunikim:
Should we also add this banner in sign transaction page?

jeesunikim (approved):
added a request to also add it in sign tx page

PR #1374, opened 2025-04-29
Network Selector: load custom network values correctly

quietbits:
[Example network settings](https://laboratory-pr1374.previews.kube001.services.stellar-ops.com/transaction/build?$=network$id=mainnet&label=testnet&horizonUrl=https:////evil.com&rpcUrl=https:////evil.com&passphrase=EVIL%20CORP;&transaction$build$classic$operations@$operation_type=&params@;;)

> [!NOTE] 
> The example RPC and Horizon URLs are not valid, so the banner will be displayed because of failed health checks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@ledgerhq/hw-app-str": "^7.2.0",
     "@ledgerhq/hw-transport-webusb": "^6.29.4",
     "@monaco-editor/react": "^4.7.0",
-    "@stellar/design-system": "^2.0.0-beta.17",
+    "@stellar/design-system": "^3.1.1",
     "@stellar/stellar-sdk": "^13.0.0",
     "@stellar/stellar-xdr-json": "^22.0.0-rc.1.1",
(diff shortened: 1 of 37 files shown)


jeesunikim:
banner

Where's the messaging coming from? The message is inaccurate.

janewang:
@jeesunikim How do I repro that message? I think we do a healthcheck to Horizon and RPC and show that banner but I'm not sure how to repro

jeesunikim:
@janewang @fnando I used the link that @quietbits provided in the description which had an invalid rpc url.

quietbits:
The message is meant for the Quickstart. We'll need to add more logic if we want to display a generic message that works for all invalid URLs.

fnando:
Maybe we can use something less assertive:

> RPC network couldn't be reached. If is still ingesting data, which can take several minutes. Until then, network requests will fail.

fnando (approved):
lgtm!

jeesunikim (approved):

PR #1382, opened 2025-04-30
Contract Explorer: add dev containers to open source code

quietbits:
image

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/SourceCode.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/SourceCode.tsx
@@ -1,8 +1,15 @@
-import { Alert, Code, Link, Loader, Text } from "@stellar/design-system";
+import {
+  Alert,
+  Code,
+  Link,
+  Loader,
+  Select,
+  Text,
+} from "@stellar/design-system";
 import { CodeEditor } from "@/components/CodeEditor";
 import { Box } from "@/components/layout/Box";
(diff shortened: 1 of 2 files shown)


jeesunikim (approved):

janewang (approved):

anataliocs:
I do see some errors in the logs but nothing that will crash the startup.  The most likely thing to cause errors at this point is that VSCode is installing a huge amount of extensions which may spike CPU to 100% and make the codespace difficult to use.

I think this will work just fine for reviewing code.

However, if users want to run tests or debug or get better context and tooling we will probally have to explore using a template to allow for more customization

anataliocs:
image
Confirmed codespace starts up successfully

src/app/(sidebar)/smart-contracts/contract-explorer/components/SourceCode.tsx
```diff
@@ -65,7 +72,58 @@ export const SourceCode = ({
...
+          fieldSize="sm"
+          onChange={(e) => {
+            handleOpenInContainer(e.target.value as ContainerResource);
+          }}
+          value=""
+        >
+          <option value="" disabled={true}>
+            Open in
+          </option>
+          <option value="devcontainer">Dev Container</option>
```
anataliocs, line 122: So just FYI devcontainers is the OSS technology and then Github Codespaces and Codeanywhere are the providers :)

janewang: We will add more info later :)

PR #1385, opened 2025-05-01
Network Selector: show new network settings modal from URL on Mainnet

quietbits:
On Mainnet, if the user loads custom Mainnet settings for the first time that don't match our defaults, we'll show a modal to accept the new settings.

[Example URL](https://laboratory-pr1385.previews.kube001.services.stellar-ops.com/?$=network$id=mainnet&label=Testnet&horizonUrl=https:////evil-horizon.org&rpcUrl=https:////evil-rpc.org&passphrase=EVIL;;)

image


kanwalpreetd:
@quietbits @janewang Do we want to add additional verbage like`You should only use Horizon and RPC URLs that you trust. Using unknown/untrusted Horizon and RPC URLs can make your connection insecure`. Not exactly this maybe, but something like this.
Looks good otherwise.

kanwalpreetd (approved):

jeesunikim (approved):

PR #1386, opened 2025-05-01
Enable an array of non object in invoke contract form

jeesunikim:
**Summary:**
- removed `setNestedValueWithArr` for lodash's `set` ([set doc](https://lodash.com/docs/#set))
- removed `getNestedValue` for lodash's `get` ([get doc](https://lodash.com/docs/#get))
- Add an additional `JsonSchemaFormRenderer` rendering for an array of primitive types (previously we only handled array of object types)
- Delete item from `formError` when a trash button icon is clicked
- update `dereferenceSchema` to keep the key `items`. previously we only kept the values of `items`

**Screenshot:**
- success for adding relayers using `CCQXWMZVM3KRTXTUPTN53YHL272QGKF32L7XEDNZ2S6OSUFK3NFBGG5M`
add_relayers

### DIFF
--- a/src/components/JsonSchema/JsonSchemaFormRenderer.tsx
+++ b/src/components/JsonSchema/JsonSchemaFormRenderer.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { Button, Card, Icon, Input, Text } from "@stellar/design-system";
 import type { JSONSchema7 } from "json-schema";
 import { parse } from "lossless-json";
+import { get, set } from "lodash";
 
 import { useStore } from "@/store/useStore";
 import { validate } from "@/validate";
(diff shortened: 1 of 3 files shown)


janewang:
Thank you @jeesunikim 

There are some contracts that don't have a contract spec. Like this one `CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA`. In the contract explorer, it shows that it doesn't have a Wasm. It's a SAC contract. I think we should handle this error slightly better, by indicating there's no Wasm. `There is no Wasm for this contract`, instead the error checking length of e.



Screenshot 2025-05-06 at 4 03 22 PM

quietbits (approved):

PR #1392, opened 2025-05-07
Saved Smart Contract IDs

quietbits:
image

image


jeesunikim (approved):

jeesunikim:
This doesn't have to be included in this PR, but I'd like the 'save' button to be clickable by keyboard `enter`. Wdyt? @quietbits 
click-enter

quietbits:
> This doesn't have to be included in this PR, but I'd like the 'save' button to be clickable by keyboard `enter`. Wdyt? @quietbits
> 
> click-enter

@jeesunikim , I like that suggestion. :+1:  We can make Enter and Escape control the buttons. This would affect other "save" modals, so it's better to make it a separate ticket.

jeesunikim:
@quietbits I created a ticket https://github.com/stellar/laboratory/issues/1401

PR #1395, opened 2025-05-07
[Contract Storage] cross out expired items

jeesunikim:
expired-item

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
@@ -157,7 +157,7 @@ export const ContractStorage = ({
             ),
           },
           { value: capitalizeString(vh.durability) },
-          { value: formatNumber(vh.ttl) },
+          { value: formatNumber(vh.ttl), isExpired: vh.expired },
           {
             value: formatEpochToDate(vh.updated, "short") || "-",
             isWrap: true,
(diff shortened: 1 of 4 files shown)


janewang:
@jeesunikim @sdfcharles Maybe it should say `Expired` in addition to crossing out the TTL for more clarity? Wdyt?

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
```diff
@@ -157,7 +188,25 @@ export const ContractStorage = ({
...
+                  {...(vh.expired ? { "data-style": "expired" } : {})}
+                >
+                  {formatNumber(vh.ttl)}
+                </span>
+                {vh.expired && vh.durability !== "temporary" ? (
+                  <Text size="sm" as="p">
+                    <Link onClick={() => handleRestore(vh.key, vh.durability)}>
+                      Restore
+                    </Link>
+                  </Text>
```
quietbits, line 205: We shouldn't need the wrapper `Text` element here. We can turn the above `span` into `div` or make `.TtlBox` box into a flexbox to stack the elements.

quietbits (approved):

janewang:
@jeesunikim I think what you have is good. We the restore button, we don't need to clarify it's expired. It should be quite obvious. If you can make the restore more like a button that would be awesome. Otherwise, looks good to me!

jeesunikim:
@janewang, @sdfcharles recommended a text link. I can make it bold?

janewang:
@jeesunikim What you had in the video looks fine :)

PR #1409, opened 2025-05-14
Vertical navigation

quietbits:
- Refactor horizontal nav to vertical nav.
- Simplify the layout because now we have a consistent nav + layout on all pages.
- I had to change the Horizon Endpoints URLs to include `horizon` (we already had `rpc` in the URLs of RPC Methods). Otherwise, it would have been very complicated to render the nested nav.
  - Added redirects to support old Horizon Endpoints URLs.
  - Added some top-level routes to use in nav rendering.
- Refactored the mobile nav source to use the same as the desktop.

image
image

image

### DIFF
--- a/src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
+++ b/src/app/(sidebar)/endpoints/[[...pages]]/page.tsx
@@ -959,7 +959,7 @@ export default function Endpoints() {
     return <EndpointsLandingPage />;
   }
 
-  if (pathname === Routes.ENDPOINTS_SAVED) {
+  if (pathname === Routes.SAVED_ENDPOINTS) {
     return <SavedEndpointsPage />;
   }
 
--- a/src/app/(sidebar)/transaction/saved/page.tsx
+++ b/src/app/(sidebar)/transaction/saved/page.tsx
@@ -253,7 +253,7 @@ export default function SavedTransactions() {
         title="Looking for your saved requests?"
         placement="inline"
       >
-        <NextLink href={`${Routes.ENDPOINTS_SAVED}`} sds-variant="primary">
+        <NextLink href={`${Routes.SAVED_ENDPOINTS}`} sds-variant="primary">
           See saved requests
         </NextLink>
       </Alert>


package.json
```diff
@@ -7,7 +7,8 @@
     "node": ">=22.0.0"
   },
   "scripts": {
-    "dev": "NEXT_PUBLIC_COMMIT_HASH=$(git rev-parse --short HEAD) NEXT_PUBLIC_ENABLE_EXPLORER=true next dev",
+    "dev": "NEXT_PUBLIC_COMMIT_HASH=$(git rev-parse --short HEAD) next dev",
+    "dev-qs": "NEXT_PUBLIC_COMMIT_HASH=$(git rev-parse --short HEAD) NEXT_PUBLIC_ENABLE_EXPLORER=true next dev",
```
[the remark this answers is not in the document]
fnando: Oh, that's a great call, actually. What if we display everything in dev mode, but not in production mode?

Then we can do a check like `process.env.NEXT_PUBLIC_ENABLE_EXPLORER === "true" || whatever_check_for_development_is`

jeesunikim:
this happens in production site too but there's a brief moment of `set a testnet rpc url` warning message.

screenshot

https://github.com/user-attachments/assets/beee04fe-5451-4bfe-8809-f6531b565948

src/app/error.tsx
```diff
@@ -14,47 +13,45 @@ export default function Error({
...
 
-            <Text size="sm" as="p">
-              Uh-oh, we didn’t handle this error. We would appreciate it if you
-              opened an issue on GitHub, providing as many details as possible
-              to help us fix this bug.
-            </Text>
-          </Box>
+          <Text size="sm" as="p">
+            Uh-oh, we didn’t handle this error. We would appreciate it if you
+            opened an issue on GitHub, providing as many details as possible to
```
jeesunikim, line 25: it'd be nice to have a link to an issue!

quietbits: We can add a link to the issue, but looking at Sentry logs might be more helpful (once we have it in `main`).

PR #1415, opened 2025-05-20
Refactor ScValPrettyJson to add types

quietbits:


src/components/ScValPrettyJson/index.tsx
```diff
@@ -79,269 +72,322 @@ export const ScValPrettyJson = ({
...
+          return value;
+        }
+
+        return (
+          <Value key={`${type}-${uuidv4()}`} type={type} hasComma={hasComma}>
+            {value}
+          </Value>
+        );
+      case "bool":
+      case "string":
```
jeesunikim, line 158: can add a case for `dataurl (or bytes)` as well ([doc](https://developers.stellar.org/docs/learn/encyclopedia/contract-development/types/built-in-types#bytes-strings-bytes-bytesn-string))?

this is what stellar.expert does:
Screenshot 2025-05-20 at 11 44 41 AM

quietbits: Looks like they are strings, so they should be handled by the default case. I don't think we need to list all the types, unless we need to handle them differently.

janewang: Sorry, accidentally clicked resolved when I wanted to upload an image. 

I think there's benefit in distinguish between each string types. Today we show differet types with different colors.
Screenshot 2025-05-22 at 8 39 04 PM

quietbits:
Contrast for large numbers is not good in the code block. @sdfcharles , could you please do your magic here? :pray: 

image
image

sdfcharles:
color updates here:

top: order in which the legend should show, also update to `badge` component instead of current dots
bottom: color values for each type

CleanShot 2025-06-10 at 16 00 02@2x

quietbits:
> Screenshot 2025-06-11 at 5 16 22 PM
> Do people understand these two pills at the end. They are not officially primitive types without the `u` and the `i`. [Here](https://developers.stellar.org/docs/learn/encyclopedia/contract-development/types/built-in-types) are the types.

@janewang, we removed the color legends as they are no longer needed now that we show the types. There was no nice way to clearly show the integers. cc: @sdfcharles

jeesunikim:
@quietbits I am running into an unhandled error with the contract id `CBHQGTSBJWA54K67RSG3JPXSZY5IXIZ4FSLJM4PQ33FA3FYCU5YZV7MZ` on TESTNET when I click contract storage. This doesn't happen on production site.

jeesunikim (approved):

janewang (approved):

PR #1417, opened 2025-05-21
Restore Trezor

quietbits:


quietbits:
I just successfully submitted a transaction using a Trezor device.

PR #1444, opened 2025-06-11
Create Account: add recovery phrase

quietbits:


src/app/(sidebar)/account/create/page.tsx
```diff
@@ -70,15 +71,23 @@ export default function CreateAccount() {
   const generateKeypair = () => {
     resetStates();
 
-    const keypair = Keypair.random();
+    // Generate 24-word mnemonic
+    const mnemonic = StellarHDWallet.generateMnemonic({
+      entropyBits: 256,
```
jeesunikim, line 76: I like that you explicitly mentioned those even though they are the default values!

jeesunikim (approved):

PR #1447, opened 2025-06-18
Persist wallet in localstorage

jeesunikim:
**Summary:**
- create a `localStorage` for wallet connect key to persist `walletKit.id`
- Persist the state as long as user doesn't disconnect
- [this](https://github.com/blend-capital/blend-ui/blob/main/src/contexts/wallet.tsx) is how blend implemented

https://github.com/user-attachments/assets/8a4b2b5c-1ac3-4d00-ba89-8c2e604e0ead


src/components/WalletKit/ConnectWallet.tsx
```diff
@@ -12,59 +12,108 @@ import { ConnectedModal } from "@/components/WalletKit/ConnectedModal";
...
+      !connected &&
+      savedWallet?.id !== undefined &&
+      savedWallet?.id !== "false" &&
+      savedWallet?.id !== "wallet_connect"
+    ) {
+      // timeout ensures chrome has the ability to load extensions
+      setTimeout(() => {
+        walletKitInstance.walletKit?.setWallet(savedWallet?.id);
+        handleSetWalletAddress();
+      }, 750);
```
quietbits: We should clear the timeout to make sure there won't be any memory leaks.

src/components/WalletKit/ConnectWallet.tsx
```diff
@@ -12,59 +12,108 @@ import { ConnectedModal } from "@/components/WalletKit/ConnectedModal";
...
+    setConnected(false);
+    localStorageSavedWallet.remove();
   };
 
+  useEffect(() => {
+    if (
+      !connected &&
+      savedWallet?.id !== undefined &&
+      savedWallet?.id !== "false" &&
+      savedWallet?.id !== "wallet_connect"
```
quietbits: Could we write it like this? `savedWallet?.id` can only have one value, so it might be clearer.

```suggestion
      ![undefined, "false", "wallet_connect"].includes(savedWallet?.id)
```

src/components/WalletKit/ConnectWallet.tsx
```diff
@@ -12,59 +12,108 @@ import { ConnectedModal } from "@/components/WalletKit/ConnectedModal";
...
+  async function handleSetWalletAddress(): Promise<boolean> {
+    try {
+      const addressResult = await walletKitInstance.walletKit?.getAddress();
+
+      if (!addressResult?.address) {
+        return false;
+      }
+      const publicKey = addressResult.address;
+
+      if (publicKey === "" || publicKey == undefined) {
```
quietbits:
```suggestion
      if (!publicKey) {
```

src/components/WalletKit/ConnectWallet.tsx
```diff
@@ -12,59 +12,108 @@ import { ConnectedModal } from "@/components/WalletKit/ConnectedModal";
...
+
+      updateWalletKit({
+        publicKey,
+        walletType: savedWallet?.id,
+      });
+      setConnected(true);
+
+      return true;
+    } catch (e: any) {
+      console.error("Unable to load wallet information: ", e);
```
quietbits: Do we need this here?

quietbits (commented):
Should we clear the wallet when network changes?

src/components/WalletKit/ConnectWallet.tsx
jeesunikim: thank you! imma use `delayedAction`!

quietbits:
> > Should we clear the wallet when network changes?
> 
> I implemented that part when I added `connect wallet` initially 👀. `Connect Wallet` to appear whenever the network changes. It resets on my end or am I thinking something else, @quietbits ?

@jeesunikim , it works fine when the network changes in the same session, but it picks it up again on page reload. We might need to add an extra check there for the network when restoring.

https://github.com/user-attachments/assets/2fde84f3-0686-4438-95fa-725c89901c34

src/components/WalletKit/ConnectWallet.tsx
```diff
@@ -7,64 +7,119 @@ import { useAccountInfo } from "@/query/useAccountInfo";
 
 import { shortenStellarAddress } from "@/helpers/shortenStellarAddress";
 import { getNetworkHeaders } from "@/helpers/getNetworkHeaders";
+// import { useDelayedAction } from "@/hooks/useDelayedAction";
 
 import { ConnectedModal } from "@/components/WalletKit/ConnectedModal";
 import { WalletKitContext } from "@/components/WalletKit/WalletKitContextProvider";
 
 import { trackEvent, TrackingEvent } from "@/metrics/tracking";
+import { localStorageSavedWallet } from "@/helpers/localStorageSavedWallet";
```
quietbits: We can remove the commented out import and group helpers imports together.

PR #1458, opened 2025-06-30
[Invoke Contract] unit tests

jeesunikim:
**Summary:**
- Added a jest library for unit tests
- Moved playwright e2e tests to its own folder `/e2e`
- Updated the testing commands from yarn run test to:
-- `yarn run test:e2e` (playwright)
-- `yarn run test:unit` (jest)
- Fixed tests that were breaking on e2e by adding `data-testid`
- Added `Enum with Integer Variant` that was missing
- Added a helper function that detects the type of bytes encoding. previously, it had assumed everything as `base64`


package.json
```diff
@@ -12,11 +12,12 @@
     "start": "next start",
     "lint": "next lint",
     "lint:ts": "tsc --noEmit",
-    "test": "playwright test",
+    "test:unit": "jest",
+    "test:e2e": "playwright test",
```
quietbits, line 16: We could also create a `test` script that includes them both.
```suggestion
    "test:unit": "jest",
    "test:e2e": "playwright test",
    "test": "yarn test:unit && yarn test:e2e",
```

src/app/(sidebar)/account/muxed-create/page.tsx
```diff
@@ -89,6 +89,7 @@ export default function CreateMuxedAccount() {
 
           <PubKeyPicker
             id="muxed-public-key"
+            data-testid="muxed-public-key"
```
quietbits: Why are these tests failing now? Ideally, we should use `data-testid` only where we can't use UI elements (target by text, etc).

src/components/SmartContractJsonSchema/renderOneOf.tsx
```diff
@@ -103,14 +106,28 @@ export const renderOneOf = ({
...
         {schema.oneOf.map((oneOf, index) => {
           if (typeof oneOf === "boolean") return null;
 
+          if (oneOf.enum && oneOf.enum.length > 0) {
+            return (
+              <option value={oneOf.enum[0]?.toString()} key={`${oneOf.title}`}>
+                {oneOf.title} = {oneOf.enum[0]?.toString()}
+              </option>
+            );
+          }
```
quietbits: We can create a variable for the value we use in both places.

```suggestion
          if (oneOf.enum && oneOf.enum.length > 0) {
            const val = oneOf.enum[0]?.toString();
          
            return (
              <option value={val} key={`${oneOf.title}`}>
                {oneOf.title} = {val}
              </option>
            );
          }
```

src/helpers/sorobanUtils.ts
```diff
@@ -286,7 +286,12 @@ const convertEnumToScVal = (obj: Record<string, any>, scVals?: xdr.ScVal[]) => {
     return tupleScValsVec;
   }
 
-  // ENUM CASE
+  // ENUM INTEGER VARIANT CASE
```
quietbits: Do we need all caps in this comment and below? :wink:

quietbits (approved):
Thanks for the thorough tests! :raised_hands:

src/helpers/sorobanUtils.ts
jeesunikim: haha I'll update it to stop screaming

src/app/(sidebar)/account/muxed-create/page.tsx
jeesunikim: Maybe adding an `await` was enough. I no longer see an issue.

PR #1472, opened 2025-07-02
Improve transaction parsing and handle cache better.

fnando:


src/constants/settings.ts
```diff
@@ -16,6 +16,8 @@ export const LOCAL_STORAGE_SAVED_CONTRACTS = "stellar_lab_saved_contract_ids";
 export const LOCAL_STORAGE_SAVED_THEME = "stellarTheme:Laboratory";
 export const LOCAL_STORAGE_SAVED_EXPLORER_TRANSACTIONS =
   "stellar_lab_saved_explorer_transactions";
+export const LOCAL_STORAGE_EXPLORER_TRANSACTIONS_RPC_REF =
+  "stellar_lab_saved_explorer_transactions_rpc_ref";
```
quietbits, line 20: We have a `LOCAL_STORAGE_SAVED_NETWORKS_PREVIOUS` that saves the last used network settings per network. Not sure if we could use that here.

quietbits (approved):

src/constants/settings.ts
fnando: I don't think we can, because we'd be checking `network.rpcUrl` against itself, right?

PR #1474, opened 2025-07-03
add color pill for types for invoke contract

jeesunikim:
color-pills-strukt

color-pills-multi

color-pills-u32

### DIFF
--- a/src/components/ColorPillType/index.tsx
+++ b/src/components/ColorPillType/index.tsx
@@ -1,0 +1,16 @@
+import "./styles.scss";
+
+export const ColorPillType = ({ type }: { type: string }) => {
+  const lowercaseType = type?.toLowerCase();
+
+  return (
+    <div className="ColorPillType">
+      <div
+        className="ColorPillType__value"
+        {...(lowercaseType ? { "data-type": lowercaseType } : {})}
+      >
+        {lowercaseType}
+      </div>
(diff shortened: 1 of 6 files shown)


src/components/ColorPillType/index.tsx
```diff
@@ -0,0 +1,16 @@
+import "./styles.scss";
+
+export const ColorPillType = ({ type }: { type: string }) => {
```
quietbits: Nit: Would it be clearer to name it `ColorTypePill`? The output of this component is a pill, not a type. :thinking:

src/components/ColorPillType/styles.scss
```diff
@@ -0,0 +1,83 @@
...
+  border: 1px solid var(--ColorPill-color-border);
+
+  background-color: var(--ColorPill-color-background);
+  color: var(--ColorPill-color-text);
+
+  // Colors
+  // Symbol
+  [data-type="symbol"] {
+    color: var(--sds-clr-navy-10);
+    border-color: var(--sds-clr-navy-10);
```
quietbits: Since we have CSS variables for colors, we should probably use them to style different types.

```suggestion
    --ColorPill-color-text: var(--sds-clr-navy-10);
    --ColorPill-color-border: var(--sds-clr-navy-10);
```

PR #1475, opened 2025-07-07
Simulate Transaction: add Submit button

quietbits:
image


quietbits:
@janewang, I added a Submit button that takes the user to the Submit page, where the submission happens. It's consistent with how we handle it on other pages (for example, Sign Transaction). Adding a submission on the same page would add some complexity.

jeesunikim (approved):

PR #1480, opened 2025-07-07
Transaction Dashboard: tabs placeholder

quietbits:
- All tabs (only State Change and Signatures enabled for now)
- Refactored `getTxData()` method to use in multiple places
- State Change placeholder component
- Showing only for Soroban transactions

image


jeesunikim:
can we have a higher contrast in colors between active and inactive items? Only `State Change` and `Signatures` are the active items, but it's hard to distinguish that with other menu items.

dark-mode
light-mode

src/app/(sidebar)/transaction-dashboard/components/TransactionInfo.tsx
```diff
@@ -125,7 +108,7 @@ export const TransactionInfo = ({
     }
 
     const baseProps = {
-      isSorobanTx: IS_SOROBAN_TX,
+      isSorobanTx: isSorobanTx,
```
jeesunikim: it can just be `isSorobanTx`

jeesunikim (approved):

PR #1482, opened 2025-07-08
fix the bug with fetching sequence number

jeesunikim:
**Summary:**
- always fetch the latest sequence number (bug that was reported by fifo)
- add a timeout for a case when user x out of extension
- remove unused `border-color`

### DIFF
--- a/src/app/(sidebar)/transaction-dashboard/components/StateChange.tsx
+++ b/src/app/(sidebar)/transaction-dashboard/components/StateChange.tsx
@@ -1,0 +1,9 @@
+import { Text } from "@stellar/design-system";
+
+export const StateChange = () => {
+  return (
+    <Text as="div" size="sm" weight="regular">
+      Coming soon
+    </Text>
+  );
+};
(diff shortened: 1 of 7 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
```diff
@@ -144,14 +144,26 @@ export const InvokeContractForm = ({
...
       try {
-        const result = await walletKitInstance.walletKit.signTransaction(
+        // Add timeout to prevent endless loading
+        // when user exits out of the extension
+        const timeoutPromise = new Promise<never>((_, reject) => {
+          setTimeout(() => {
+            reject(
+              new Error("Transaction signing timed out. Please try again."),
+            );
+          }, 20000);
```
quietbits: We need to make sure we clear the timeout. We can set a variable outside of the `signTx()` method (`let timeout = null`, for example). And then clear it on line 139 whenever `signTx()` method fires. And then:

```suggestion
          timeout = setTimeout(() => {
            reject(
              new Error("Transaction signing timed out. Please try again."),
            );
            clearTimeout(timeout)
          }, 20000);
```

jeesunikim: I'll set the timeout variable within `signTx` so that in case of being clicked multiple times, it won't use the same `timeout` variable

quietbits (approved):

PR #1484, opened 2025-07-09
[Invoke Contract] Enable simulate and submit in one step

jeesunikim:
**Summary:**

Previously on **Invoke Contract**, `submit` button used the txn from simulated response in both `testnet` and `mainnet`. Let's say user triggers simulating function A then simulating function B then submitting function A. When they submit function B, it will throw an error because its txn is using the old sequence number from the old simulation. I renamed `submit` to `simulate and submit` so that when user submits a tx, it'll always use the latest sequence number.

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
@@ -103,7 +103,6 @@ export const InvokeContractForm = ({
 
   const {
     mutateAsync: prepareTx,
-    data: prepareTxData,
     isPending: isPrepareTxPending,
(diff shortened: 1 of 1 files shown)


quietbits (commented):
Updates look good to me. A couple of questions:
1. Does Horizon support simulate?
2. When doing simulate + submit, should we change the button label to reflect that? I'm not a fan of hidden actions, so I'd like to see a clear indication on the UI that I will be doing both.

jeesunikim:
@quietbits :

- Does Horizon support simulate? **No**
- When doing simulate + submit, should we change the button label to reflect that? I'm not a fan of hidden actions, so I'd like to see a clear indication on the UI that I will be doing both. **Updated**

quietbits (approved):

PR #1498, opened 2025-07-11
P23

jeesunikim:


jeesunikim:
I need to debug why `getTransaction` on `useSubmitRpc` using the tx from `sendTransaction` isn't working. I'm using the custom type contract `CBHQGTSBJWA54K67RSG3JPXSZY5IXIZ4FSLJM4PQ33FA3FYCU5YZV7MZ`. debug-invoke-contract. This works fine on production so it has to do with P23 changes.

The call stack is:
1. hitting `getTransaction`'s `Object.assign(foundInfo, parseTransactionInfo(raw));` ([code](https://github.com/stellar/js-stellar-sdk/blob/886abae593f1542a48b48d1c9c2fc3241e0dc382/src/rpc/server.ts#L559-L561))
2. it then throws an error at [the line 53](https://github.com/stellar/js-stellar-sdk/compare/v14.0.0-rc.1...protocol-23#diff-786dbafe96164e75829b0f37da14ceea9e213c664cec409d9a0d4558e199be92R53-R56) `raw.events?.transactionEventsXdr.map` with the message `TypeError: Cannot read properties of undefined (reading 'map')`.

cc: @Shaptic but I'll also take a look tomorrrow

src/components/formComponentTemplateEndpoints.tsx
```diff
@@ -965,6 +963,38 @@ export const formComponentTemplateEndpoints = (
...
+            </option>
+            <option id="enforce" value="enforce">
+              enforce
+            </option>
+            <option id="record" value="record">
+              record
+            </option>
+            <option id="record_allow_nonroot" value="record_allow_nonroot">
+              record_allow_nonroot
+            </option>
```
quietbits: We could simplify this:

```suggestion
            {
              ["none", "enforce", "record", "record_allow_nonroot"].map((i) => (
                <option key={i} id={i} value={i === "none" ? "" : i}>
                  {i}
                </option>
              ));
            }
```

jeesunikim: I'll keep it as it is since other components use a similar structure

jeesunikim: actually other components are different because its text value has spacing in between. applied. thanks

jeesunikim:
@Shaptic "https://soroban-testnet.stellar.org"?

PR #1510, opened 2025-07-18
Contract Explorer: Support Stellar Asset Contracts

quietbits:
SAC ID on Testnet: `CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC`

lab-sac

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractSpec.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractSpec.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useState } from "react";
 import { Button, Icon } from "@stellar/design-system";
+import { parse, stringify } from "lossless-json";
 
 import { CodeEditor, SupportedLanguage } from "@/components/CodeEditor";
 import { Box } from "@/components/layout/Box";
@@ -9,12 +10,18 @@ import { PoweredByStellarExpert } from "@/components/PoweredByStellarExpert";
 import { useWasmBinaryFromRpc } from "@/query/useWasmBinaryFromRpc";
 import { useGitHubFile } from "@/query/useGitHubFile";
 
+import * as StellarXdr from "@/helpers/StellarXdr";
 import { downloadFile } from "@/helpers/downloadFile";
 import { renderContractSpecViewStatus } from "@/helpers/renderContractSpecViewStatus";
 import { getWasmContractData } from "@/helpers/getWasmContractData";
 import { STELLAR_ASSET_CONTRACT } from "@/constants/stellarAssetContractData";
+import { useIsXdrInit } from "@/hooks/useIsXdrInit";
 
-import { ContractSections, ContractSectionName } from "@/types/types";
+import {
+  ContractSections,
+  ContractSectionName,
+  AnyObject,
+} from "@/types/types";
 
 export const ContractSpec = ({
   rpcUrl,
@@ -40,6 +47,8 @@ export const ContractSpec = ({
   const [contractSections, setContractSections] =
     useState<ContractSections | null>();
 
+  const isXdrInit = useIsXdrInit();
+
   const {
     data: wasmBinary,
     error: wasmBinaryError,
@@ -59,10 +68,7 @@ export const ContractSpec = ({
   } = useGitHubFile({
     repo: STELLAR_ASSET_CONTRACT.contractSpecRepo,
     path: STELLAR_ASSET_CONTRACT.contractSpecPath,
-    file:
-      selectedFormat.sac === "xdr"
-        ? `${STELLAR_ASSET_CONTRACT.contractSpecFileName}.xdr`
-        : `${STELLAR_ASSET_CONTRACT.contractSpecFileName}.json`,
+    file: `${STELLAR_ASSET_CONTRACT.contractSpecFileName}.json`,
     isActive: Boolean(isActive && isSacType),
   });
 
@@ -94,17 +100,36 @@ export const ContractSpec = ({
     return viewStatus;
   }
 
+  const formatSacData = (
+    dataString: string | null | undefined,
+    type: "json" | "xdr",
+  ) => {
+    if (!dataString) {
+      return "";
+    }
+
+    const jsonData = parse(dataString) as AnyObject[];
+
+    if (type === "json") {
+      return `${jsonData.map((d) => stringify(d, null, 2)).join(",\n\n")}`;
+    }
+
+    return isXdrInit
+      ? `${jsonData.map((d) => StellarXdr.encode("ScSpecEntry", stringify(d) || "")).join("\n\n")}`
+      : "";
+  };
+
   const formatValue = (sectionName: ContractSectionName) => {
-    // Adding the first line as a comment with a section name. For example:
-    // // contractspecv0
     switch (selectedFormat[sectionName]) {
       case "json":
         return isSacType
-          ? `${sacData}`
-          : `// ${sectionName} \n\n${contractSections?.[sectionName].json?.join(",\n\n")}`;
+          ? formatSacData(sacData, "json")
+          : // Adding the first line as a comment with a section name. For example:
+            // // contractspecv0
+            `// ${sectionName} \n\n${contractSections?.[sectionName].json?.join(",\n\n")}`;
       case "xdr":
         return isSacType
-          ? `${sacData}`
+          ? formatSacData(sacData, "xdr")
           : `// ${sectionName} \n\n${contractSections?.[sectionName].xdr?.join("\n\n")}`;
       case "text":
       default:


src/constants/stellarAssetContractData.ts
```diff
@@ -0,0 +1,7 @@
+export const STELLAR_ASSET_CONTRACT = {
+  sourceRepo: "stellar/rs-soroban-env",
+  sourceCommit: "339d5957d2a258ca888d2f7eb50e92e425555256",
+  contractSpecRepo: "stellar/stellar-asset-contract-spec",
+  contractSpecPath: "refs/heads/main",
+  contractSpecFileName: "stellar-asset-spec",
```
quietbits, line 6: The XDR returns content that looks like markdown instead of XDR. Is that expected ([source](https://github.com/stellar/stellar-asset-contract-spec/blob/main/stellar-asset-spec.xdr))?



image

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
```diff
@@ -42,11 +43,13 @@ export const ContractInfo = ({
   wasmData,
   network,
   isLoading,
+  isSacType,
 }: {
   infoData: ContractInfoApiResponse | undefined;
   wasmData: WasmData | null | undefined;
   network: Network | EmptyObj;
   isLoading: boolean;
+  isSacType?: boolean;
```
quietbits, line 52: What should we display for SAC in the tabs 'Build Info', 'Version History', and 'Bindings'?

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
```diff
@@ -276,9 +299,21 @@ export const ContractInfo = ({
           </>
         ),
       },
+      builtIn: {
+        badge: (
+          <Badge variant="success" icon={<Icon.CheckCircle />}>
+            Built-in Contract
+          </Badge>
+        ),
+        message: <>TODO: needs description</>,
```
quietbits: What copy should we show here?
image

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
```diff
@@ -124,6 +139,10 @@ export default function ContractExplorer() {
     !contractIdInput ||
     Boolean(contractIdInputError);
 
+  const isSacType = contractType
+    ? contractType === "contractExecutableStellarAsset"
+    : undefined;
+
```
quietbits, line 145: For Invoke Contract, currently, we're getting functions from the Wasm file. How should we get them for SAC? I tried to parse [the spec JSON from the repo](https://github.com/stellar/stellar-asset-contract-spec/blob/main/stellar-asset-spec.json), but it didn't like some characters. :pray:

src/constants/stellarAssetContractData.ts
Shaptic: Those are probably the rust comments from the contract which can be markdown-formatted

src/constants/stellarAssetContractData.ts
```diff
@@ -0,0 +1,7 @@
+export const STELLAR_ASSET_CONTRACT = {
+  sourceRepo: "stellar/rs-soroban-env",
+  sourceCommit: "339d5957d2a258ca888d2f7eb50e92e425555256",
```
Shaptic: I recommend using a tag here

src/query/useGetContractTypeFromRpcById.ts
```diff
@@ -0,0 +1,67 @@
...
+  contractId,
+  rpcUrl,
+  headers = {},
+}: {
+  contractId: string;
+  rpcUrl: string;
+  headers?: NetworkHeaders;
+}) => {
+  const query = useQuery<
+    "contractExecutableWasm" | "contractExecutableStellarAsset" | null
```
Shaptic: I wouldn't hardcode these strings if you can avoid it,

```suggestion
    ContractExecutableType.contractExecutableWasm().name | 
    ContractExecutableType.contractExecutableStellarAsset().name | 
    null
```

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
Shaptic: wdym by "didn't like some characters"? :/

src/constants/stellarAssetContractData.ts
quietbits: We use commit for non-SAC contracts. Why is tag better here? :pray:

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
janewang: There is no Build Info (because it's built in). There is no History. I'm not sure about 'Bindings'.

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
quietbits: image

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
janewang: The Stellar Asset Contract (SAC) is a built-in smart contract that provides a standardized, programmatic interface to Stellar assets on the network. It implements the [SEP‑41 Token Interface](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0041.md).

janewang: @quietbits Added the description

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
quietbits: Should we disable those tabs or add some text?

janewang: I think we can disable, wdyt?

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
Shaptic: :confused: that seems like a flaw of the spec itself. It looks like there should be a comma there to split the `}{`... Can we regenerate this with valid JSON, @leighmcculloch?

src/constants/stellarAssetContractData.ts
Shaptic: Because the built-in contracts don't change across major versions (I'd hope!), so the tag would be more self-describing (e.g. you'd see which last major version it changed in).

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
quietbits: Disabling makes sense to me. Thank you!

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
Shaptic: I'd also add that it's an implementation of [CAP-46-06](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-06.md), e.g.

> It implements the [SEP‑41 Token Interface](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0041.md) via [CAP-46-06](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-06.md).

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
Shaptic: There def is history (e.g. in P23 the event formats changed), but it'd likely be a bigger lift to describe this (since the changes are outlined in CAPs vs. commits).

quietbits: @Shaptic, currently, we get history from StellarExpert's API ([request](https://api.stellar.expert/explorer/testnet/contract/CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC/version)), which doesn't return anything. Should we leave this as is, because it might have something in the future?

janewang: When we talk about history, we are talking about Wasm version history. Like this:

Screenshot 2025-07-18 at 3 54 46 PM

@Shaptic Do you mean there is another Wasm? I thought the contract is built in?

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
janewang: When we buidl, we find 🐞

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
Shaptic: Yeah that's what I meant about a bigger lift; it'd likely need to be hard-coded (or, possibly, use the GitHub API to fetch changes to [this folder](https://github.com/stellar/rs-soroban-env/tree/main/soroban-env-host/src/builtin_contracts/stellar_asset_contract)). I'd disable it for now until there's a better strategy; this may not even be that important to display imo.

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
Shaptic: @quietbits you could download the XDR then xdr2json it in the meantime? :shrug: though I'm not even sure that spec is up to date, given its age+archived status

quietbits: @Shaptic, is there another way we can get the XDR? The one in the repo gives us this: https://github.com/stellar/laboratory/pull/1510#discussion_r2216656850 :woman_shrugging:

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
Shaptic: Ah if you just mean wasm history, as opposed to just "history of this contract changing" (which can apply to built-ins just the same), then yeah you can just ignore this :+1:

janewang: Yeah wasm history, I assume we keep the contract the same for stability reasons

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
Shaptic: But that XDR seems valid - string comments in the spec (same way if you `grep`'d on a raw binary you'd see things like plaintext error strings in it). Idk why you're rendering it as text, though, as opposed to base64 or hex

quietbits: @Shaptic, this is what we get for non-SAC spec. Do we need to encode the XDR from the repo? When I see `.xdr` file, I assume it's an encoded XDR string. :sweat_smile: 

https://github.com/user-attachments/assets/7f09b8d0-1a1b-430c-812c-c4c0dd29cd53

Shaptic: Ah no the .xdr file is literally the raw binary XDR

Shaptic: Whereas the RPC will give you the already-encoded XDR since it’s JSON

src/query/useGetContractTypeFromRpcById.ts
```diff
@@ -0,0 +1,74 @@
...
+          ?.executable();
+
+        if (!executable) {
+          throw "Could not get executable from contract data.";
+        }
+
+        const contractType = executable?.switch()?.name;
+
+        if (
+          ["contractExecutableWasm", "contractExecutableStellarAsset"].includes(
```
Shaptic: Same idea here: re strings btw

src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
leighmcculloch: > 😕 that seems like a flaw of the spec itself. It looks like there should be a comma there to split the `}{`... Can we regenerate this with valid JSON, @leighmcculloch?

It's a stream of JSON objects. We can change the repo to generate a newline delimited stream, or an array instead. @quietbits Feel free to resurrect the repo and open a PR making it output an array.

janewang (approved):
Check out in the UI, not code. 

It looks great, thank you!

PR #1516, opened 2025-07-25
[Invoke Contract] Enable Stellar Asset Contract

jeesunikim:
SAC ID on Testnet: CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC

**Summary:**
- created `useSacXdrData` hook so I can re-use it for `Build Tx -> Invoke Contract` later ([#1517](https://github.com/stellar/laboratory/issues/1517))
- simplified the `InvokeContractForm` code by removing the params `infoData`, `network`; passing `contractSpec` so that I don't have to re-call it within that form

Here's [the transfer contract tx](https://stellar.expert/explorer/testnet/tx/3823ddbd3f661d09a03bd1a40fdd6d336bf14f8d76ac93104d245749036aa5b0) that I did while testing this feature

Screenshot 2025-07-25 at 12 38 09 PM

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
@@ -3,6 +3,7 @@
 import { useEffect, useState } from "react";
 import { Alert, Button, Icon, Input } from "@stellar/design-system";
 import { useQueryClient } from "@tanstack/react-query";
+import { contract } from "@stellar/stellar-sdk";
 
 import { useStore } from "@/store/useStore";
 import { useSEContractInfo } from "@/query/external/useSEContractInfo";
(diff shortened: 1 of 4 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
```diff
@@ -27,6 +28,7 @@ import { trackEvent, TrackingEvent } from "@/metrics/tracking";
 
 import { ContractInfo } from "./components/ContractInfo";
 import { InvokeContract } from "./components/InvokeContract";
+import { useSacXdrData } from "@/hooks/useSacXdrData";
```
quietbits: Nit: this import should go before relative imports

src/hooks/useSacXdrData.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (!sacData) {
+    return {
+      SacXdrData: "",
+      sacDataError: null,
+      isSacDataFetching: false,
+      isSacDataLoading: false,
+    };
+  }
+
+  const SacXdrData = (dataString: string | null | undefined): string[] => {
```
quietbits: If it's just a formatter and not a component, we shouldn't capitalize it.
```suggestion
  const sacXdrData = (dataString: string | null | undefined): string[] => {
```

PR #1527, opened 2025-08-01
Disable 'memo' in build tx for soroban operations for P23

jeesunikim:
- I disable it at `Prepare Transaction` button
- Added a note underneath the memo field

Screenshot 2025-07-31 at 5 29 57 PM

### DIFF
--- a/src/components/FormElements/MemoPicker.tsx
+++ b/src/components/FormElements/MemoPicker.tsx
@@ -2,11 +2,12 @@ import React from "react";
 import { MemoType, MemoValue } from "@stellar/stellar-sdk";
 import { Input } from "@stellar/design-system";
 
+import { Box } from "@/components/layout/Box";
 import { SdsLink } from "@/components/SdsLink";
 import { RadioPicker } from "@/components/RadioPicker";
 import { ExpandBox } from "@/components/ExpandBox";
+
 import { EmptyObj } from "@/types/types";
-import { Box } from "@/components/layout/Box";
 
 export type MemoPickerValue = {
   type: MemoType | string | undefined;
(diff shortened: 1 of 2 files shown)


src/components/FormElements/MemoPicker.tsx
```diff
@@ -2,9 +2,11 @@ import React from "react";
 import { MemoType, MemoValue } from "@stellar/stellar-sdk";
 import { Input } from "@stellar/design-system";
 
+import { SdsLink } from "@/components/SdsLink";
 import { RadioPicker } from "@/components/RadioPicker";
 import { ExpandBox } from "@/components/ExpandBox";
 import { EmptyObj } from "@/types/types";
+import { Box } from "@/components/layout/Box";
```
quietbits: Nit: import order
```suggestion
import { Box } from "@/components/layout/Box";

import { EmptyObj } from "@/types/types";
```

src/components/SmartContractJsonSchema/index.tsx
```diff
@@ -113,9 +113,13 @@ export const JsonSchemaForm = ({
     setSubmitTxError("");
     resetPrepareTx();
 
+    const updatedTxnParams = {
+      ...txnParams,
+      memo: {}, // starting with p23, soroban tx no longer support memos
+    };
```
quietbits, line 126: Instead of removing the memo behind the scenes, we could display a message if it's been removed. We can check that when a Soroban operation is selected. Also, we can disable memo picker for Soroban operations.

quietbits (approved):

src/components/SmartContractJsonSchema/index.tsx
jeesunikim: oh i like that

jeesunikim: we don't know that if it's a soroban opration until user selects the operation in the step 2. i will display a message once they select the soroban operation that the memo has been removed if they inputted.

PR #1532, opened 2025-08-06
[Invoke Contract] Add a tab view for json and base64

jeesunikim:
**Summary:**
- calling both xdrFormat `base64` and `json` via `Promise.all` and return responses in two different types using `<Tab />`
- use `<PrettyJson />` instead of `<PrettyJsonTransaction />`
json

base64

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
@@ -7,13 +7,12 @@ import {
   useState,
 } from "react";
 import {
+  Badge,
   Button,
   Card,
-  Select,
+  Icon,
   Text,
   Textarea,
-  Badge,
-  Icon,
   Tooltip,
 } from "@stellar/design-system";
 import { BASE_FEE, contract } from "@stellar/stellar-sdk";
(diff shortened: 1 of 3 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
```diff
@@ -661,50 +665,78 @@ export const InvokeContractForm = ({
...
+              <Button
+                size="md"
+                variant="secondary"
+                isLoading={isExtensionLoading || isSubmitRpcPending}
+                disabled={isSubmitDisabled}
+                onClick={handleSimulateAndSubmit}
+              >
+                Simulate & Submit
+              </Button>
             </Box>
```
quietbits: We shouldn't use absolute positioning for layout that affects the flow. In this case, the buttons are out of the layout flow and overlap the tabs. A couple of things we could do:
1. Update `TabView` component to accept an optional element to display on the right (and handle wrapping to make sure the layout doesn't break).
2. Don't place the buttons on the same line as tabs. To me, this looks a bit confusing: tabs now seem like buttons. But I agree that this makes the layout more balanced.

image

jeesunikim: Thank you for the suggestion. I will remove `absolute` and Minkyeong oked having a tab view in a separate line :)

PR #1533, opened 2025-08-06
[Invoke Contract] Show only the result by default; add 'show full response' toggle

jeesunikim:
https://github.com/user-attachments/assets/599b6618-1ad6-4141-8282-ee5fb00b2f64

- Display only the `results` by default
- Add a show full response toggle to display a full response


janewang:
This looks awesome!

PR #1537, opened 2025-08-13
Transaction Dashboard: Fee breakdown

quietbits:


elizabethengelman (commented):
This is looking great - I just compared this with the output of the CLI and everything matches! The one difference is that the CLI shows total refunded, rather than just refunded resource fee. So that is why the number on the very right of these screenshots is a bit different than the refunded section in lab. Perhaps we should make some updates to the CLI to match that! 

soroban tx:
Image

soroban with fee bump:
Image

PR #1549, opened 2025-08-18
Build Transaction: handle Invoke Contract Function op for SAC

quietbits:
Ideally, we should simplify this logic (there are too many chained events that might cause issues) and use the same in both Smart Contracts > Invoke Function and Build Transaction > Invoke Contract Function operation. I wanted to keep this update contained to only the Build Transaction to minimize the risk of breaking something.

SAC ID on Testnet: `CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC`
Contract ID on Testnet: `CDFDDMOZZLNU3QV73I35U4FIW6OS3AKB2N5XLMNHORFSUVJU2DF5XERJ`


src/components/FormElements/FetchContractMethodPickerWithQuery.tsx
```diff
@@ -34,45 +40,123 @@ export const FetchContractMethodPickerWithQuery = ({
...
-    if (isSuccess && memoizedMethods) {
-      const methodNames = memoizedMethods.map((method) =>
-        method.name().toString(),
-      );
+    if (contractType) {
+      setFetchType(isSacType ? "sac" : "contract");
+    }
+
+    if (contractTypeError) {
+      setFetchError(null);
```
jeesunikim: do we need to do both `null` then `contractTypeError?.message`?

jeesunikim (approved):

src/components/FormElements/FetchContractMethodPickerWithQuery.tsx
quietbits: That's a mistake, thanks for catching it! :raised_hands:

PR #1572, opened 2025-08-28
Home: add slider CTA

quietbits:


jeesunikim:
Screenshot 2025-08-28 at 10 42 12 AM

Screenshot 2025-08-28 at 10 42 07 AM

@quietbits I am noticing bottom padding discrepancies

src/styles/home.scss
```diff
@@ -167,7 +167,7 @@
       }
 
       .Lab__home__introSlider__cardItem__card {
-        height: pxToRem(344px);
+        height: pxToRem(400px);
```
jeesunikim, line 171: would `fit-content` work here?

quietbits: No, because we don't want the height to change when we open the larger cards or when we scroll through them.

PR #1576, opened 2025-08-28
Various tweaks

quietbits:


quietbits:
@stellar/security-team, could you please take a look at this when you have a chance? :pray:

playwright.config.ts
```diff
@@ -24,7 +24,7 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry */
-  retries: process.env.CI ? 2 : 1,
+  retries: process.env.CI ? 2 : 2,
```
jeesunikim, line 27: can we just do `2` instead of a ternary condition?

src/helpers/isValidUrl.ts
```diff
@@ -1,11 +1,7 @@
 export const isValidUrl = (url: string) => {
-  if (!url.startsWith("http")) {
-    return false;
-  }
-
   try {
-    new URL(url);
-    return true;
+    const parsedUrl = new URL(url);
+    return parsedUrl.protocol === "http:" || parsedUrl.protocol === "https:";
```
jeesunikim, line 4: should we only support `https:` for security

src/helpers/sanitizeUrl.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (e) {
+    // Return empty string if URL is invalid
+    return "";
+  }
+};
+
+const stripHtml = (str: string): string => {
+  return str.replace(/</g, "&lt;").replace(/>/g, "&gt;");
```
jeesunikim, line 48: til: nice

tests/unit/sanitizeUrl.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it("should preserve path", () => {
+    expect(sanitizeUrl("https://stellar.org/some/path")).toBe(
+      "https://stellar.org/some/path",
+    );
+  });
+
+  it("should not return a URL with a trailing slash if the path is empty", () => {
+    expect(sanitizeUrl("https://stellar.org")).toBe("https://stellar.org");
```
jeesunikim, line 53: the expect test doesn't match the description

src/helpers/isValidUrl.ts
jeesunikim: oh are we supporting `http:` for `localhost`?

src/helpers/jsonCipher.ts
```diff
@@ -0,0 +1,25 @@
...
+  return str
+    .split("")
+    .map((char, i) =>
+      String.fromCharCode(char.charCodeAt(0) ^ key.charCodeAt(i % key.length)),
+    )
+    .join("");
+};
+
+export const encryptJson = (data: object): string => {
+  return btoa(xorCipher(JSON.stringify(data), SALT));
```
kanwalpreetd, line 16: Seems like we are using `Networks.TESTNET` to encode the user secret key?

Just want to point out that this offers some level of obscurity, but it doesn't offer much protection. A more elaborate protection will be to derive the encryption/decryption key (for encrypting user's secret key) from a user provided secret, but that will affect the current workflow for the user: a user will need to provide that secret when we encrypt the secret key for storage and then again when they want to use it. Another option could be to never store the secret key in localstorage, and user provides it everytime they want to sign something.

The above requires significant changes to user experience, so I guess it is fine for now. But if we want to dial up security around the secret key usage, it is something we could think about in future (especially if changing the user experience is fine).

src/helpers/isValidUrl.ts
kanwalpreetd: Not sure if we need to support http generally, but if http is just required for `localhost`, perhaps we can only allow http where URL is like `http://localhost:*`, `http://127.0.0.1:*` or `http://[::1]:*` (for IPv6)

playwright.config.ts
quietbits: I'd like to keep this in the long form for now. I fixed some flaky tests in another PR, so hopefully, we can go back to only one try locally.

src/helpers/isValidUrl.ts
quietbits: The most common use case for `http` is probably `localhost`, but it could be any domain a developer uses locally. We can add extra checks for Mainnet network settings to make sure only `https` links are used.

tests/unit/sanitizeUrl.test.ts
quietbits: Would `pathname` instead of `path` be clearer?

image

src/helpers/jsonCipher.ts
quietbits: Yes, this is only to obscure saved test keypairs in local storage. We don't allow saving them on Mainnet. Adding proper security, requiring users to enter a password every time we need to access this (for example, to read saved keypairs to show them on the UI), won't be good UX, as you say. Maybe it's something we can revisit in the future.

tests/unit/sanitizeUrl.test.ts
jeesunikim: should it be `expect(sanitizeUrl("https://stellar.org/")).toBe("https://stellar.org");`

quietbits: Oh, I see what you mean. This test is to make sure we're not adding `/` at the end. If it's not there, then it's good by default.

jeesunikim (approved):

PR #1580, opened 2025-08-29
Bug fixes

quietbits:


src/app/(sidebar)/smart-contracts/contract-explorer/page.tsx
```diff
@@ -169,6 +169,23 @@ export default function ContractExplorer() {
...
   };
 
+  const handleLoadContract = async () => {
+    try {
+      // throwOnError will throw an error instead of setting the error state.
+      // This will prevent other queries from running when fetching a contract
+      // type fails.
+      await fetchContractType({ throwOnError: true });
+      await fetchContractInfo({ throwOnError: true });
+      await fetchWasmContractClient();
```
quietbits, line 179: With this update, if the `fetchContractType` fails, `fetchContractInfo` won't be called. Previously, we made both calls, which wasn't optimal.

src/app/(sidebar)/xdr/diff/page.tsx
```diff
@@ -229,6 +229,8 @@ export default function DiffXdr() {
               language="json"
               original={prettifyJsonString(jsonOne || DEFAULT_JSON_ONE)}
               modified={prettifyJsonString(jsonTwo || DEFAULT_JSON_TWO)}
+              keepCurrentModifiedModel={true}
+              keepCurrentOriginalModel={true}
```
quietbits, line 233: This fixes some issues where the Diff editor state was causing errors.

src/styles/home.scss
```diff
@@ -217,8 +217,6 @@
       overflow: visible;
 
       img {
-        max-width: 100%;
-        max-height: 100%;
```
quietbits, line 221: Next.js `Image` was complaining about changing image dimensions in CSS. I simplified the code to use static dimensions only, which works in this case.

src/components/WalletKit/WalletKitContextProvider.tsx
```diff
@@ -59,6 +59,11 @@ export const WalletKitContextProvider = ({
   }, []);
 
   const walletKitInstance = useMemo(() => {
+    // Only initialize on client side to avoid "window is not defined" errors in terminal
+    if (typeof window === "undefined") {
+      return undefined;
+    }
```
quietbits, line 65: This was causing errors in the terminal where `window` doesn't exist.

jeesunikim (approved):

PR #1588, opened 2025-09-03
Create macrohard

macrohard-cmd:


quietbits:
Thanks for the PR, but we don't need this in the Lab.

PR #1591, opened 2025-09-03
Fund Account: handle muxed account

quietbits:


src/app/(sidebar)/account/fund/page.tsx
```diff
@@ -80,6 +84,26 @@ export default function FundAccount() {
     return <SwitchNetwork />;
   }
 
+  const handleMuxedAccount = (muxedAddress: string) => {
+    if (muxedAddress.startsWith("M")) {
```
jeesunikim, line 88: can we use [isValidMed25519PublicKey](https://stellar.github.io/js-stellar-sdk/StrKey.html#.isValidMed25519PublicKey) method to check if it's a valid muxed address?

quietbits: At this point, we already know that it's a valid G or M address. I thought a simple check might be enough here. Do you think we should still do a proper check?

jeesunikim: ah i see. This is taken care of in `validate.getPublicKeyError`. Thanks

PR #1621, opened 2025-09-10
Deploy Contract page

quietbits:


src/components/FilePicker/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const containerClasses = [
+    "FilePicker",
+    ...(isDropInProgress ? ["FilePicker--active"] : []),
+    ...(isDropDisabled ? ["FilePicker--disabled"] : []),
+  ].join(" ");
+
+  return (
+    <div ref={dropAreaElRef} className={containerClasses}>
+      <div className="FilePicker__button" aria-disabled={Boolean(isDisabled)}>
+        <label
```
jeesunikim, line 180: can we use our existing `<Input />` that takes `leftElement` and `type`

quietbits: In SDS, `Input` styles are applied to the `input` (and container), but here we want to style the `label` as a button that triggers the hidden `input`. This is a special case, so we can't use the `Input` component here.

PR #1638, opened 2025-09-16
Apply redesign for getLedgerEntries

jeesunikim:
**Summary:**
- Apply the redesign for `getLedgerEntries`
- [Figma](https://www.figma.com/design/9j3YQ5URseAsn3j0tf2FFq/Laboratory?node-id=475-42717&t=2uFQ3F0c4FTvw9Fs-0)
- Use the new design for `XDR Format`
- fixed input fields not resetting upon clicking the `delete` icon

⚠️ claimable balance ledger key isn't rendering correctly. I created an issue #1639.

**Screenshots for each Ledger Key:**
- `account`
account-ledger
account-xdr

- `trustline`
trustline-alp4-ledger
trustline-lp-xdr

- `offer`
offer-ledger
offer-xdr

- `data`
data-ledger
data-xdr

- `liquidity pool`
lp-ledger
lp-xdr

- `contract data`
contract-data-ledger
contract-data-xdr

- `contract code`
contract-code-ledger
contract-code-xdr

- `config setting`
config-setting-ledger
config-setting-xdr

- `ttl`
ttl-ledger
ttl-xdr


src/components/FormElements/AssetPicker.tsx
```diff
@@ -196,6 +199,7 @@ export const AssetPicker = ({
           error: assetError?.issuer || "",
         }}
         disabled={disabled}
+        hasCopyButton={isReadOnly}
```
quietbits: Since `AssetPicker` is a generic component, I think we should have a dedicated prop `hasCopyButton`, and not tie it to the `isReadyOnly` prop. There might be cases where we want to have a copy button for a picker that is not ready only.

src/components/FormElements/XdrLedgerKeyPicker.tsx
```diff
@@ -276,17 +278,14 @@ export const XdrLedgerKeyPicker = ({
...
+    if (value) {
       validateLedgerKeyXdr(value);
     }
-  }, [
-    value,
-    ledgerKeyXdrToJsonString,
-    validateLedgerKeyXdr,
-    selectedLedgerKey,
-  ]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
quietbits, line 288: This is not needed anymore.

src/components/Tabs/styles.scss
```diff
@@ -23,6 +24,37 @@
...
+          position: absolute;
+          bottom: -2px;
+          left: 0;
+          left: 50%;
+          transform: translateX(-50%);
+        }
+        background-color: transparent;
+        --Tabs-default-text: var(--sds-clr-lilac-11);
+        --Tabs-default-background: transparent;
+        border-radius: 0;
```
quietbits: Nit: CSS variables should be as high up as possible to make it easier to notice them and not introduce unexpected behavior.

```suggestion
        --Tabs-default-text: var(--sds-clr-lilac-11);
        --Tabs-default-background: transparent;
        
        &:after {
          content: "";
          height: 2px;
          width: 95%;
          background: var(--sds-clr-lilac-11);
          position: absolute;
          bottom: -2px;
          left: 0;
          left: 50%;
          transform: translateX(-50%);
        }
        background-color: transparent;
        border-radius: 0;
```

src/components/FormElements/XdrLedgerKeyPicker.tsx
jeesunikim: it asked to add `validateLedgerKeyXdr` and I didn't want to add a function in there. wdyt

quietbits: Ahh, I see. Could we add a comment like [here](https://github.com/stellar/laboratory/blob/main/src/app/(sidebar)/account/fund/page.tsx#L73) saying what is excluded? It seems more intentional that way.

PR #1640, opened 2025-09-17
Transaction Dashboard: info + fee breakdown tests

quietbits:
- Added some optimizations to fix flaky tests and speed up runs
- Added transaction info and fee breakdown tests


playwright.config.ts
```diff
@@ -16,32 +16,35 @@ const baseURL = `http://localhost:${PORT}`;
...
   /* Run tests in files in parallel */
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry */
-  retries: process.env.CI ? 2 : 2,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  retries: 2,
+  workers: process.env.CI ? 2 : "25%",
```
jeesunikim, line 28: how did we come up with CPU core `25%`?

jeesunikim (approved):
one question

playwright.config.ts
quietbits: Trial and error :sweat_smile: . Too many workers at once produce too many flaky tests, so I tried to find the best balance between speed and stability that worked well on my machine.

PR #1655, opened 2025-09-25
Transaction Dashboard: Signatures tab

jeesunikim:


src/helpers/fetchTxSignatures.ts
```diff
@@ -54,11 +56,11 @@ export const fetchTxSignatures = async ({
       tx = tx.innerTransaction;
     }
 
-    sourceAccounts[convertMuxedAccountToEd25519Account(tx.source)] = [];
+    sourceAccounts[getPublicKey(tx.source)] = [];
```
jeesunikim, line 59: you can check that this works by comparing the following xdr: `AAAAAgAAAQAAAAAAAAAAAWFxfzA7qAZ1IadwQX6k0CezUgpXQCPz52f+IYvwDTtcAAAAZAADOlkAAAAUAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAABAAAAAMN3H5F5/3vVEpiPgud3bTKtfJLYzPuUzoD4zOujBu+wAAAAAAAAAABZaC8AAAAAAAAAAAHwDTtcAAAAQMOV7H6sLaIsXtVNKMIzSlCzA3hvpS9nU1DyPK0X2qLOGYILe+wgTlo+ZDNM1jReH1uP6NfomiLQ9eiTl4rYCgQ=` on pr preview vs. production site :)

src/app/(sidebar)/transaction-dashboard/components/Signatures.tsx
```diff
@@ -0,0 +1,193 @@
...
+      },
+    );
+  };
+
+  return (
+    <Box gap="md" addlClassName="Signatures">
+      <div className="Signatures__gridTableContainer">
+        <table role="table">
+          <thead>
+            <tr role="row">
```
quietbits: Nit: I don't think we need to specify the role if we use a native element.

src/app/(sidebar)/transaction-dashboard/styles.scss
```diff
@@ -228,3 +228,82 @@
...
+// =============================================================================
+// Signatures
+// =============================================================================
+.Signatures {
+  color: var(--sds-clr-gray-11);
+
+  &__text {
+    font-size: pxToRem(14px);
+    line-height: pxToRem(20px);
+    font-weight: var(--sds-fw-medium);
```
quietbits: If we use the `Text` component, these should be set through the component's props.

src/app/(sidebar)/transaction-dashboard/styles.scss
```diff
@@ -228,3 +228,82 @@
...
+    white-space: nowrap;
+
+    &[data-is-header="true"] {
+      text-align: left;
+      font-size: pxToRem(12px);
+      line-height: pxToRem(18px);
+      font-weight: var(--sds-fw-medium);
+      color: var(--sds-clr-gray-11);
+      white-space: nowrap;
```
quietbits: Nit: We should update only the props that are different. For example, `font-weight` and `white-space` don't change.

src/helpers/getPublicKey.ts
```diff
@@ -0,0 +1,13 @@
+import { MuxedAccount, StrKey } from "@stellar/stellar-sdk";
+
+// Helper to decode muxed account to ed25519 public key
+export const getPublicKey = (publicKey: string) => {
```
quietbits, line 4: I'm not sure it's clear that this helper can be used with any public key, as it sits in `src/helpers/decodeMuxedToEd25519.ts`. 🤔 Maybe name the file `getPublicKey.ts`?

src/components/TabView/index.tsx
```diff
@@ -67,7 +67,7 @@ export const TabView = ({
       <div className="TabView__content">
         {tabContent.map((tc) => (
           <div key={tc.id} data-is-active={activeTabId === tc.id}>
-            {tc.content}
+            {activeTabId === tc.id ? tc.content : null}
```
jeesunikim, line 70: @quietbits `txDashStateChange.test.ts` had an issue of finding elements after adding `<Signatures/>` because `<Signatures/>` has an async call like `isXdrInit()` and it waited for that to to be completed before rendering all the content in tabs via `TabView`. I updated `TabView` to only render the content if it's an active tab.

I didn't catch this issue locally - the tests passed fine but on a fresh build in github action, it complained that it couldn't find the element (hidden by css)

tests/e2e/txDashStateChange.test.ts
```diff
@@ -19,6 +19,7 @@ test.describe("Transaction Dashboard: State Change", () => {
   });
 
   test.describe("KALE", () => {
+    // ERROR
```
quietbits: Do we need this comment?

quietbits (approved):

tests/e2e/txDashStateChange.test.ts
jeesunikim: removed

Issue #1610, opened 2025-09-05
Use xdr-json for contract data json export


leighmcculloch:
The smart contract explorer shows a user the data stored for a contract and has an export to JSON button.

<img width="444" height="174" alt="Image" src="https://github.com/user-attachments/assets/82ffaebd-cc2c-4dc7-9c26-690bff0a0d30" />

The button provides a CSV file containing JSON, and that JSON appears to be the internal JavaScript objects for the js-stellar-base xdr types. That JSON is hard to work with, and is really an internal implementation detail of the js-stellar-base type system that nobody should be relying on.

<img width="454" height="67" alt="Image" src="https://github.com/user-attachments/assets/c76c3590-cb17-466c-9a3f-921e3b39d3ea" />

Instead the button should provide the JSON encoded in SEP-51 XDR-JSON.

Issue #1547, opened 2025-08-18
Contract invoke error result XDR is rendered with JS internals instead of XDR-JSON


leighmcculloch:
When using the contract invoke screen in the contract explorer:
- https://lab.stellar.org/smart-contracts/contract-explorer

if the transaction errors, the XDR is rendered in the response with the `@stellar/stellar-base` XDR internals.

The worse part of the base lib's XDR rendering as JS objects is that strings become integer arrays.

It would be easier to work with if the response was rendered with XDR-JSON instead. XDR-JSON renders XDR strings appropriately as ASCII. Rendering strings as text is critical to debugging why the invocation failed.

For example, this is what I see when an invocation fails:

<img width="974" height="938" alt="Image" src="https://github.com/user-attachments/assets/74447c65-46dd-4401-9158-b51956c2f751" />

leighmcculloch:
This is also happening on the tx submit page, which makes ti really hard to inspect the response for a transaction.

PR #1662, opened 2025-09-29
Use XDR-JSON instead of stellar base on 'export json' and 'error' response

jeesunikim:
**Summary:**
- tackles two issues #1547 and #1610
- parsing `errorResult` to xdr internal happens in [sdk code](https://github.com/stellar/js-stellar-sdk/blob/28e1e1b8986ce69eedd3978fae4e3ff1fe9312a1/src/rpc/parsers.ts#L28)

#1547 befor and after

before:

before-CDLZ7M4PMDCI3N67ZVTWYX6TJKRUVIEFZKVOJ5ET7BCEUDHC4FJ5NMCL

after:

after-CDLZ7M4PMDCI3N67ZVTWYX6TJKRUVIEFZKVOJ5ET7BCEUDHC4FJ5NMCL

#1610 before and after

before: 

1610-before

after:

1610-after

### DIFF
--- a/src/components/Home/Tutorials.tsx
+++ b/src/components/Home/Tutorials.tsx
@@ -9,7 +9,7 @@ export const Tutorials = () => {
       title: "Create Account",
       description: "Creates and funds a new Stellar account.",
       youTubeLink:
-        "https://www.youtube.com/embed/Sou5wUsfsZw?si=dtQFORe7YBkmApHe",
+        "https://www.youtube.com/embed/7j5t69f40dM?si=4svimk6VRVqnGmCj",
     },
     {
       title: "Payment",


src/query/useSubmitRpcTx.ts
```diff
@@ -39,6 +45,17 @@ export const useSubmitRpcTx = () => {
         const sentTx = await rpcServer.sendTransaction(transaction);
 
         if (sentTx.status !== "PENDING") {
+          if (sentTx.errorResult) {
+            const errorResult = parse(
+              StellarXdr.decode(
+                "TransactionResult",
+                sentTx.errorResult?.toXDR("base64"),
+              ),
```
quietbits, line 55: Should we make sure that `StellarXdr` is available (`useIsXdrInit()`)?

quietbits (approved):

src/query/useSubmitRpcTx.ts
jeesunikim: thank you! I will use `await StellarXdr.initialize();`

Issue #1663, opened 2025-09-30
Float Notification: implement auto closing


quietbits:
Add an optional `autoClose` prop to allow auto-closing notifications.

Issue #1528, opened 2025-08-02
Redesign Signing in Lab


janewang:
### What problem does your feature solve?

Lab should be able to support signing using wallet, as well as all supported private key in the Saved section. Reconsider how signing is surface across the product

minkyeongshin:
Final design here:

Component:
<img width="868" height="306" alt="Image" src="https://github.com/user-attachments/assets/c766052c-c993-4b09-9b78-1defaf102f07" />

Usecase:
<img width="857" height="842" alt="Image" src="https://github.com/user-attachments/assets/adbcf192-75e8-495e-9ed7-fdd193ae3a8e" />

jeesunikim:
I will work on this once `Deploy Contract` and design question are resolved

PR #1665, opened 2025-10-01
Deploy Contract updated UI

quietbits:
Updated Deploy Contract UI
- Renamed page to Upload and Deploy Contract.
- Created a separate component for Source Account (`SourceAccountPicker`) that should be updated in this issue (https://github.com/stellar/laboratory/issues/1528) to allow selecting saved accounts.
- Float notifications don't auto-close. It will be handled in this issue (https://github.com/stellar/laboratory/issues/1663).
- Created `SignTransactionXdr` component that handles all the signing with the new UI. There's a lot of copy/paste code from the Sign Transaction file, but this is only temporary. Once we validate that the new component works as expected, we can use it on the Sign Transaction page.

You can create new smart contract Wasm files to test [here](https://soropg.com/).

### DIFF
--- a/src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
+++ b/src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
@@ -9,6 +9,7 @@ import {
   Link,
   Icon,
   Text,
+  Alert,
 } from "@stellar/design-system";
 import { Address, contract, Operation } from "@stellar/stellar-sdk";
 import { parseContractMetadata } from "@stellar-expert/contract-wasm-interface-parser";
@@ -26,6 +27,7 @@ import { SignTransactionXdr } from "@/components/SignTransactionXdr";
 import { TransactionSuccessCard } from "@/components/TransactionSuccessCard";
 import { RpcErrorResponse } from "@/components/TxErrorResponse";
 import { SdsLink } from "@/components/SdsLink";
+import { SwitchNetworkButtons } from "@/components/SwitchNetworkButtons";
 
 import { useStore } from "@/store/useStore";
 import { useBuildRpcTransaction } from "@/query/useBuildRpcTransaction";
@@ -45,7 +47,6 @@ import { stellarExpertAccountLink } from "@/helpers/stellarExpertAccountLink";
 
 import { NetworkType } from "@/types/types";
 
-// TODO: handle futurenet
 export default function DeployContract() {
   const CONSTRUCTOR_KEY = "__constructor";
 
@@ -423,8 +424,23 @@ export default function DeployContract() {
   };
 
   const renderContent = () => {
-    if (network.id === "mainnet") {
-      return null;
+    if (network.id === "futurenet") {
+      return (
+        <div>
+          <Text size="sm" as="p">
+            Upload and deploy contract is not supported on Futurenet. Please
+            switch to Testnet or Mainnet to get started.
+          </Text>
+
+          <Box gap="md" direction="row">
+            <SwitchNetworkButtons
+              includedNetworks={["testnet", "mainnet"]}
+              buttonSize="md"
+              page="contract list"
+            />
+          </Box>
+        </div>
+      );
     }
 
     return (
@@ -595,6 +611,24 @@ export default function DeployContract() {
                       {pageError}
                     </Notification>
                   ) : null}
+
+                  {constructorSchema ? (
+                    <Alert
+                      title="This contract has a constructor."
+                      variant="primary"
+                      placement="inline"
+                    >
+                      Please add the arguments for the constructor.
+                    </Alert>
+                  ) : (
+                    <Alert
+                      title="This contract has no constructor."
+                      variant="primary"
+                      placement="inline"
+                    >
+                      There are no arguments to add for the constructor.
+                    </Alert>
+                  )}
 
                   {constructorSchema ? (
                     <Card>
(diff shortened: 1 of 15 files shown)


src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
```diff
@@ -396,392 +489,617 @@ export default function DeployContract() {
...
+                isExpanded={isUploadExpanded}
+                onDoneAction={setIsUploadExpanded}
+                txHash={submitUploadTxResponse?.result?.txHash}
+                networkId={network.id}
+              />
+
+              <ExpandBox offsetTop="sm" isExpanded={isUploadExpanded}>
+                <Box gap="md">
+                  <FilePicker
+                    key={`filePicker-${resetFilePicker}`}
```
quietbits, line 585: This is a nice way to reset the state in the child component. React will re-mount the component when the key changes.

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
jeesunikim: Is this supposed to use `useIsXdrInit()` instead of `StellarXdr`? I pulled the branch locally and noticed this.

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
```diff
@@ -176,88 +234,192 @@ export default function DeployContract() {
...
-      const stepRef = getStepRef(currentStep);
-
-      if (stepRef?.current) {
-        scrollElIntoView(stepRef);
-      }
+    if (isSubmitDeployTxSuccess) {
+      addFloatNotification({
+        id: "deploy-submit-tx-success",
+        title: "Contract deployed!",
+        description: `${selectedFile?.name} file has been successfully deployed.`,
```
jeesunikim, line 403: Image

can we wrap the text?

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
```diff
@@ -396,392 +493,617 @@ export default function DeployContract() {
...
+                    key={`deploy-tx-${signTxResetKey}`}
+                    id="deploy-tx"
+                    title="Add signature to deploy"
+                    xdrToSign={deployTx?.preparedXdr || null}
+                    onDoneAction={({
+                      signedXdr,
+                      successMessage,
+                      errorMessage,
+                    }) => {
+                      setSignedDeployTx(signedXdr);
```
jeesunikim, line 837: Thoughts on using `useReducer` for 

```
setSignedDeployTx(signedXdr);
setSignDeploySuccess(successMessage);
setSignDeployError(errorMessage);
```

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
quietbits: Yes, we should use the check from the hook here, thanks!

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
quietbits: Could you please clarify what you mean by "wrap" here?

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
quietbits: I don't have much experience using `useReducer`. This state is simple, and we're assigning values that we get back from the request. I don't know if using action here instead simplifies the logic (I would prefer keeping the state simple and not using an object). Plus, the overhead of setting up the reducer and actions. Do you think it would be better to use `useReducer` here?

src/app/(sidebar)/smart-contracts/deploy-contract/page.tsx
jeesunikim: > Could you please clarify what you mean by "wrap" here?

`successfully` is broken into two

src/components/SignTransactionXdr/index.tsx
```diff
@@ -0,0 +1,906 @@
...
+    secretKey: 0,
+    hardwareWallet: 0,
+    extensionWallet: 0,
+    signature: 0,
+  };
+
+  const [selectedTab, setSelectedTab] = useState(defaultSignatureType);
+
+  // Secret key
+  const [secretKeyInputs, setSecretKeyInputs] = useState<string[]>([""]);
```
jeesunikim, line 83: two suggestions::
1. can we break this large file into sub components for each tab? SecretKeyTab, ExtensionWalletTab, HardwareWalletTab, and SignatureTab?
2. create a custom hook for signature aggregation and use that in the parent component here (`SignTransactionXdr/index.tsx`) and lift state up?

jeesunikim:
https://github.com/user-attachments/assets/9b79244a-ac02-4c0b-b9a2-2036920cbbd6

Drag and drop isn't working on my end. The same file gets uploaded when I go via `Select a File`

quietbits:
> Should `clear signature` also reset the inputs?
> 
>  clear-signature.mov

This is the current behavior. Should we change it, @janewang? If we do, we should also change the button label "Clear signature".

https://github.com/user-attachments/assets/b2c79778-58de-4876-bc11-f441b6b59381

janewang:
@quietbits Oh yes, clear signature should clear the input. Good with change the name of the button to "Clear signature".

src/components/SignTransactionXdr/index.tsx
quietbits: Great suggestions. Let's polish this once we add the tests (https://github.com/stellar/laboratory/issues/1666). That way, it will be safer to refactor without breaking anything.

PR #1674, opened 2025-10-09
Update Signing in Lab

jeesunikim:
- When there's neither keypairs nor wallet connect

no-connected

- when there is only wallet connect
whe-only-connected

- when there are both keypairs and wallet connect

Screenshot 2025-10-14 at 7 03 54 AM

- keypairs in secret key mode

secret-key

- smaller screen

smaller-screen


src/components/FormElements/MultiPicker.tsx
```diff
@@ -74,18 +76,23 @@ export const MultiPicker = ({
...
-                        }}
-                        placement="right"
-                        icon={<Icon.Trash01 />}
-                        addlClassName="MultiPicker__delete"
-                      />
-                    ) : null
+                    <>
+                      {typeof rightElement === "function"
+                        ? rightElement(index)
+                        : rightElement}
```
quietbits: Can `rightElement` be anything other than a function defined in type (`rightElement?: (index: number) => React.ReactNode;`) or undefined?

src/components/SignerSelector/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  label: string;
+  items: string | SavedKeypair[];
+  onChange: (val: string) => void;
+  mode: SignerMode;
+}) => {
+  const itemsArray: Array<SavedKeypair | { publicKey: string }> = Array.isArray(
+    items,
+  )
+    ? items
+    : [{ publicKey: items }];
```
quietbits: To simplify this logic, we could always pass in an array of items. So we can handle this directly for WalletKit.

src/components/SignerSelector/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+            key={item.publicKey}
+            onClick={() => {
+              const value = isSavedKeypair
+                ? renderKey(item as SavedKeypair).value
+                : item.publicKey;
+              onChange(value);
+            }}
+          >
+            {isSavedKeypair
+              ? `[Test Keypair ${index + 1}] ${renderKey(item as SavedKeypair).label}`
```
quietbits: Should we use the saved keypair name here instead? We could set the max name length to make sure it won't get too long.

Image

src/components/SignerSelector/styles.scss
```diff
@@ -0,0 +1,58 @@
...
+  &__options {
+    padding: pxToRem(4);
+    border-radius: pxToRem(4);
+
+    &__item {
+      @media (hover: hover) {
+        &:hover {
+          background-color: var(--sds-clr-gray-04);
+        }
+      }
```
quietbits: Is this used anywhere?

jeesunikim: yes, when we hover over an item

Screenshot 2025-10-10 at 4 38 20 AM

src/components/SignerSelector/index.tsx
jeesunikim: @janewang question! However, I wonder if it'd frustrate users - what would be the ideal max length? All my previous names are pretty long lol

src/components/FormElements/MultiPicker.tsx
jeesunikim: at the moment, no!

quietbits: Let's simplify the check, then. We can update in the future if the type changes.

src/components/SignerSelector/index.tsx
quietbits: Having the name for the keypair defeats the purpose, then, if I need to remember the address. :thinking:

jeesunikim: true - should we remove public key then to give more space?

janewang: I think this is a bit of a design question but I would prioritize name first and then perhaps the g address as a sanity check. Is there an option to see more?

janewang: I think this is a bit of a design question but I would prioritize name first and then perhaps the g address as a sanity check. Is there an option to see more?

jeesunikim: I tried: 

updated

src/components/FormElements/PubKeyPicker.tsx
```diff
@@ -14,31 +14,41 @@ interface PubKeyPickerProps extends Omit<InputProps, "fieldSize"> {
...
+        onChange={onChange}
+        infoLinkIcon={<Icon.InfoCircle />}
+        {...props}
+      />
+    </div>
+  ),
 );
+
+// eslint rule react/display-name
+PubKeyPicker.displayName = "PubKeyPicker";
```
jeesunikim: eslint yelled at me to add this 👀 you have more experience with forward ref (design system). does this look good @quietbits

src/components/SignerSelector/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  Button: (props: ButtonProps) => JSX.Element;
+  Dropdown: (props: DropdownProps) => JSX.Element;
+}
+
+const getTitle = ({ mode }: { mode: SignerMode }) => {
+  switch (mode) {
+    case "public":
+      return "Get address";
+    case "secret":
+      return "Use secret key";
```
quietbits, line 38: We should add `default` case even if it's empty. I think some of the linters might not be happy if we don't have it.

src/components/SignerSelector/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  if (mode === "secret" && !hasKeypairs) {
+    return <></>;
+  }
+
+  // Public Signer mode with only wallet - show direct button
+  if (mode === "public" && !hasKeypairs && hasWallet) {
+    return (
+      <InputSideElement
+        variant="button"
+        onClick={() => onClick()}
```
quietbits:
```suggestion
        onClick={onClick}
```

src/components/SignerSelector/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+      >
+        Get connected wallet address
+      </InputSideElement>
+    );
+  }
+
+  return (
+    <InputSideElement
+      variant="button"
+      onClick={() => onClick()}
```
quietbits:
```suggestion
      onClick={onClick}
```

src/components/SignerSelector/styles.scss
```diff
@@ -0,0 +1,74 @@
...
+    top: calc(var(--sds-gap-xxl) * 2);
+    transform: translateY(0);
+
+    left: 0;
+    right: 0;
+    z-index: 1000;
+
+    background-color: var(--sds-clr-gray-01);
+    border: 1px solid var(--sds-clr-gray-06);
+    border-radius: pxToRem(4);
```
quietbits: We should add `px` for consistency, even if it works without it.

```suggestion
    border-radius: pxToRem(4px);
```

src/components/SignerSelector/styles.scss
```diff
@@ -0,0 +1,74 @@
...
+    padding: pxToRem(3) pxToRem(4);
```
quietbits:
```suggestion
    padding: pxToRem(3px) pxToRem(4px);
```

src/components/SignerSelector/styles.scss
```diff
@@ -0,0 +1,74 @@
...
+
+    max-height: pxToRem(240px);
+    overflow-y: auto;
+
+    &__item {
+      &__label {
+        font-size: pxToRem(12px);
+        line-height: pxToRem(18px);
+
+        padding: pxToRem(3) pxToRem(8);
```
quietbits:
```suggestion
        padding: pxToRem(3px) pxToRem(8px);
```

src/components/SignerSelector/styles.scss
```diff
@@ -0,0 +1,74 @@
...
+        color: var(--sds-clr-gray-12);
+        font-size: pxToRem(14px);
+        line-height: pxToRem(20px);
+        font-weight: var(--sds-fw-medium);
+        display: grid;
+        align-items: center;
+
+        gap: pxToRem(8px);
+
+        padding: pxToRem(3) pxToRem(8);
```
quietbits:
```suggestion
        padding: pxToRem(3px) pxToRem(8px);
```

PR #1677, opened 2025-10-13
Transaction Dashboard: Events

quietbits:
Example: `b01bb450a2e34c95bad30ad4ad99b9682f3653c9449c2adeb267af899d5668d3` (Mainnet)

image


janewang:
@minkyeongshin Do you think we are fully using the horizontal space. We truncate the contract address `ABCD...EFGH` for consistency but I wonder if we should show this data here in full?

cc @quietbits Let us know your thoughts as well.

janewang:
>  The only drawback to showing the full contract ID is that we can't shorten it dynamically
Let's leave them as is then, thank you!

jeesunikim (approved):

PR #1681, opened 2025-10-15
Contract Explorer: UI updates

quietbits:
image

image


jeesunikim:
1. This change doesn't affect SAC `CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA`. We can implement this for SAC separately. FYI that SAC doesn't render `Contract & Env Meta`, we might want to disable this button

1681-sac

3. `ENUMS` is not rendering. Here's a comparison to SE. `CBHQGTSBJWA54K67RSG3JPXSZY5IXIZ4FSLJM4PQ33FA3FYCU5YZV7MZ` on TESTNET

PR #1683, opened 2025-10-16
Transaction Dashboard: update not found UI

quietbits:
- Update not found UI when the transaction is not found.
- Clean up mock data (removed diagnostic events) to make files shorter.

image


src/app/(sidebar)/transaction-dashboard/page.tsx
```diff
@@ -67,9 +71,43 @@ export default function TransactionDashboard() {
...
+              : "https://testnet.lumenscan.io/txns";
+          openUrl(`${lsUrl}/${transactionHashInput}`);
+        }
+      },
+    },
+    {
+      id: "goldsky",
+      label: "Goldsky",
+      onClick: () => {
+        openUrl("https://goldsky.com/chains/stellar");
```
jeesunikim, line 106: why do we link to a general goldsky site without a hash link?

quietbits: How can we link to it without creating an account?

jeesunikim: even with a goldsky account, goldsky is not a blockchain explorer. why do we have a goldsky link here?

quietbits: I think we also want to include indexers. @janewang, is this link OK? Or should we link to something else? :pray:

jeesunikim: I see! @janewang, I like the idea of promoting indexer, but ux on retrieving transaction is quiet complicated compared to the other two block chain explorer links. 

create-pipeline

janewang: Hey @johncanneto, is there an endpoint we could use to retrieve transactions from Goldsky? Thank you in advance 🙏

PR #1685, opened 2025-10-20
API Explorer: updated intro page

quietbits:
image


jeesunikim (approved):

src/app/(sidebar)/endpoints/components/EndpointsLandingPage.tsx
```diff
@@ -1,65 +1,173 @@
...
+          ))}
+        </div>
+      </HomeSection>
+
+      <HomeSection
+        title="Save & Sharing"
+        description="You can save methods in the browser’s local storage and get shareable link to provide to others."
+      >
+        <div className="Endpoints__imageBox">
+          {/* eslint-disable-next-line @next/next/no-img-element */}
```
jeesunikim, line 84: how come we don't use `Image` from NextJS

quietbits: Because `Image` requires setting a fixed width and height, but this image is dynamic.

PR #1686, opened 2025-10-22
Fix: Smart Contract Invoke excludes arguments after a `Vec`

jblewnormal:
fix(sorobanUtils): replace `return` statements with `continue` in `getScValsFromArgs()`

When invoking a smart contract function that has heterogeneous argument types, line 531 of the `sorobanUtils.ts` iterates through each argument, parses it, but mistakenly returns the parsed value - ending the for loop before iterating through every argument.

You can recreate the original bug that prompted this PR by using `CCPHUHQYFOJJ6WQUGUYHHPJYQGFLRQHJJTRJNWQG54MHCHPRFLWQI7SE` and calling the `set_privileged_addrs` function. Pass in any assortment of addresses and you'll notice you receive this error:

Screenshot 2025-10-22 at 1 04 06 PM

The arguments logged at the end of this error do not include the final argument `system_fee_admin`. This is because of this bug.

This results in function arguments being excluded from execution and the invocation/simulation therefore failing.

This commit replaces these `return` statements with `continue` so the for loop can complete iteration through all passed arguments.

### DIFF
--- a/tests/e2e/signerSelector.test.ts
+++ b/tests/e2e/signerSelector.test.ts
@@ -119,7 +119,7 @@ test.describe("Signer Selector", () => {
       await pageContext.getByText("Use secret key").nth(1).click();
       const { values: values2 } = await validateSignerSelectorOptions(
         pageContext,
-        multiPickerInputs.nth(1),
+        // multiPickerInputs.nth(1),
       );
       await values2.nth(1).click();
       await expect(multiPickerInputs.nth(1).locator("input")).toHaveValue(


quietbits:
Thanks for the PR, @jblewnormal ! :tada: @jeesunikim will review it next week when she's back.

quietbits:
@jblewnormal, thanks for digging into this. Really appreciate the help. :heart:

tests/e2e/signerSelector.test.ts
```diff
@@ -119,7 +119,7 @@ test.describe("Signer Selector", () => {
       await pageContext.getByText("Use secret key").nth(1).click();
       const { values: values2 } = await validateSignerSelectorOptions(
         pageContext,
-        multiPickerInputs.nth(1),
+        // multiPickerInputs.nth(1),
```
jeesunikim, line 122: why was this commented out? @jblewnormal

jblewnormal: This test was failing since `validateSignerSelectorOptions` only takes in a single parameter, not two. Happy to remove the comment before merging if you'd like.

jeesunikim: Thanks for catching it!

jeesunikim:
good afternoon, I will work on this tomorrow! Thanks!

PR #1701, opened 2025-11-12
fix transaction explorer setLedger not setting properly

jeesunikim:
- added a loading message when it is fetching instead of the default "No transactions found."
- shortened the fetching interval to `1` seconds instead of `5` for `LOCAL NETWORK`
- there was an error with `const txs = Array.from(transactions.values()).toSorted((tx) => tx.createdAt)`. It had a bug `TypeError: Cannot convert a BigInt value to a number at Array.toSorted (<anonymous>)` which stopped the execution of the code, which blocked `setStartLedger` from setting, which caused `useGetRpcTxs` to fetch the latest since no `startLedger` was set

### DIFF
--- a/src/app/(sidebar)/transactions-explorer/page.tsx
+++ b/src/app/(sidebar)/transactions-explorer/page.tsx
@@ -59,6 +59,11 @@ export default function Explorer() {
     startLedger,
   });
 
+  const isLocalNetwork =
+    network.id === "custom" &&
+    network.rpcUrl.includes("localhost:8000/rpc") &&
+    network.passphrase === "Standalone Network ; February 2017";
+
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/app/(sidebar)/transactions-explorer/page.tsx
```diff
@@ -90,7 +92,7 @@ export default function Explorer() {
         return;
       }
 
-      setNextFetchAt(Date.now() + 5000);
+      setNextFetchAt(Date.now() + 3000);
```
leighmcculloch: Can this be 1000 for when connected to a local instance? Local instances close ledgers every 1 second.

leighmcculloch (approved):

PR #1706, opened 2025-11-13
Contract Explorer: Build Info verification instructions

quietbits:
image

### DIFF
--- a/src/components/SmartContractJsonSchema/renderArrayType.tsx
+++ b/src/components/SmartContractJsonSchema/renderArrayType.tsx
@@ -53,7 +53,7 @@ export const renderArrayType = ({
       return renderer({
         name: nestedPath,
         schema: item,
-        path: [nestedPath],
+        path: [...path, String(index)],
         parsedSorobanOperation,
         onChange,
         formError,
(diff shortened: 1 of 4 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/BuildInfo.tsx
```diff
@@ -15,26 +15,86 @@ export const BuildInfo = ({
...
+                <Text as="h4" size="sm" weight="medium">
+                  Workflow Permissions
+                </Text>
+
+                <Text as="div" size="sm">
+                  In order to create a release, the workflow needs{" "}
+                  <code>id-token: write</code>, <code>contents: write</code> and{" "}
+                  <code>attestations: write</code> permissions. Default workflow
+                  permissions for a repository can be found at{" "}
+                  {'"Settings" -> "Actions" -> "Workflow permissions"'}.{" "}
```
quietbits, line 73: Adding `{}` to escape characters (we can't make the whole paragraph a template because of `code` tags).

quietbits:
> Then should we remove `Build Info` from the menu? Or include a message that Build Info is only available for contracts with WASM files on chain

We landed on tagging SAC contracts as `Built-in Contract`. Usually we prefer to disable tabs instead of removing them to avoid layout shift. We can change that if needed, but that wouldn't be part of this ticket.

image

jeesunikim (approved):

PR #1731, opened 2025-11-25
Transaction Explorer: Token Summary

quietbits:
Test transactions on Mainnet
- `a85b2c8821c39e53b4cbc500f4a040f0899349d1fb5e25446b29c478730fefbd` - SAC
- `43642f6951928e7378d30ebd52a640b5e88d942330f93f331d9d4d3ff9b7f7d7 `- SoroSwap
- `b6f662cf1bc0a8270bca6d963dc3222c94e8bb15c31cb63e1f21d382ab299476` - KALE (no transfer events)

image

### DIFF
--- a/src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
+++ b/src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
@@ -74,6 +74,8 @@ export const TokenSummary = ({
   const seNetwork = getStellarExpertNetwork(network.id);
 
   const formatAssetAddress = (address: string) => {
+    const isContractAddress = address.startsWith("C");
+
     return (
       <Box
         gap="sm"
@@ -81,7 +83,9 @@ export const TokenSummary = ({
         align="center"
         addlClassName="TransactionTokenSummary__address"
       >
-        <Link href={`${STELLAR_EXPERT}/${seNetwork}/contract/${address}`}>
+        <Link
+          href={`${STELLAR_EXPERT}/${seNetwork}/${isContractAddress ? "contract" : "account"}/${address}`}
+        >
           {shortenStellarAddress(address)}
         </Link>
 


src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
```diff
@@ -0,0 +1,212 @@
...
+      const addressTo = ev.body.v0.topics[2].address;
+
+      const token = ev.body.v0.topics?.[3]?.string;
+
+      let assetCode = "";
+      let assetIssuer = "";
+
+      if (token) {
+        [assetCode, assetIssuer] = token.split(":");
+      }
```
quietbits, line 201: Token is optional in a transfer event as seen in `80a57738bc6e0ce01a1ee25dbb1e59a4c4f677e0b8f0a2232c285b3a14990f9a` transaction. We can't know for sure what properties an event has until we get bindings. This is our best attempt not to break anything. 🤞 

Thanks, @jeesunikim , for catching this!

src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
```diff
@@ -0,0 +1,212 @@
+import { Fragment } from "react";
```
jeesunikim, line 1: is this what you prefer vs. `<></>`

quietbits: I prefer `<></>` where possible. In cases like this, we need to add a `key` prop, so `Fragment` is our only option.

src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
```diff
@@ -0,0 +1,212 @@
...
+        </CopyText>
+      </Box>
+    );
+  };
+
+  const formatAssetAmount = (amount: string) => {
+    return (
+      <span className="TransactionTokenSummary__amount">
+        {formatAmount(
+          new BigNumber(amount).dividedBy(10_000_000).toNumber(),
```
jeesunikim, line 107: this is useful — @janewang this outputs amount `"1000000"` in `i128` to `0.10`. I can see this conversion being useful for `invoke contract`. can we add it?

src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const formatAssetAddress = (address: string) => {
+    return (
+      <Box
+        gap="sm"
+        direction="row"
+        align="center"
+        addlClassName="TransactionTokenSummary__address"
+      >
+        <Link href={`${STELLAR_EXPERT}/${seNetwork}/contract/${address}`}>
```
jeesunikim: there are cases when it's a public key. we need to differentiate the url to be either `/account/` or `/contract/`

src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
```diff
@@ -0,0 +1,212 @@
...
+      }
+    });
+
+    // Return only groups with items
+    return Object.values(groups).filter((group) => group.items.length);
+  };
+
+  const { network } = useStore();
+
+  const seNetwork = getStellarExpertNetwork(network.id);
```
jeesunikim, line 74: this means we can only check for `public` or `testnet` network. I wonder if re-directing them to `getLedgerEntries` endpoint with the account pre-filled works for `local` or `futurenet`? @janewang

jeesunikim (approved):

src/app/(sidebar)/transaction-dashboard/components/TokenSummary.tsx
quietbits: Good call, I'll fix it.

PR #1740, opened 2025-12-02
Transaction Explorer: Resource Profiler

quietbits:
We're dynamically fetching network limits for Mainnet, Testnet, and Futurenet in `predev` and `prebuild` scripts, which run before every `dev` and `build` scripts automatically.

Example transaction on Mainnet: `43642f6951928e7378d30ebd52a640b5e88d942330f93f331d9d4d3ff9b7f7d7`

image


src/helpers/getTxResourceBreakdown.ts
```diff
@@ -0,0 +1,302 @@
+import { RpcTxJsonResponse } from "@/types/types";
+
+// Network limits
+export const NETWORK_LIMITS = {
```
jeesunikim: where do we get this? can we add a source link

quietbits: It's from the same link we have in the tab https://developers.stellar.org/docs/networks/resource-limits-fees. I'll add it here.

scripts/fetch-network-limits.mjs
```diff
@@ -0,0 +1,255 @@
...
+`;
+  }
+  content += `} as const;
+`;
+
+  return content;
+};
+
+const main = async () => {
+  console.log("🚀 Fetching Stellar network limits...\n");
```
quietbits, line 188: We can simplify the output for this if this feels too noisy.

package.json
```diff
@@ -7,7 +7,10 @@
     "node": ">=22.0.0"
   },
   "scripts": {
+    "fetch-limits": "node scripts/fetch-network-limits.mjs",
+    "predev": "pnpm fetch-limits",
     "dev": "NEXT_PUBLIC_COMMIT_HASH=$(git rev-parse --short HEAD) NEXT_PUBLIC_ENABLE_EXPLORER=true next dev",
+    "prebuild": "pnpm fetch-limits",
```
quietbits: `predev` and `prebuild` scripts run automatically before `dev` and `build` respectively (npm feature).

scripts/fetch-network-limits.mjs
```diff
@@ -0,0 +1,254 @@
...
+        ?.tx_max_contract_events_size_bytes,
+    contract_data_key_size_bytes:
+      dataKeySizeEntry?.dataJson?.config_setting?.contract_data_key_size_bytes,
+    contract_max_size_bytes:
+      maxSizeEntry?.dataJson?.config_setting?.contract_max_size_bytes,
+  };
+};
+
+const generateTypeScriptFile = (networkLimits) => {
+  const timestamp = new Date().toISOString();
```
jeesunikim: do we need this? i don't think it's used in this codebase

scripts/fetch-network-limits.mjs
```diff
@@ -0,0 +1,254 @@
...
+
+  let content = `// This file is auto-generated by scripts/fetch-network-limits.js
```
jeesunikim: `.mjs` not `.js`

scripts/fetch-network-limits.mjs
```diff
@@ -0,0 +1,254 @@
...
+
+  const networkLimits = {};
+
+  for (const network of NETWORKS) {
+    console.log(`📡 Fetching limits for ${network.name}...`);
+
+    if (!network.rpcUrl) {
+      console.error(`  ❌ Error: RPC URL not configured for ${network.name}`);
+      console.error(
+        `  Please add the RPC URL in scripts/fetch-network-limits.js\n`,
```
jeesunikim: `.mjs` not `.js`

scripts/fetch-network-limits.mjs
quietbits: Ahh, thanks for catching these! I missed these after refactoring.

Issue #1733, opened 2025-11-26
Contract parameter ordering bug


jsmaxi:
To reproduce the issue, use a lockable contract address (deployed on testnet).
https://github.com/jsmaxi/soroban-lockable-token
CB4TBVSEBM6437GLPLNA3KBPMTBWE6DIHB674ARKO4HRN6XLYWMD7ZQK

Open Contract Explorer:
https://lab.stellar.org/smart-contracts/contract-explorer?$=network$id=testnet&label=Testnet&horizonUrl=https:////horizon-testnet.stellar.org&rpcUrl=https:////soroban-testnet.stellar.org&passphrase=Test%20SDF%20Network%20/;%20September%202015;;

Enter Contract ID:
CB4TBVSEBM6437GLPLNA3KBPMTBWE6DIHB674ARKO4HRN6XLYWMD7ZQK

Load contract > Invoke contract.

Find the `lock` function.
Enter the parameters out of sequential order, for example
`of` : your account's public address
`reason` : 414243440000000000000000000000000000000000000000000000000000000a
`amount` : 10
`time` : 60

Click Simulate.
Output:
```
Prepare: HostError: Error(WasmVm, InvalidAction) Event log (newest first): 0: [Diagnostic Event] contract:CB4TBVSEBM6437GLPLNA3KBPMTBWE6DIHB674ARKO4HRN6XLYWMD7ZQK, topics:[error, Error(WasmVm, InvalidAction)], data:["VM call trapped: UnreachableCodeReached", lock] 1: [Diagnostic Event] topics:[fn_call, CB4TBVSEBM6437GLPLNA3KBPMTBWE6DIHB674ARKO4HRN6XLYWMD7ZQK, lock], data:[GD..., Bytes(414243440000000000000000000000000000000000000000000000000000000a), 10, 60] 
```

Notice that the parameter data is in an incorrect order: `[GD..., Bytes(414243440000000000000000000000000000000000000000000000000000000a), 10, 60]`

---

Next, reload the contract and enter the parameters in the correct sequential order, then click Simulate again. This time, the output returns success.

---

First observed in: https://github.com/stellar/laboratory/issues/1676 and https://github.com/stellar/laboratory/pull/1686

PR #1756, opened 2025-12-05
Fix contract parameter ordering bug

jeesunikim:
- arguments need to match the order of the schema
- run `orderArgsBySchema` when converting native to scval

**contracts I used for testing**
1. multi_args on `CBHQGTSBJWA54K67RSG3JPXSZY5IXIZ4FSLJM4PQ33FA3FYCU5YZV7MZ`
2. description in https://github.com/stellar/laboratory/issues/1733

**Summary: **
1. Initialize all keys in `SorobanInvokeValue.args` upfront from schema
2. Only update existing keys (we never add new ones)

- [ ] Add a code that checks mismatch on submission and add a log


src/helpers/sorobanUtils.ts
```diff
@@ -229,14 +234,75 @@ export const getTxWithSorobanData = ({
...
+export const orderArgsBySchema = (
+  args: AnyObject,
+  schema?: Record<string, any>,
+): AnyObject => {
+  if (!schema?.properties || typeof args !== "object" || Array.isArray(args)) {
+    return args;
+  }
+
+  const schemaKeys = Object.keys(schema.properties);
+  const orderedArgs: AnyObject = {};
```
quietbits: Should we use an array to keep the order? In an object, the order is not guaranteed. Or maybe add a sort index or something if we have to have an object here. 🤔

PR #1760, opened 2025-12-05
Optimize Dependabot

quietbits:
Updated cadence to once a month (we usually update all dependencies once a month, so this aligns better)
  - npm and GitHub Actions: Ignores patches, gets minor/major updates, 1 PR monthly for each bucket
  - Docker: Gets patch/minor updates (for security), ignores major versions (we use LTS versions), 1 PR monthly


jeesunikim (approved):
ty!!

PR #1778, opened 2025-12-17
Fund account: update tests

quietbits:


tests/e2e/fundAccountPage.test.ts
```diff
@@ -70,163 +81,271 @@ test.describe("[futurenet/testnet] Fund Account Page", () => {
...
+          });
+        } else {
+          await route.fulfill({
+            status: 200,
+            contentType: "application/hal+json; charset=utf-8",
+            body: JSON.stringify(MOCK_ACCOUNT_FUNDED_RESPONSE_WITH_USDC),
+          });
+        }
+
+        accountsCallCount++;
```
jeesunikim, line 114: I don't understand this part `accountsCallCount++`. Does test get re-run after `accountsCallCount++` is called?

quietbits: In this flow, we need to make multiple `accounts` calls (initial fetch, after fund XLM, after added trustline). The counter is used to set the mock response for each step.

PR #1792, opened 2026-01-02
Update JSON Response Component

jeesunikim:
Summary:
- Replaced `PrettyJson` with `CodeEditor` for big transaction response. 
- Responses with custom content like `PrettyJsonTransaction` is still using `PrettyJson` so that links and tags are still working as expected. `CodeEditor` doesn't allow custom content
-- `JsonCodeWrapToggle` removed from ones replaced with `CodeEditor`
-- `InvokeContract`'s simulation response's format is now located in the header in dropdown format
- CodeEditor updates:
-- Added calculating height based on number of lines for <CodeEditor/>. It makes it look better for RPC payload
-- `title` is now optional

**Screenshots:**
- (Tx Response) when transaction failed response
on-error

- `PrettyJson` style update
Screenshot 2026-01-02 at 3 04 23 PM

- RPC method payload
rpc-endpoint-payload

- Simulation Result
simualte-tx-base64

- Simulation Response on Invoke Contract
invoke-contract

### DIFF
--- a/src/app/(sidebar)/transaction/sign/components/Overview.tsx
+++ b/src/app/(sidebar)/transaction/sign/components/Overview.tsx
@@ -453,7 +453,14 @@ export const Overview = () => {
     let errorMsg = "";
 
     if (!isClear) {
-      const txSig = txHelper.decoratedSigFromHexSig(sigInputs);
+      const existingSigs = sign.importXdr
+        ? txHelper.extractSignaturesFromTx({
+            txXdr: sign.importXdr,
+            networkPassphrase: network.passphrase,
+          })
+        : [];
+
+      const txSig = txHelper.decoratedSigFromHexSig(sigInputs, existingSigs);
 
       signature = txSig.signature;
       successMsg = txSig.successMsg || "";
--- a/src/components/NetworkSelector/index.tsx
+++ b/src/components/NetworkSelector/index.tsx
@@ -147,16 +147,29 @@ export const NetworkSelector = () => {
             savedPreviousNetwork?.passphrase || networkPreset?.passphrase || "",
         } as Network;
 
+        const mainnetPreset = getNetworkById("mainnet");
+
         // Loaded network settings
-        const newNetwork = {
-          ...network,
-          label: networkPreset?.label,
-        } as Network;
+        const newNetwork =
+          network.passphrase === mainnetPreset?.passphrase
+            ? {
+                ...network,
+                ...mainnetPreset,
+                rpcUrl: network.rpcUrl,
+                horizonUrl: network.horizonUrl,
+              }
+            : ({
+                ...network,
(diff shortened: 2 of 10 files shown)


src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
```diff
@@ -588,40 +587,21 @@ export const InvokeContractForm = ({
   const renderResponse = () => {
     if (jsonResponse?.fullResponse || base64Response?.fullResponse) {
       return (
-        <TabView
```
jeesunikim, line 591: We use a dropdown to select xdr or json view. Therefore, it no longer needs `TabView`

src/components/CodeEditor/index.tsx
```diff
@@ -11,8 +11,10 @@ import "./styles.scss";
 
 export type SupportedLanguage = "json" | "xdr" | "text" | "interface";
 
+const LINE_HEIGHT_AND_PADDING_IN_PX = 20;
+
 type CodeEditorProps = {
-  title: string;
+  title?: string;
```
jeesunikim, line 17: `title` looks weird on payload section on rpc endpoint so I made it optional

src/components/CodeEditor/index.tsx
```diff
@@ -67,6 +77,19 @@ export const CodeEditor = ({
     }
   };
 
+  const handleEditorDidMount: OnMount = (editor) => {
+    const model = editor.getModel();
+    const lineCount = model?.getLineCount() || 0;
+    const titleHeight = title ? headerHeight : 0;
+
+    if (isAutoHeight) {
```
jeesunikim, line 89: `autoHeight` makes rpc payload looks much better (they're shorter in content)

src/components/CodeEditor/index.tsx
```diff
@@ -67,6 +77,19 @@ export const CodeEditor = ({
     }
   };
 
+  const handleEditorDidMount: OnMount = (editor) => {
```
quietbits, line 80: To optimize this a bit, we could return early if `isAutoHeight` is not enabled. Then we don't need the other check.

```suggestion
  const handleEditorDidMount: OnMount = (editor) => {
    if (!isAutoHeight) return; 
```

src/components/PrettyJson/styles.scss
```diff
@@ -15,6 +16,10 @@
   &__nested {
     padding-left: pxToRem(16px);
 
+    .PrettyJson__bracket {
+      color: #319331; // To match Monaco Editor's green color for nested brackets
```
quietbits: We should use CSS variables for these and set light and dark mode colors. I think we use dark and light modes in MonacoEditor, so we should be able to get both colors.

PR #1794, opened 2026-01-08
Add network limits for testnet, mainnet, futurenet

jeesunikim:
I'll handle tests in #1795

Link: https://laboratory-pr1794.previews.kube001.services.stellar-ops.com/network-limits

**Notable Changes:**
1. adding `stateArchivalEntry`, `bandwidthEntry`, `historicalDataEntry` in `[fetch-network-limits.mjs](https://github.com/stellar/laboratory/pull/1794/changes#diff-e9834fc7d041335182e55505454c08e271b27df08879803e329e16bc4996101e)`
2. `/network-limits` page with custom tables similar to Resource Fee in Smart Contract
3. Updated `formatFileSize` since the doc indicated `KiB` in one of the values
4. Added `formatLedgersToTime` helper function

**After merging this, add a link to this page on https://developers.stellar.org/docs/networks/resource-limits-fees**

### DIFF
--- a/src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
+++ b/src/app/(sidebar)/smart-contracts/contract-explorer/components/InvokeContractForm.tsx
@@ -28,6 +28,7 @@ import { ErrorText } from "@/components/ErrorText";
 import { JsonSchemaRenderer } from "@/components/SmartContractJsonSchema/JsonSchemaRenderer";
 import { TransactionSuccessCard } from "@/components/TransactionSuccessCard";
 import { WalletKitContext } from "@/components/WalletKit/WalletKitContextProvider";
+import { TabView } from "@/components/TabView";
 import { CodeEditor, SupportedLanguage } from "@/components/CodeEditor";
 
 import { TransactionBuildParams } from "@/store/createStore";
@@ -587,21 +588,42 @@ export const InvokeContractForm = ({
   const renderResponse = () => {
     if (jsonResponse?.fullResponse || base64Response?.fullResponse) {
       return (
-        <SimulatedResponse
-          jsonResult={jsonResponse}
-          base64Result={base64Response}
-          isFullResponseEnabled={isFullResponseEnabled}
-          xdrFormat={xdrFormat}
-          onXdrFormatChange={(format: string) => {
-            setXdrFormat(format);
+        <TabView
+          tab1={{
+            id: "json",
+            label: "JSON",
+            content: jsonResponse?.fullResponse && (
+              <SimulatedResponse
+                xdrFormat="json"
+                result={jsonResponse}
+                isFullResponseEnabled={isFullResponseEnabled}
+              />
+            ),
+            isDisabled: !jsonResponse?.fullResponse,
+          }}
+          tab2={{
+            id: "base64",
+            label: "Base64",
+            content: base64Response?.fullResponse && (
+              <SimulatedResponse
+                xdrFormat="xdr"
+                result={base64Response}
+                isFullResponseEnabled={isFullResponseEnabled}
+              />
+            ),
+            isDisabled: !base64Response?.fullResponse,
+          }}
+          activeTabId={xdrFormat}
+          onTabChange={(id) => {
+            setXdrFormat(id);
             trackEvent(
               TrackingEvent.SMART_CONTRACTS_EXPLORER_INVOKE_CONTRACT_SELECTED_XDR_FORMAT,
               {
-                xdrFormat: format,
+                xdrFormat: id,
               },
             );
           }}
-          isFullResponseToggle={
+          rightElement={
             <FullResponseToggle
               isChecked={isFullResponseEnabled}
               onChange={setIsFullResponseEnabled}
@@ -756,23 +778,17 @@ export const FullResponseToggle = ({
 };
 
 export const SimulatedResponse = ({
-  jsonResult,
-  base64Result,
+  result,
   isFullResponseEnabled = false,
   xdrFormat,
-  onXdrFormatChange,
-  isFullResponseToggle,
 }: {
-  jsonResult: SimulatedResponseType | null;
-  base64Result: SimulatedResponseType | null;
+  result: SimulatedResponseType | null;
   isFullResponseEnabled: boolean;
   xdrFormat: string;
-  onXdrFormatChange: (format: string) => void;
-  isFullResponseToggle?: React.ReactNode;
 }) => {
-  const result = xdrFormat === "json" ? jsonResult : base64Result;
-
-  if (!result) return null;
+  if (!result) {
+    return null;
+  }
 
   const hasSimulationError = Api.isSimulationError(result.fullResponse);
 
@@ -789,18 +805,11 @@ export const SimulatedResponse = ({
     <Box gap="md">
       <div data-testid="invoke-contract-simulate-tx-response">
         <CodeEditor
-          isAutoHeight
-          maxHeightInRem="30"
-          title="Simulation Response"
+          key={`code-editor-${isFullResponseEnabled}`}
+          isAutoHeight={!isFullResponseEnabled}
+          heightInRem={`${isFullResponseEnabled ? "50" : "30"}`}
           value={JSON.stringify(json, null, 2)}
           selectedLanguage={selectedLanguage}
-          languages={["json", "xdr"]}
-          onLanguageChange={(newLanguage) => {
-            // Map back to the original format names
-            const format = newLanguage === "xdr" ? "base64" : newLanguage;
-            onXdrFormatChange(format);
-          }}
-          customEl={isFullResponseToggle}
         />
       </div>
     </Box>


scripts/fetch-network-limits.mjs
```diff
@@ -142,6 +146,74 @@ const getNetworkLimitsFromResponse = (response) => {
...
+    fee_read_ledger_entry:
+      ledgerCostEntry?.dataJson?.config_setting?.contract_ledger_cost_v0
+        ?.fee_disk_read_ledger_entry,
+    fee_write_ledger_entry:
+      ledgerCostEntry?.dataJson?.config_setting?.contract_ledger_cost_v0
+        ?.fee_write_ledger_entry,
+    fee_read_1kb:
+      ledgerCostEntry?.dataJson?.config_setting?.contract_ledger_cost_v0
+        ?.fee_disk_read1_kb,
+    fee_write_1kb: undefined, // @TODO will be filled in later
```
jeesunikim: @dmkozh I was able to get `fee_write_1kb` via `contract_ledger_cost_ext_v0`!!

I am still unsure where to get `(462'000/max tx)` that [the doc](https://developers.stellar.org/docs/networks/resource-limits-fees) displays: `3,500 (462'000/max tx)`

src/app/(sidebar)/network-limits/page.tsx
```diff
@@ -0,0 +1,400 @@
...
+    },
+    {
+      setting: "Transaction size",
+      perTransaction: formatBytes(limits.tx_max_write_bytes),
+      ledgerWide: formatBytes(limits.ledger_max_txs_size_bytes),
+    },
+    {
+      setting: "Events + return value size",
+      perTransaction: formatBytes(limits.tx_max_contract_events_size_bytes),
+      ledgerWide: "no explicit limit",
```
jeesunikim, line 137: can this be hard coded `no explicit limit"`? @dmkozh

src/app/(sidebar)/network-limits/page.tsx
```diff
@@ -0,0 +1,400 @@
...
+    },
+    {
+      setting: "Individual ledger key size (contract storage key)",
+      perTransaction: `${limits.contract_data_key_size_bytes} bytes`,
+      ledgerWide: undefined,
+    },
+    {
+      setting: "Individual ledger entry size (including Wasm entries)",
+      perTransaction: formatBytes(limits.contract_max_size_bytes, "binary"),
```
jeesunikim, line 148: @quietbits , the doc uses `KiB` so I updated the `formatBytes` and `formatFiles` to reflect `KiB` usage (1000 vs. 1024)

src/helpers/formatFileSize.ts
```diff
@@ -1,21 +1,30 @@
-export const formatFileSize = (bytes: number) => {
-  if (bytes < 1024) {
+export const formatFileSize = (
+  bytes: number,
+  unit: "binary" | "decimal" = "decimal",
```
jeesunikim, line 3: added an unit param to reflect `KiB` (what [doc](https://developers.stellar.org/docs/networks/resource-limits-fees) uses for `Individual ledger entry size (including Wasm entries)`)

src/app/(sidebar)/network-limits/page.tsx
```diff
@@ -0,0 +1,400 @@
...
+      setting: "1 KB of transaction size (history)",
+      value: `${formatNumber(Number(limits.fee_historical_1kb))} (${maxHistoricalFee}/max tx)`,
+    },
+    {
+      setting: "1 KB of Events/return value",
+      value: `${formatNumber(Number(limits.fee_contract_events_1kb))} (${maxEventsFee}/max tx)`,
+    },
+    {
+      setting: "30 days of rent for 1 KB of persistent storage",
+      value: `~${getDaysOfRent(limits.fee_write_1kb, limits.persistent_rent_rate_denominator, 30)}`,
```
jeesunikim: @dmkozh, what's the right formula to calculate `30 days of rent for 1 KB of persistent storage`? I asked claude and it suggested to add a hardcoded value or using the following:

```
const getDaysOfRent = (
  write_fee_1kb: number,
  rent_denominator: string,
  days: number,
) => {
  const ledgers_per_day = 86400 / 5; // 1 ledger every 5 seconds
  const rent =
    write_fee_1kb * (ledgers_per_day / Number(rent_denominator)) * days;
  return formatNumber(rent);
};
```

I don't think it's correct when I compare it to the doc's value

Image

src/app/(sidebar)/network-limits/page.tsx
```diff
@@ -0,0 +1,400 @@
...
+      setting: "Read 1 KB from disk",
+      value: `${formatNumber(Number(limits.fee_read_1kb))} (${maxReadBytesFee}/max tx)`,
+    },
+    {
+      setting: "Write 1 ledger entry",
+      value: `${formatNumber(Number(limits.fee_write_ledger_entry))} (${maxWriteEntriesFee}/max tx)`,
+    },
+    {
+      setting: "Write 1KB to disk",
+      value: `${formatNumber(Number(limits.fee_write_1kb))} (${maxWriteEntriesFee}/max tx)`, // @TODO maxWriteEntriesFee
```
jeesunikim: @dmkozh I was able to get `fee_write_1kb` via `contract_ledger_cost_ext_v0`!!

I am still unsure where to get `(462'000/max tx)` that [the doc](https://developers.stellar.org/docs/networks/resource-limits-fees) displays: `3,500 (462'000/max tx)`

dmkozh: The doc may do some rounding, and the numbers generally can be off (which is why we want this to be computed programmatically!). But every `max tx` number comes from computing the fee for the respective resource with the maxed-out transaction limit. So for e.g. `Write 1KB to disk` max tx cost we do `132 KB * (3500 stroops / 1 KB) = 462000 stroops`. Note, that 'KB' is counted as `1024` bytes for the purpose of these computations.

src/app/(sidebar)/network-limits/page.tsx
dmkozh: FWIW the canonical implementation of computing every fee component can be found here: https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/fees.rs, it might be especially useful if converted to a different language with AI.

That said, for the rent cost we need to find the rent write fee first, then plug it into computation similar to what you have, but using the proper rent denominator. Getting rent write fee might be a bit tricky though -what are the data sources that we have access to? Just the ledger?

dmkozh: Also, to be clear, the fee for writing 1 KB is 3500 and comes from `contract_ledger_cost_ext_v0`, but it's not the same as the rent write fee.

jeesunikim: >Getting rent write fee might be a bit tricky though -what are the data sources that we have access to? Just the ledger?

Yes, just the ledger. It returns `persistent_rent_rate_denominator` and `temp_rent_rate_denominator`. Do you recommend I hard code it for now? `rent write fee` is around `1001` if I want to match it to `427,000` @dmkozh

dmkozh: No, I don't recommend hardcoding this. What we should do is the following:

- get `CONFIG_SETTING_LIVE_SOROBAN_STATE_SIZE_WINDOW` config entry (https://github.com/stellar/stellar-xdr/blob/0a621ec7811db000a60efae5b35f78dee3aa2533/Stellar-contract-config-setting.x#L397)
- find the average soroban live state size as simply an average of all the values in the window
- also fetch the rent-related config params. Here is a good example of the mapping: https://github.com/stellar/rs-soroban-env/blob/97546e04908701935dcf81d36bb0a69ac761430e/soroban-simulation/src/network_config.rs#L101
- translate `compute_rent_write_fee_per_1kb` to TS: https://github.com/stellar/rs-soroban-env/blob/97546e04908701935dcf81d36bb0a69ac761430e/soroban-env-host/src/fees.rs#L245. It's a bit involved, but that seems like a great use case for AI translation.

It's a bit more work, but will save us time maintaining this in the future.

dmkozh: FWIW the expected value for this is 1000 both for testnet and mainnet and it's achieved due to minimum rent write fee being 1000. So even if you end up with 1000 I would recommend trying some test vectors from the Rust tests: https://github.com/stellar/rs-soroban-env/blob/97546e04908701935dcf81d36bb0a69ac761430e/soroban-env-host/tests/fees.rs#L980

jeesunikim: Thank you! I updated it to pull it dynamically

src/app/(sidebar)/network-limits/page.tsx
dmkozh: Yeah, even if we ever change this, it would be a rather significant protocol change, so it would be hard to miss updating this.

src/constants/networkLimits.ts
```diff
@@ -69,7 +76,43 @@ export const MAINNET_LIMITS: NetworkLimits = {
   "fee_historical_1kb": "16235",
   "fee_contract_events_1kb": "10000",
   "persistent_rent_rate_denominator": "1215",
-  "temp_rent_rate_denominator": "2430"
+  "temp_rent_rate_denominator": "2430",
+  "live_soroban_state_size_window": [
```
dmkozh: Is this some kind of fallback value? Unlike the rest of fields, this one is subject to frequent and dynamic changes, though the rent fees would only be impacted when the size is above a certain limit.

jeesunikim: that's the value we got from [scripts/fetch-network-limits.mjs‎](https://github.com/stellar/laboratory/pull/1794/changes#diff-e9834fc7d041335182e55505454c08e271b27df08879803e329e16bc4996101e) when it got triggered (which gets triggered when the lab gets deployed)

dmkozh: I see, so is this going to be refreshed every time lab gets deployed? How often would this be updated then?

jeesunikim: Likely at least once a day (during weekdays). After reading the conversation on [slack](https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1764797388378429?thread_ts=1764787539.254209&cid=C06KTGUULUF), I'll create a task to explore cronjob/CI action.

> What we do for some generated files (like XDR parsers) is that we run a CI action that generates them and compares to the value in the repo to the value generated during the CI build. so if the value is updated, you need to regenerate the file locally and merge it

dmkozh: Alright, that's mostly fine, but we should probably add a note somewhere on the page that the results may be slightly stale.

src/app/(sidebar)/network-limits/page.tsx
```diff
@@ -0,0 +1,628 @@
...
+                  {item.ledgerWide ?? null}
+                </GridTableCell>
+              </GridTableRow>
+            ))}
+          </>
+        </GridTable>
+      </div>
+
+      {/* Mobile: Per-transaction table */}
+      <Box gap="lg" addlClassName="NetworkLimits__mobileOnly">
```
quietbits: We shouldn't show both desktop and mobile markup in the source, because it's duplicate data (unnecessary code that can seem spammy). We should use `useWindowSize()` instead and render only the view that is needed. We can go by a specific window size or layout mode. It's OK to refresh the page in this case, since users typically don't switch from mobile to desktop. We already did this for the main navigation.

src/app/(sidebar)/network-limits/page.tsx
```diff
@@ -0,0 +1,628 @@
...
+      role={role}
+    >
+      {children}
+    </div>
+  );
+};
+
+// // =============================================================================
+// // Helpers
+// // =============================================================================
```
quietbits:
```suggestion
// =============================================================================
// Helpers
// =============================================================================
```

src/app/(sidebar)/network-limits/styles.scss
```diff
@@ -0,0 +1,84 @@
+@use "../../../styles/utils.scss" as *;
+
+.NetworkLimits {
+  &__desktopOnly {
+    display: block;
+
+    @media (max-width: 420px) {
+      display: none;
+    }
```
quietbits: We shouldn't handle this with CSS as per my comment above.

```suggestion
    @media (min-width: 420px) {
      display: none;
    }
```

PR #1806, opened 2026-01-14
[Build Transaction] Adjust dropdown width size for saved keypair

jeesunikim:
- Prioritized this since jayrome is making a video on build tx and spotted this being too long

Screenshot 2026-01-13 at 4 15 43 PM

### DIFF
--- a/src/helpers/truncateString.ts
+++ b/src/helpers/truncateString.ts
@@ -2,5 +2,5 @@ export const truncateString = (text: string, maxLength: number) => {
   if (text.length <= maxLength) {
     return text;
   }
-  return text.slice(0, maxLength) + "...";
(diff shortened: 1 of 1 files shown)


src/helpers/truncateString.ts
```diff
@@ -0,0 +1,6 @@
+export const truncateString = (text: string, maxLength: number) => {
+  if (text.length <= maxLength) {
+    return text;
+  }
+  return text.slice(0, maxLength) + "...";
```
quietbits: Use the ellipsis character instead of 3 dots.
```suggestion
  return text.slice(0, maxLength) + "…";
```

quietbits (approved):

PR #1821, opened 2026-01-21
Transaction Dashboard: Call stack trace tab

quietbits:
- Call stack trace view for transaction diagnostic events.
  - You can test by using any recent transaction ID on any network, then use the test sections to select the data source to view. This is for testing only and will be removed before the PR is merged.

image

- Reusing some styles and pieces from `PrettyJson` because the use case here is slightly different (everything needs to work inline).
- Params collapsing: we’re using the max params count (4), which is the only performant way to handle it. We can’t dynamically set them to fit on a single line because that would create too many size listeners on the page, making everything feel sluggish. This is not ideal, because for long params they would still wrap to the second line, but this is a compromise we need to make to keep the page snappy.

image

Tests will be added in https://github.com/stellar/laboratory/issues/1857


src/helpers/formatDiagnosticEvents.ts
```diff
@@ -0,0 +1,400 @@
...
+  | { type: "error"; value: ScError }
+  | { type: "void"; value: null };
+
+export type FormattedEventData =
+  | FormattedPrimitive
+  | { type: "vec"; value: FormattedEventData[] }
+  | { type: "map"; value: FormattedMapEntry[] }
+  | { type: "literal"; value: string }
+  | { type: "ellipsis"; value: string }
+  | { type: undefined; value: unknown };
```
jeesunikim, line 69: we can have an undefined formatted event data?

src/app/(sidebar)/transaction-dashboard/styles.scss
```diff
@@ -543,3 +543,313 @@
...
+// =============================================================================
+.CallStackTrace {
+  border: 1px solid var(--sds-clr-gray-06);
+  padding: pxToRem(24px);
+  border-radius: pxToRem(8px);
+
+  font-family: var(--sds-ff-monospace);
+  font-size: pxToRem(14px);
+  line-height: pxToRem(18px);
+  letter-spacing: -0.28px;
```
jeesunikim, line 558: `pxToRem`?

src/helpers/formatDiagnosticEvents.ts
quietbits: It's used in `formatEventData` function when we can't find the type. Should we set a different value? We can't make it optional, because the `type` prop is required.

```
// line 245
return { value: data, type: undefined };
```

src/app/(sidebar)/transaction-dashboard/styles.scss
quietbits: We don't use `pxToRem` for values 1px or smaller. It would produce a very tiny value, which might not be displayed correctly.

jeesunikim:
> Do auth signatures also appear here, or on another tab when an auth signature (not tx signature) is present?
> 
> Screenshot 2026-01-27 at 11 03 12 am

Leigh we do not show auth signatures in Signatures tab. I added a task to add this. https://github.com/stellar/laboratory/issues/1848

src/helpers/formatDiagnosticEvents.ts
```diff
@@ -0,0 +1,400 @@
...
+      coreMetrics: [],
+      errorLevel: undefined,
+    };
+  }
+
+  const { callStack, coreMetrics } = processEvents(dEvents);
+
+  return {
+    callStack,
+    coreMetrics,
```
jeesunikim: `coreMetrics` isn't being used in `CallStackTrace`. Does this need to be returned?

src/helpers/formatDiagnosticEvents.ts
```diff
@@ -0,0 +1,400 @@
...
+  // Early return for empty input
+  if (!dEvents || dEvents.length === 0) {
+    return {
+      callStack: [],
+      coreMetrics: [],
+      errorLevel: undefined,
+    };
+  }
+
+  const { callStack, coreMetrics } = processEvents(dEvents);
```
jeesunikim: same here with `coreMetrics`. Where does this being used?

src/helpers/formatDiagnosticEvents.ts
quietbits: Core metrics were in the initial design, so I kept them there in case we decide to use them later.

src/helpers/formatDiagnosticEvents.ts
quietbits: We don't have them in the latest design, but I thought it might be OK to keep them formatted in case we need them later.

src/helpers/formatDiagnosticEvents.ts
jeesunikim: Can we comment them out

src/helpers/formatDiagnosticEvents.ts
```diff
@@ -0,0 +1,401 @@
...
+const getEventName = (event: DiagnosticEvent): string => {
+  const topics = event.body?.v0?.topics;
+  if (!topics || topics.length === 0) return "";
+
+  const topicSymbol = topics[0]?.symbol;
+
+  if (event.type_ === EVENT_TYPES.DIAGNOSTIC) {
+    if (topicSymbol === EVENT_TOPIC_SYMBOLS.FN_RETURN) {
+      // Use the function name (2nd topic)
+      return topics?.[1]?.symbol || "";
```
jeesunikim, line 150: Do you know if this might change in the future? raising b/c we relies on the order of topics for `name`

quietbits: If it changes, we'll need to update this as well. Unfortunately, there is no other way to parse this data, so we need to make some assumptions.

jeesunikim (approved):

PR #1842, opened 2026-01-30
Add Stella Icon

jeesunikim:


src/components/CustomAiButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+      }}
+    >
+      <Image
+        src="/images/stella-ai-button.png"
+        alt="AI Assistant"
+        width={100}
+        height={32}
+        priority
+      />
+    </button>
```
quietbits: I'm not sure about the limitations we have, so some questions:
1. Do we have to use inline styles, or could we use a class with style overrides?
2. Are we limited to using an image, or could we style it in code?
3. It would be nice to adjust the button to the Lab theme selected (light or dark). The dark button is barely visible in the dark mode.

Image

quietbits (commented):
When the network/browser blocks the AI button code, the button doesn't show up in the current version (Lab production), but this one is always there. So, when I click on it when it's blocked, nothing happens. Could we keep the same logic and show it only when it's available?

src/components/CustomAiButton.tsx
jeesunikim: excellent points. I will apply these recommendations.

src/components/CustomAiButton/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <path d="M952.992 413.731l33.008-16.431v-73.834l-947.999 475.022v73.833l804.008-402.251 0.651 4.48c1.954 14.295 2.822 28.811 2.822 43.107 0 87.491-34.751 170.075-97.733 231.961-30.84 30.305-66.89 53.991-107.069 70.635-40.181 16.643-82.75 25.178-126.402 25.394h-2.39c-56.902 0-112.936-14.295-162.017-41.188l-2.608 1.279-66.457 33.291c23.456 16.218 48.648 30.089 75.145 41.188 49.52 20.485 101.86 30.943 155.937 30.943s106.639-10.453 155.94-30.943c47.563-19.845 90.564-48.226 127.268-84.503 36.701-36.279 65.587-78.316 85.788-125.264 20.848-48.652 31.492-100.297 31.492-153.431 0-10.029-0.218-20.273-1.084-30.302-2.39-30.514 14.115-59.112 41.699-72.983z" />
+      </svg>
```
quietbits: Is this icon different than `Logo.StellarShort` in SDS?

quietbits (approved):

PR #1890, opened 2026-02-17
Transaction dashboard: call stack trace render error obj

quietbits:
- Correctly render object in error params
- Don't truncate string that is not a valid asset 

Before
image

After
image

### DIFF
--- a/tests/e2e/txDashResourceProfiler.test.ts
+++ b/tests/e2e/txDashResourceProfiler.test.ts
@@ -117,6 +117,13 @@ const txData = async ({
 
   await loadButton.click();
 
+  // Verify Call stack trace tab is active by default
+  const callStackTraceTabButton = page.getByTestId("tx-call-stack-trace");
+  await expect(callStackTraceTabButton).toHaveAttribute(
+    "data-is-active",
+    "true",
+  );
+
   // Select tab
   const resourceProfilerTabButton = page.getByTestId("tx-resource-profiler");
   await expect(resourceProfilerTabButton).toHaveAttribute(


src/constants/networkLimits.ts
```diff
@@ -78,36 +78,36 @@ export const MAINNET_LIMITS: NetworkLimits = {
...
+    "733489487",
+    "734432187",
+    "734927731",
+    "735335063",
+    "735793159",
+    "736644795",
+    "737385667",
+    "737622323",
+    "738247343"
   ],
```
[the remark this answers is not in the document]
quietbits: This file is generated during build process when developing. We need to include this file to update to the latest network data in production. It's OK that it's not related to this PR, we're used to this by now.

src/app/(sidebar)/transaction/dashboard/components/CallStackTrace.tsx
```diff
@@ -437,7 +491,23 @@ export const CallStackTrace = ({
...
+    return false;
+  }
+
+  const [code, issuer] = parts;
+  // Asset code should be 1-12 characters, issuer must be valid G or M address
+  return (
+    code.length > 0 &&
+    code.length <= 12 &&
+    (StrKey.isValidEd25519PublicKey(issuer) ||
+      StrKey.isValidMed25519PublicKey(issuer))
```
jeesunikim: can we double check whether we should enable contract id in this conditional?

jeesunikim (approved):

src/app/(sidebar)/transaction/dashboard/components/CallStackTrace.tsx
quietbits: Sure, we can add it here. It definitely won't hurt. :+1:

PR #1893, opened 2026-02-18
 [Contract Storage] Replace SE with our own backend API

jeesunikim:
**Summary:**
- Fallback to Stellar Expert's
- Use our own [backend](https://github.com/stellar/laboratory-backend) to fetch
- removed `First` and `Last` button when we use our own backend. Sorting by `updated_at` in `asc` mode returns the last items. They'll appear for the fallback SE.
- Added tracking

### DIFF
--- a/src/app/(sidebar)/transaction/dashboard/components/Events.tsx
+++ b/src/app/(sidebar)/transaction/dashboard/components/Events.tsx
@@ -9,6 +9,7 @@ import { ScValPrettyJson } from "@/components/StellarDataRenderer";
 import { ExpandBox } from "@/components/ExpandBox";
 import { CopyJsonPayloadButton } from "@/components/CopyJsonPayloadButton";
 import { TransactionTabEmptyMessage } from "@/components/TransactionTabEmptyMessage";
+import { CodeEditor } from "@/components/CodeEditor";
 
 import { shortenStellarAddress } from "@/helpers/shortenStellarAddress";
 import { FormattedTxEvent, formatTxEvents } from "@/helpers/formatTxEvents";
@@ -223,9 +224,11 @@ const EventJson = ({ json }: { json: AnyObject }) => {
 
       <ExpandBox offsetTop="md" isExpanded={isExpanded}>
         <Box gap="md">
-          <div className="TransactionEvents__card TransactionEvents__card--inset">
-            <pre>{formattedJson}</pre>
-          </div>
+          <CodeEditor
+            value={formattedJson}
+            selectedLanguage="json"
+            isAutoHeight
+          />
           <Box gap="sm" direction="row" justify="end">
             <CopyJsonPayloadButton size="sm" jsonString={formattedJson} />
           </Box>
--- a/src/app/(sidebar)/transaction/dashboard/styles.scss
+++ b/src/app/(sidebar)/transaction/dashboard/styles.scss
@@ -356,23 +356,6 @@
     padding: pxToRem(12px);
     gap: pxToRem(12px);
 
-    &--inset {
-      padding: pxToRem(8px);
-      background-color: var(--sds-clr-gray-03);
-      border: none;
-
-      pre {
-        padding-left: pxToRem(8px);
-        padding-right: pxToRem(8px);
-        margin: 0;
-        font-family: var(--sds-ff-monospace), monospace;
-        font-size: pxToRem(14px);
-        line-height: pxToRem(20px);
-        color: var(--sds-clr-gray-12);
-        overflow: auto;
-      }
-    }
-
     &--data {
       gap: pxToRem(4px);
     }


src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
```diff
@@ -46,18 +57,82 @@ export const ContractStorage = ({
...
+    isFetching: isStorageFetching,
+  } = activeResult;
+
+  // Normalize data from both sources
+  const storageData = (() => {
+    if (isSourceStellarExpert) {
+      const data = seResult.data;
+      // SE API doesn't support sort_by, so sort by updated descending client-side
+      return data
+        ? [...data].sort((a, b) => (b.updated || 0) - (a.updated || 0))
```
jeesunikim: I thought we were showing the latest first previously. Was I mistaken? I added this to show the latest first @quietbits

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractInfo.tsx
```diff
@@ -44,13 +44,17 @@ import { Bindings } from "./Bindings";
 
 export const ContractInfo = ({
   infoData,
+  contractId,
```
jeesunikim, line 47: added this to rely less on `infoData`

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
quietbits: That is correct. The order is set when we fetch all the data in the query [here](https://github.com/stellar/laboratory/blob/main/src/query/external/useSEContracStorage.ts#L43). By default the data order is descending, so we shouldn't need to sort here again.

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
```diff
@@ -233,10 +311,62 @@ export const ContractStorage = ({
...
+                      isStorageFetching,
+                  },
+                  next: {
+                    onClick: () => {
+                      if (nextCursor) {
+                        const queryString = nextCursor.split("?")[1] || "";
+                        const params = new URLSearchParams(queryString);
+                        setCurrentCursor(params.get("cursor") || undefined);
+                        setCurrentPage(currentPage + 1);
+                      }
```
quietbits, line 351: We do the same thing for both prev and next, we could make this a shared helper method.

src/query/external/useBackendContractStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+      cursor,
+      sortBy,
+      order,
+    ],
+    queryFn: async () => {
+      const network = networkId === "mainnet" ? "pubnet" : networkId;
+      const baseUrl = `${BACKEND_ENDPOINT}/${network}/api/contract/${contractId}/storage?limit=10${sortBy ? `&sort_by=${sortBy}` : ""}${order ? `&order=${order}` : ""}`;
+      const fetchUrl = cursor
+        ? `${baseUrl}&cursor=${encodeURIComponent(cursor)}`
+        : baseUrl;
```
quietbits: To make the params easier to read, we could use `URLSearchParams` ([how we handle the SE fetch](https://github.com/stellar/laboratory/blob/main/src/query/external/useSEContracStorage.ts#L41:L52)).

quietbits (commented):
Works great! I think we can try to tweak TanStack Query to keep the previous data until the new data is loaded. Right now, we show loader when switching to a new page or using sort features, which makes the page content jump around.
Also, when going back to a page that I just fetched, we shouldn't need to fetch it again. This is also something we can control in TanStack Query.

src/app/(sidebar)/smart-contracts/contract-explorer/components/ContractStorage.tsx
jeesunikim: not sure if this is the case anymore. If you go to [the live lab](https://lab.stellar.org/smart-contracts/contract-explorer?$=network$id=testnet&label=Testnet&horizonUrl=https:////horizon-testnet.stellar.org&rpcUrl=https:////soroban-testnet.stellar.org&passphrase=Test%20SDF%20Network%20/;%20September%202015;&smartContracts$explorer$contractId=CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC;;) for XLM SAC, its storage shows its genesis (earliest) first by default.

quietbits: It looks like `order` flag is ignored in the [API request](https://api.stellar.expert/explorer/testnet/contract-data/CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC?order=desc&limit=10). It might be better to fix this at the fetch time, as the UI already expects the data to be returned in descending order.

jeesunikim: updated

PR #1900, opened 2026-02-23
add a design doc for new transaction flow

jeesunikim:
Readable view: https://github.com/stellar/laboratory/blob/df1be05fb9d42d09e66076050ce43e6c72bf538b/docs/new-transaction-flow.md


docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1043 @@
...
+- **"Next" in the footer is the only way to advance to a step that has not yet
+  been completed.** It advances to the next step in the applicable list and
+  updates `highestCompletedStep`
+- "Back" sets `activeStep` to previous step in the list (content is preserved in
+  store)
+- Clicking a step in the stepper only works for **already-completed steps** —
+  users can navigate back to any completed step, but cannot skip forward.
+  Forward navigation always goes through the "Next" footer button
+
+### Shared link handling (MVP)
```
jeesunikim: Shared link handling @janewang

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1043 @@
...
+
+This page now owns the Build flow end-to-end. `activeStep` lives in the Zustand
+transaction store and is synced to the `?step=` query param via
+`zustand-querystring`. `highestCompletedStep` is local state (session-only, not
+URL-persisted).
+
+New structure — `page.tsx` renders `<BuildFlow />` directly; the `<Tabs />`
+component contains nav links, not local state:
+
+> **Tabs component concern:** The existing `src/components/Tabs/index.tsx` uses
```
jeesunikim, line 404: @quietbits what do you think of this approach?

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+
+- **`/transaction/build`** — New transaction (Build flow)
+- **`/transaction/import`** — Import transaction XDR (Import flow)
+
+The tabs at the top are navigation links (`<Link href="/transaction/import">`);
+switching tabs is a page navigation, not local state. Within each page, all
+steps are rendered on that single URL. The stepper tracks which step is active
+and users can navigate back to any previously completed step. Step content is
+conditionally rendered based on the active step — the `?step=` URL param updates
+but there is no page navigation or reload between steps within a flow.
```
quietbits: Since we're trying to move away from storing params in URL, should we save `step` in session storage?

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+### Step 4: Create `useTransactionFlow` hook
+
+**New file:** `src/hooks/useTransactionFlow.ts`
+
+Shared hook used by both `BuildFlow` and `ImportFlow` to eliminate duplicated
+step navigation logic.
+
+### Step 5: Refactor main `page.tsx` as single-page flow
+
+**Modify:** `src/app/(sidebar)/transaction/build/page.tsx`
```
quietbits, line 395: We could keep the old flow under another route until we finish the new flow. It might be good to keep it around for testing.

docs/new-transaction-flow.md
quietbits: That's good with me 👍

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+```typescript
+// BuildFlow.tsx
+const { activeStep, setActiveStep } = useStore(transactionStore);
+const isSoroban = /* derived from transaction type in build store */;
+const hasAuthEntries = /* derived from simulation result in store */;
+const steps: TransactionStepName[] = isSoroban
+  ? hasAuthEntries
+    ? ["build", "simulate", "sign", "validate", "submit"]
+    : ["build", "simulate", "sign", "submit"]
+  : ["build", "sign", "submit"];
```
quietbits: If ESLint complains about double nesting, we can make this a function.

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+
+The Simulate step is Soroban-only. The Validate step is Soroban-only and
+conditional — it only appears when the simulation returns auth entries that
+require signing. The steps array updates dynamically after simulation completes.
+
+| Variant                      | Build flow steps                                      | Import flow steps (replaces `"build"` with `"import"`)  |
+| ---------------------------- | ----------------------------------------------------- | ------------------------------------------------------- |
+| **Classic**                  | `["build", "sign", "submit"]`                         | `["import", "sign", "submit"]`                          |
+| **Soroban, no auth entries** | `["build", "simulate", "sign", "submit"]`             | `["import", "simulate", "sign", "submit"]`              |
+| **Soroban, auth entries**    | `["build", "simulate", "sign", "validate", "submit"]` | `["import", "simulate",  "sign", "validate", "submit"]` |
```
sagpatil, line 48: Should the order be 

> simulate → sign_auth → validate → sign_envelope → submit

 instead?

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+`/transaction/submit`) redirect to `/transaction/build?step=<step>&xdr=...`.
+
+New-flow shared links use the `step` query param directly:
+
+| URL                                        | Initial step | Previous steps                     |
+| ------------------------------------------ | ------------ | ---------------------------------- |
+| `/transaction/build?xdr=...`               | `"build"`    | None                               |
+| `/transaction/build?step=simulate&xdr=...` | `"simulate"` | Build disabled                     |
+| `/transaction/build?step=sign&xdr=...`     | `"sign"`     | Build, Simulate disabled           |
+| `/transaction/build?step=validate&xdr=...` | `"validate"` | Build, Simulate, Validate disabled |
```
sagpatil: Build, Simulate, Sign disabled

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+```
+
+- `activeStep: TransactionStepName` — controls which content renders
+- `highestCompletedStep: TransactionStepName | null` — tracks the furthest step
+  reached (`null` when entering mid-flow via shared link)
+- **Build flow and Import flow each own their own `steps` array** — they live on
+  separate routes (`/transaction/build` → `BuildFlow`; `/transaction/import` →
+  `ImportFlow`). Each `page.tsx` is self-contained.
+- Steps are derived from the step variants table above; ordering uses array
+  position: `const stepIndex = (s: TransactionStepName) => steps.indexOf(s)`
```
sagpatil, line 104: What happens when the steps array changes mid-session? 
Example: user completes Simulate (no auth entries, 4-step variant), advances to Sign, then navigates back to Simulate and re-runs it. The new simulation now returns auth entries, inserting Validate into the array. highestCompletedStep points to Sign, but the step ordering has changed — is Submit still reachable? Is the Sign step's data still valid given the new auth entries?

I'd suggest: any re-simulation resets all subsequent step state (highestCompletedStep reverts to "simulate", Sign/Validate/Submit data cleared). Worth documenting the exact rule here.

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+
+Once XDR is pasted, there are three possible states:
+
+### Valid XDR without existing signatures
+
+- Success alert at top
+- Transaction info (parameters, operations) displayed in page body
+- Stepper updates to correct variant based on `parsedTxType`
+- **Footer**: derives next-step label from `steps` array:
+  - Soroban → "Next: Simulate transaction"
```
sagpatil, line 975: For Soroban imports with existing signatures, the doc says "Next: Submit transaction" (skip to Submit, line 861). But a Soroban tx may still need simulation for resource estimation even if it has signatures. Should the skip-to-Submit behavior be Classic-only?

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+
+Auth signing lives in the Simulate step (not deferred to the Sign step) because:
+
+1. **Auth entries are produced by simulation** — they don't exist before
+   simulation runs
+2. **`assembleTransaction` must happen before Sign step** — after auth entries
+   are signed, `assembleTransaction(tx, simulationResult)` attaches the signed
+   auth + resource data to produce the final XDR. The Sign step then signs this
+   assembled envelope — a completely different operation
+3. **Short expiration window** — auth entry signatures expire after ~12–60
```
sagpatil: Since auth entries expire in ~1–5 minutes, what happens if the user takes too long on the Sign step? Consider adding an expiration warning or auto-invalidation that sends the user back to Simulate when auth entries are likely stale.

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,940 @@
...
+    ? ["build", "simulate", "sign", "validate", "submit"]
+    : ["build", "simulate", "sign", "submit"]
+  : ["build", "sign", "submit"];
+
+const { highestCompletedStep, stepIndex, handleNext, handleBack, handleStepClick } =
+  useTransactionFlow({
+    steps,
+    activeStep,
+    setActiveStep,
+    initialStep: activeStep === "build" ? "build" : null,
```
sagpatil: If activeStep comes from the URL param (Zustand + zustand-querystring) Line 270:  and the user refreshes mid-flow, highestCompletedStep resets to null since it's local state. This means a refresh behaves identically to a shared link (all previous steps disabled)

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+- **"Next" in the footer is the only way to advance to a step that has not yet
+  been completed.** It advances to the next step in the applicable list and
+  updates `highestCompletedStep`
+- "Back" sets `activeStep` to previous step in the list (content is preserved in
+  store)
+- Clicking a step in the stepper only works for **already-completed steps** —
+  users can navigate back to any completed step, but cannot skip forward.
+  Forward navigation always goes through the "Next" footer button
+
+### State Persistence Model
```
jeesunikim, line 120: @quietbits added this!

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+users paste XDR directly. The existing `/transaction/sign` URL-based import path
+also remains available as a legacy option.
+
+**What persists where:**
+
+| Storage          | Scope                                                     | What it holds                                                                                           |
+| ---------------- | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------- |
+| `sessionStorage` | Current browser tab; survives refresh and back navigation | Full flow state: `activeStep`, `highestCompletedStep`, build params, XDR, simulation result, signed XDR |
+| `localStorage`   | Cross-session; explicit user action                       | Saved transactions (via "Save transaction" button)                                                      |
+| URL (querystring)| App-wide, shareable                                       | Network settings only (network ID, Horizon URL, RPC URL, passphrase) — managed by the main store via `zustand-querystring` |
```
jeesunikim: keeping the network settings in the `url` since that's useful!

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+- **`beforeunload` warning**: When the flow store has unsaved state (i.e., the
+  user has progressed past the initial step or modified build params), closing
+  or navigating away from the tab triggers a browser confirmation dialog. This
+  protects against accidental tab close (Cmd+W), browser crash recovery prompts,
+  and mobile tab kills. The warning is suppressed after "Clear all" or when the
+  flow is in its initial empty state.
+
+**Implementation — Separate Zustand Store:**
+
+Create a new `TransactionFlowStore`, completely separate from the main store.
```
jeesunikim, line 179: create a separate store for the new tx

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+  `xdr.SorobanAuthorizationEntry.fromXDR(str, 'base64')`)
+- **Hydration**: Use `skipHydration: true` with `rehydrate()` in a client-side
+  `useEffect` (sessionStorage is not available during SSR)
+- **`highestCompletedStep`** is part of the persisted store (not
+  component-local state), so browser back navigation correctly restores which
+  steps are accessible in the stepper
+
+**Error handling:**
+
+All sessionStorage access must be wrapped in try-catch. The existing codebase
```
jeesunikim, line 241: good suggestion by claude

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+
+**Lifecycle:**
+
+- State is **preserved** on page refresh and browser back/forward navigation
+  within the same session
+- State is **cleared** when the tab is closed (sessionStorage is tab-scoped)
+- State is **cleared** by "Clear all" (explicit user action)
+- Navigating between Build and Import tabs does **not** clear the other tab's
+  state; each flow has its own session storage key
+- **`beforeunload` warning**: When the flow store has unsaved state (i.e., the
```
jeesunikim: i liked this idea by claude code

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+
+Rather than splitting state across two persistence layers (some fields in URL,
+some in sessionStorage), **all** transaction flow state — including build params
+and operations — lives in sessionStorage. This avoids split-brain
+inconsistencies where URL-persisted build params and sessionStorage-persisted
+step state could contradict each other.
+
+**URL sharing vs. transaction sharing:** URL sharing is not supported for this
+flow. However, sharing partially-signed transactions (e.g., for multi-sig
+workflows) is a separate concern — that is handled by the Import flow, where
```
jeesunikim, line 145: should we allow sharing via url for `Import` for multi sig?

docs/new-transaction-flow.md
```diff
@@ -0,0 +1,1056 @@
...
+Flow state is stored in **`sessionStorage`**, not the URL. This is an
+intentional departure from the `zustand-querystring` pattern used elsewhere in
+the app.
+
+**Why sessionStorage instead of URL params:**
+
+The current codebase already excludes large data from the URL —
+`transaction.build.classic.xdr` and `transaction.build.soroban.xdr` are set to
+`xdr: false` in the store's `select()` config. Only build parameters, operation
+config, and form inputs are URL-persisted today (~1–3 KB).
```
quietbits: This might sound a little misleading: we do keep transaction XDR in the URL.

docs/new-transaction-flow.md
quietbits: Technically, there is nothing that prevents us from sharing URLs, I think. The main concern would be to allow sharing only valid data (for example, unsigned or signed transaction without errors). We might run into the same issue as we do now when the URLs are too long for browsers to process, though.

Sharing for multisig sounds like a must have.

docs/new-transaction-flow.md
quietbits: We need to keep them in the URL at least for now, because other pages need it. Eventually, we could also move it to sessionStorage.

docs/new-transaction-flow.md
quietbits: How do we envision saving states? For example, for form inputs they should be saved on blur, because we don't want to lose that information on page refresh. What other states do we need to handle?

This feature is neat, but I'm not sure how relevant it could be for the transaction flow.

docs/new-transaction-flow.md
jeesunikim: We don't save the state in url in the new flow.

docs/new-transaction-flow.md
jeesunikim: confirmed with ryan and I'm removing this sentence since the user sets the ledger sequence on their side https://github.com/stellar/js-stellar-base/blob/e3d6fc3351e7d242b374c7c6057668366364a279/src/auth.js#L50-L52

docs/new-transaction-flow.md
jeesunikim: That's a good question - I think this is probably a fee bump transaction that wraps a soroban tx, but we can clarify in the huddle

docs/new-transaction-flow.md
jeesunikim: `sign_auth` is within simulate. We validate after sign tranaction. Here's [design](https://www.figma.com/design/9j3YQ5URseAsn3j0tf2FFq/Laboratory?node-id=5766-210124&t=M73i2keTgFaGAti2-0) and [conversation with Marcelo.](https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1771877456580959)

docs/new-transaction-flow.md
jeesunikim: we are keeping the original flow. It doesn't make sense for a user to import a tx with valid signatures, have them simulate, and then re-sign the tx.

docs/new-transaction-flow.md
jeesunikim: Added this:

> After a user has gone through all steps and then they decide to change a param during the Build transaction, a confirmation dialog will warn that simulation and signing progress will be lost. If confirmed, `highestCompletedStep` is reset and all downstream step data (simulation result, signed auth entries, assembled XDR) is cleared

PR #1935, opened 2026-03-06
Add a new build folder, <TransactionFlowFooter/>, <TransactionStepper/>, and etc

jeesunikim:
- moved the existing `/transaction/build` to `/transaction/old-build` so you can disregard the changes there


src/app/(sidebar)/transaction/build/components/Params.tsx
```diff
@@ -268,120 +267,118 @@ export const Params = () => {
   };
 
   return (
-    <PageCard heading="Build transaction">
```
jeesunikim, line 271: this file looks like a lot of change, but I only removed `PageCard`

PR #1949, opened 2026-03-11
Update tests to not use a hard coded `localhost:3000`

jeesunikim:
- removing `localhost:3000` so that I can run multiple repos in parallel, make updates, and able to pass tests.


leighmcculloch (approved):
There's one comment from Copilot that's worth looking at; otherwise it looks good to me.

tests/e2e/buildTransaction.test.ts
```diff
@@ -3,7 +3,7 @@ import { mockSimulateTx } from "./mock/helpers";
 
 test.describe("Build Transaction Page", () => {
   test.beforeEach(async ({ page }) => {
-    await page.goto("http://localhost:3000/transaction/build");
+    await page.goto("/transaction/build");
```
quietbits: I think we should use absolute URL in tests. Relative navigation can be tricky and we might get more flaky tests.

tests/e2e/networkLimitsPage.test.ts
```diff
@@ -1,5 +1,7 @@
 import { test, expect, type Page } from "@playwright/test";
 
+const BASE_URL = `http://localhost:${process.env.PORT || 3000}`;
```
quietbits: To avoid writing this in every test file, we can try to make it a config or a helper (not sure where `process` would be available).

Can we use `baseURL` from Playwright config: https://github.com/stellar/laboratory/blob/main/playwright.config.ts#L12?

PR #1951, opened 2026-03-13
add `useTransactionFlow`

jeesunikim:
- added `useTransactionFlow` that determines the highest level of step taken, next/previous step.
- replaced the previous store's `transaction.` with `useBuildFlowStore`. 
- ⚠️ `transaction.updateBuildXdr` and `transaction.updateSorobanBuildXdr` are left in there since they're used for savedTransaction and contract-explorer. I added an item in the doc to remove these in launch list checklist
- Created `NewValidationResponseCard` that is same as `ValidationResponseCard` without title prop. I have a design question on this so I created a new to avoid replacing the old one with the new one in existing components

Screenshot 2026-03-13 at 12 13 17 PM

### DIFF
--- a/src/app/(sidebar)/transaction/build/components/SorobanTransactionXdr.tsx
+++ b/src/app/(sidebar)/transaction/build/components/SorobanTransactionXdr.tsx
@@ -53,7 +53,7 @@ export const SorobanTransactionXdr = () => {
           networkPassphrase={network.passphrase}
           txnHash={txnHash}
           dataTestId="build-soroban-transaction-envelope-xdr"
-          txType="smart contract"
+          txType="smart-contract"
         />
       );
     } catch (e: any) {


docs/new-transaction-flow.md
```diff
@@ -98,8 +98,8 @@ type TransactionStepName =
 - `highestCompletedStep: TransactionStepName | null` — tracks the furthest step
   reached (`null` on fresh page load with no session state)
 - **Build flow and Import flow each own their own `steps` array** — they live on
-  separate routes (`/transaction/build` → `BuildFlow`; `/transaction/import` →
-  `ImportFlow`). Each `page.tsx` is self-contained.
+  separate routes (`/transaction/build` → `BuildTransaction`;
```
jeesunikim, line 101: just updating name

docs/new-transaction-flow.md
```diff
@@ -565,6 +563,23 @@ Add events:
 - "Save transaction" opens save modal (available on Build)
 - Loading a saved transaction restores full flow state and step position
 
+## Pre-launch Checklist
+
+- [ ] Confirm `transaction.sign.importXdr` (`updateSignImportXdr`) is no longer
```
jeesunikim, line 568: added a checklist

src/app/(sidebar)/transaction/build/components/ClassicTransactionXdr.tsx
```diff
@@ -43,20 +41,22 @@ const MAX_INT64 = "9223372036854775807";
 
 export const ClassicTransactionXdr = () => {
   const { transaction, network } = useStore();
-  const router = useRouter();
   const { classic, params: txnParams, isValid } = transaction.build;
-  const { updateSignActiveView, updateSignImportXdr, updateBuildXdr } =
-    transaction;
+  const { updateBuildXdr } = transaction;
```
jeesunikim, line 45: `updateBuildXdr` is still used by `saved/page.tsx` and `contract-explorer/ContractStorage.tsx`, so the store method needs to stay at the moment but will be removed. This point is added under launch list

src/components/NewValidationResponseCard/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+ * replace ValidationResponseCard once all consumers are migrated.
+ *
+ * @example
+ * <NewValidationResponseCard
+ *   variant="success"
+ *   response={<div>XDR content here</div>}
+ *   footerRightEl={<ViewInXdrButton xdrBlob={xdr} />}
+ * />
+ */
+export const NewValidationResponseCard = ({
```
jeesunikim, line 26: It's basically the same thing as `ValidationResponseCard` but without the `title` prop. There's some design question I have for MK so I created a new one just for the new tx flow so that it doesn't disturb the existing components that use `ValidationResponseCard`

src/app/(sidebar)/transaction/build/components/SorobanTransactionXdr.tsx
```diff
@@ -2,99 +2,70 @@
...
-  try {
-    if (!sorobanXdr) {
-      return null;
+      return (
+        <TransactionXdrDisplay
+          xdr={sorobanXdr}
+          networkPassphrase={network.passphrase}
+          txnHash={txnHash}
+          dataTestId="build-soroban-transaction-envelope-xdr"
+          txType="smart contract"
```
quietbits: Do we want to have spaces in names like this?
```suggestion
          txType="smart-contract"
```

PR #1953, opened 2026-03-14
New Tx Flow:  Build Step Cleanup

jeesunikim:
- Replace old store with `TransactionFlowStore` in `<Params/>`, `<Operations/>`, `<ClassicOperations/>`, `<SorobanOperations/>`, `<ClassicTransactionXdr/>`

### DIFF
--- a/src/app/(sidebar)/transaction/build/components/Operations.tsx
+++ b/src/app/(sidebar)/transaction/build/components/Operations.tsx
@@ -951,7 +951,7 @@ export const Operations = () => {
     addDisabled = false,
     addHelperText,
   }: {
-    onAdd: () => void;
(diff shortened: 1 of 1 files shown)


src/app/(sidebar)/transaction/build/components/Operations.tsx
```diff
@@ -921,6 +944,61 @@ export const Operations = () => {
...
   };
 
+  const OperationActions = ({
+    onAdd,
+    onClear,
+    clearLabel,
+    addDisabled = false,
+    addHelperText,
+  }: {
+    onAdd: () => void;
```
quietbits: We could make this optional if it's not needed for Soroban.

PR #1956, opened 2026-03-16
Fix JSON scroll behavior

jeesunikim:


src/components/CodeEditor/index.tsx
```diff
@@ -10,6 +10,7 @@ import { WithInfoText } from "@/components/WithInfoText";
 import { downloadFile } from "@/helpers/downloadFile";
 
 import "./styles.scss";
+import { trackEvent, TrackingEvent } from "@/metrics/tracking";
```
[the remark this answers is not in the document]
quietbits: We should update imports to match other files.

src/components/CodeEditor/index.tsx
```diff
@@ -246,6 +253,7 @@ export const CodeEditor = ({
           value={value}
           onMount={handleEditorDidMount}
           options={{
+            scrollbar: { alwaysConsumeMouseWheel: false },
```
quietbits, line 256: Is this to avoid double scrollbar from the container + editor?

jeesunikim: once user lands on `<CodeEditor/>` the container scroller takes priority over the outer scroll. This can become a frustrating experience in Invoke Contract where <CodeEditor/> takes up the entire space. Once the inner container scroll reaches the end, it should enable the outer scroll for better UX.

PR #1957, opened 2026-03-17
[New Tx] Add Simulate Step

jeesunikim:
- Extracted `GridTable`, `GridTableRow`, `GridTableCell` to its own component
- Removed `prepare transaction` related code from `SorobanOperation` & `SmartContractJsonSchema` now that we have a propert simulate step
- Replaced the old state store for build tx with the new store
- Created `XdrFormat` component
- Updated Claude to reflect the naming convention (camelCase, no underscore)

Screenshot 2026-03-20 at 3 35 22 PM

I will add auth entries next once this is merged. While this is being reviewed, I will work on Sign.

### DIFF
--- a/src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
+++ b/src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
@@ -28,6 +28,7 @@ import { SdsLink } from "@/components/SdsLink";
 import { ExpandBox } from "@/components/ExpandBox";
 
 import { getNetworkHeaders } from "@/helpers/getNetworkHeaders";
+import { shortenStellarAddress } from "@/helpers/shortenStellarAddress";
 
 import { validate } from "@/validate";
 
@@ -102,22 +103,16 @@ export const SimulateStepContent = () => {
   const extractAuthEntries = useCallback(
     (responseData: SimulationResponse): string[] => {
       const authEntries: string[] = [];
+      const results = responseData?.result?.results as
+        | Array<{ auth?: string[] }>
+        | undefined;
 
-      try {
-        const result = responseData?.result;
-        const results = result?.results as
-          | Array<{ auth?: string[] }>
-          | undefined;
-
-        if (Array.isArray(results)) {
-          for (const r of results) {
-            if (Array.isArray(r.auth)) {
-              authEntries.push(...r.auth);
-            }
+      if (Array.isArray(results)) {
+        for (const r of results) {
+          if (Array.isArray(r.auth)) {
+            authEntries.push(...r.auth);
           }
         }
-      } catch {
-        // If parsing fails, return empty — no auth entries detected
       }
 
       return authEntries;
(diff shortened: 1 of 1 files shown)


src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+            if (Array.isArray(r.auth)) {
+              authEntries.push(...r.auth);
+            }
+          }
+        }
+      } catch {
+        // If parsing fails, return empty — no auth entries detected
+      }
+
+      return authEntries;
```
quietbits, line 118: In this case, is either not having auth entries (try block has no items) or error (catch block) the same? I mean, we don't need to surface the error if it happens?

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+        );
+
+        const assembled = StellarRpc.assembleTransaction(
+          transaction,
+          simResponse,
+        ).build();
+
+        setAssembledXdr(assembled.toXDR());
+      } catch (e) {
+        console.warn("Failed to assemble transaction:", e);
```
quietbits: Is this warning temporary or we don't need to throw the error?

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+        transactionXdr: builtXdr,
+        instructionLeeway: simulate.instructionLeeway,
+        headers: getNetworkHeaders(network, "rpc"),
+        xdrFormat: xdrFormat as XdrFormatType,
+        authMode: simulate.authMode,
+      });
+
+      if (result) {
+        // Store the full simulation result JSON
+        setSimulationResult(JSON.stringify(result, null, 2));
```
quietbits: We might not need to format this JSON string to save some space (without extra spaces and tabs). It will probably be formatted at the point when used.

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+                .
+              </>
+            }
+          >
+            <option value="">Select a key</option>
+
+            {[
+              { id: "record", label: "Record" },
+              { id: "enforce", label: "Enforce" },
+              { id: "record_allow_nonroot", label: "Record (allow non-root)" },
```
quietbits: Don't we usually use dash instead of underscore `record-allow-nonroot` vs `record_allow_nonroot`?

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+                          Memory bytes
+                        </Text>
+                        <Text as="div" size="xs" weight="medium">
+                          {resourceInfo.memBytes}
+                        </Text>
+                      </Box>
+                    )}
+                  </Box>
+                )}
+              </Box>
```
quietbits: We don't want to animate expand/collapse resources? We have `ExpandBox` component that handles this.

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+        {isExpanded && (
+          <Box gap="xs">
+            <Box gap="xs" direction="row" justify="space-between">
+              <Text as="div" size="xs">
+                Credentials type
+              </Text>
+              <Text as="div" size="xs" weight="medium">
+                {info.credentialsType}
+              </Text>
+            </Box>
```
quietbits: We use the same code multiple times below, it should be a component to make it more DRY and easier to read.

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -0,0 +1,656 @@
...
+                  <Button
+                    size="sm"
+                    variant="tertiary"
+                    icon={
+                      isResourcesExpanded ? (
+                        <Icon.ChevronUp />
+                      ) : (
+                        <Icon.ChevronDown />
+                      )
+                    }
```
quietbits: I don't think I saw CSS to animate the arrow. We usually animate them.

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
jeesunikim: I removed this for now. I'll add assemble part next in sign tx

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
jeesunikim: added

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
quietbits: @jeesunikim , this looks unchanged to me. What I meant is that we can create a component that could be used something like `<Item label="Credentials type" value={info.credentialsType} />`. We don't need to repeat `Box` + `Text` + `Text` for these, and it will be easier to maintain the code.

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
jeesunikim: removed `try catch` since this part is only parsing the returned data

src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
jeesunikim: updated @quietbits

Issue #1845, opened 2026-02-02
Dropdown needed for Source Account field when adding operations


minkyeongshin:
### What problem does your feature solve?
When adding an operation in Lab > Build Transaction,
if there are saved keypairs or connected wallets available,
a dropdown should be shown in the Source Account (optional) field at the bottom to allow users to select from them.

(Current)
<img width="1120" height="730" alt="Image" src="https://github.com/user-attachments/assets/3ccb8ec1-46f4-45a4-a056-d993444c782c" />

(Need an update)
<img width="667" height="184" alt="Image" src="https://github.com/user-attachments/assets/4566fe2b-ffc2-46a4-89af-3779a2c9a8f7" />

Issue #1789, opened 2025-12-18
Use secret key dropdown not shown when selecting a saved keypair in Upload and Deploy Contract


minkyeongshin:
### What did you do?

In **Upload and deploy contract** tab, when selecting a saved keypair as the source account, the `Use secret key` dropdown should appear in signature component. However, the dropdown is not currently being displayed.

<img width="830" height="569" alt="Image" src="https://github.com/user-attachments/assets/eb47f48b-e534-4559-bd07-34ee65a95ff4" />

### What did you expect to see?

[Design file here](https://www.figma.com/design/9j3YQ5URseAsn3j0tf2FFq/Laboratory?node-id=1671-30973&t=THlI3W4m0vqR9q42-1)
Show `Use secret key` dropdown after selecting Saved keypair address.
<img width="730" height="560" alt="Image" src="https://github.com/user-attachments/assets/0062df48-cd77-49bc-a1b0-a6e780912d8f" />

Issue #1703, opened 2025-11-12
Replace public and secret key fields with saved keypairs dropdown input


jeesunikim:
Some examples: operations.source_account, invoke contract's address field, etc.

PR #1968, opened 2026-03-21
Replace public/secret key field with SignerSelector

jeesunikim:
Screenshot 2026-03-20 at 4 45 10 PM
Screenshot 2026-03-20 at 4 41 11 PM

tackles #1703, #1789, #1845

### DIFF
--- a/src/components/FormElements/PubKeyPickerWithSignerSelector.tsx
+++ b/src/components/FormElements/PubKeyPickerWithSignerSelector.tsx
@@ -12,17 +12,21 @@ type PubKeyPickerWithSignerSelectorProps = {
   placeholder?: string;
   value: string;
   error: string | undefined;
-  onChange: (e: React.ChangeEvent<any>) => void;
-  onValueChange?: (val: string) => void;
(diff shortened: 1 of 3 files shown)


src/components/FormElements/PubKeyPickerWithSignerSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onValueChange?: (val: string) => void;
+};
+
+/**
+ * PubKeyPicker wrapped with SignerSelector to allow selecting from
+ * saved keypairs or connected wallet.
+ *
+ * @param onChange - standard input change handler (e.target.value)
+ * @param onValueChange - optional direct value change handler used by
+ *   SignerSelector dropdown (receives the selected address string directly)
```
quietbits: Do we need to describe all the other props as well?

jeesunikim: thanks for flagging - i also think we only need one function here. `onChange` and `onValueChange` are doing p much the same thing

PR #1982, opened 2026-03-25
[New TX] Add 'submit transaction' page

jeesunikim:
Screenshot 2026-03-25 at 2 25 47 PM

Screenshot 2026-03-25 at 2 25 39 PM


src/helpers/scrollElIntoView.ts
```diff
@@ -1,7 +1,7 @@
 import { delayedAction } from "@/helpers/delayedAction";
 
 export const scrollElIntoView = (
-  scrollEl: React.MutableRefObject<HTMLDivElement | HTMLAnchorElement | null>,
```
jeesunikim, line 4: `MutableRefObject` is deprecated

src/app/(sidebar)/transaction/build/components/BuildStepHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Box gap="md" direction="row" justify="space-between" align="center">
+      <PageHeader heading={heading} as={headingAs} />
+
+      {wrapClearAllInText ? (
+        <Text as="div" size="xs">
+          {clearAllLink}
+        </Text>
+      ) : (
```
quietbits: If we only change text size with this wrapper, why not set size on the `<Link size="xs">...`?

quietbits (approved):
Will the mobile version be handled in a separate PR? I see steps only on large screen.

PR #1984, opened 2026-03-26
[New Tx Flow] Add Validate Step

jeesunikim:
- extracted `AuthModePicker` to re-use in both Simulate and Validate
Screenshot 2026-03-25 at 6 03 50 PM

### DIFF
--- a/src/app/(sidebar)/transaction/build/page.tsx
+++ b/src/app/(sidebar)/transaction/build/page.tsx
@@ -21,6 +21,7 @@ import { ClassicTransactionXdr } from "./components/ClassicTransactionXdr";
 import { SorobanTransactionXdr } from "./components/SorobanTransactionXdr";
 import { SimulateStepContent } from "./components/SimulateStepContent";
 import { SignStepContent } from "./components/SignStepContent";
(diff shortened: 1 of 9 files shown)


src/app/(sidebar)/transaction/build/page.tsx
```diff
@@ -44,8 +46,14 @@ export default function BuildTransaction() {
   const { soroban } = build;
   const isSoroban = Boolean(soroban.operation.operation_type);
 
+  const hasAuthEntries = Boolean(
+    simulate.authEntriesXdr && simulate.authEntriesXdr.length > 0,
```
quietbits:
```suggestion
    simulate.authEntriesXdr?.length > 0,
```

quietbits (approved):

src/app/(sidebar)/transaction/build/page.tsx
jeesunikim: claude is a silly boo

jeesunikim: i had to revert because typescript doesn't like comparing `undefined` with `>`

PR #2006, opened 2026-04-03
[New Tx Flow] Fix bugs and update 'restore footprint' & 'extend ttl' functinality

jeesunikim:
- Applied the new store to contract storage to enable link via session storage via url
- Removed the old prepare transaction for restore footprint/extend footprint related operation
- `authMode` to be enabled only for `invoke_contract_function`
- Always show submit via RPC if it's a soroban op and rpc is available
- Fix `View in Dashboard` link. The previous `openUrl` didn't work well with messy url

- restore_footprint build step 
restore-footprint-01

- restore_footprint simulate step
restore-footprint-02

- successfully restored
restore-footprint-03


src/app/(sidebar)/transaction/build/components/SimulateStepContent.tsx
```diff
@@ -231,6 +233,26 @@ export const SimulateStepContent = ({
...
+              const parsedSim = StellarRpc.parseRawSimulation(
+                simBase64Response.result,
+              );
+              const assembled = StellarRpc.assembleTransaction(
+                rawTx,
+                parsedSim,
+              ).build();
+              setAssembledXdr(assembled.toXDR());
+            } catch (e) {
+              console.error("Auto-assembly after simulation failed:", e);
```
quietbits: Do we need to surface this error to the user?

src/app/(sidebar)/transaction/build/components/SubmitStepContent.tsx
```diff
@@ -415,7 +420,11 @@ export const SubmitStepContent = () => {
                   const href = buildEndpointHref(Routes.TRANSACTION_DASHBOARD, {
                     transactionHash: submitHorizonResponse.hash,
                   });
-                  openUrl(href);
+                  window.open(
+                    `${window.location.origin}${href}`,
```
quietbits: We should sanitize the URL just in case (even if we're building it). There is a helper `sanitizeUrl()` method.

quietbits (approved):

src/app/(sidebar)/transaction/build/components/SubmitStepContent.tsx
jeesunikim: replaced with `openUrl(`${window.location.origin}${href}`)`

PR #2013, opened 2026-04-07
[New TX Flow] remove previosu step; fix 'view tx dashboard' link

jeesunikim:
- Remove `previous step` button
- Comment out "Import transaction XDR" for now
- `openUrl`'s `sanitizeUrl` re-serailzed the query string through URLSearchParams which prevented the dashboard page from parsing the hash out of the URL. I replaced it with `window.open`


src/app/(sidebar)/transaction/build/components/SubmitStepContent.tsx
```diff
@@ -306,7 +306,11 @@ export const SubmitStepContent = () => {
                   const href = buildEndpointHref(Routes.TRANSACTION_DASHBOARD, {
                     transactionHash: submitRpcResponse.hash,
                   });
-                  openUrl(`${window.location.origin}${href}`);
+                  window.open(
+                    `${window.location.origin}${href}`,
+                    "_blank",
+                    "noopener,noreferrer",
+                  );
```
quietbits, line 315: If we're not sanitizing URL, we should add a comment to make the decision clear. Otherwise, we'll get security pings about it.

quietbits (approved):

PR #2038, opened 2026-04-15
update js-stellar-sdk and stellar-xdr-json

jeesunikim:


quietbits (approved):
Looks like it still doesn't like something in tests. Updated packages look good.

PR #2066, opened 2026-05-02
Add 'import transaction' step

jeesunikim:
### Adding only the first step:

- import-tx-02
- import-tx-01


src/app/(sidebar)/transaction/build/components/BuildStepHeader.tsx
```diff
@@ -48,7 +48,7 @@ export const BuildStepHeader = ({
       onClearAll={onClearAll}
       rightElement={
         showHeaderButtons ? (
-          <>
+          <div className="additional-buttons">
```
quietbits: We normally use BEM naming for styles, so this wouldn't work.
```suggestion
          <div className="AdditionalButtons">
```

quietbits (approved):
Should we disable "New transaction` and Save buttons when there is no input?

Image

src/app/(sidebar)/transaction/build/components/BuildStepHeader.tsx
jeesunikim: Thanks for flagging this. I ended up not using this className. Removing it

PR #2068, opened 2026-05-05
New "Import Transaction"

jeesunikim:
Big changes are:
- [transaction/components/SimulateStepContent.tsx‎](https://github.com/stellar/laboratory/pull/2068/changes#diff-81fc8220cbdcbd22f6444fcade327dcba58c3834728765f7a2a5039f8b8273e0). Largely the same code as before except that it's updated to go from store-coupled to prop driven so that we can reuse it for both `build tx` and `import tx`.


src/app/(sidebar)/transaction/build/components/SorobanTransactionXdr.tsx
```diff
@@ -9,6 +9,7 @@ import { ValidationResponseCard } from "@/components/ValidationResponseCard";
 import { getTxWithSorobanData } from "@/helpers/sorobanUtils";
 
 import { TransactionXdrDisplay } from "./TransactionXdrDisplay";
+import { NewValidationResponseCard } from "@/components/NewValidationResponseCard";
```
quietbits: Nit: group this import with other component imports.

quietbits (approved):

PR #2125, opened 2026-06-24
Add `@tanstack/eslint-plugin-query` and update `@typescript-eslint/eslint-plugin`

### LINKED ISSUE #1924: Add @tanstack/eslint-plugin-query
`@tanstack/eslint-plugin-query` will give us a warning if we are not following tanstack's best practices

jeesunikim:


src/app/(sidebar)/transactions-explorer/page.tsx
```diff
@@ -53,7 +53,11 @@ export default function Explorer() {
     return map;
   });
 
-  const txsQuery = useGetRpcTxs({
+  const {
+    refetch: refetchTxs,
```
jeesunikim, line 57: the object returned from the query hook is not referentially stable. [the official doc](https://tanstack.com/query/latest/docs/eslint/no-unstable-deps) recommends to destructure the return value of the query hook and pass them into dep array instead.

src/app/(sidebar)/transactions-explorer/page.tsx
```diff
@@ -144,16 +148,16 @@ export default function Explorer() {
...
     localStorageKey,
   ]);
 
-  const errorElement = txsQuery.isError ? (
+  const errorElement = isError ? (
     <Alert variant="error" placement="inline">
-      {txsQuery.error.message}
+      {error?.message}
     </Alert>
   ) : null;
```
quietbits: To be safe, should we me checking for the `error.message` instead of `isError`?

```suggestion
  const errorElement = error?.message ? (
<Alert variant="error" placement="inline">
      {error.message}
</Alert>
) : null;
```

src/query/useAccountInfo.ts
```diff
@@ -11,7 +12,7 @@ export const useAccountInfo = ({
   headers: NetworkHeaders;
 }) => {
   const query = useQuery({
-    queryKey: ["accountInfo", publicKey],
+    queryKey: ["accountInfo", publicKey, horizonUrl, headers],
```
quietbits, line 15: Is `headers` a string or an object? Just want to make sure the query won't be triggered too often.

quietbits (approved):

src/query/useAccountInfo.ts
jeesunikim: it won't be triggered because tanstack hashes those keys ([doc](https://tanstack.com/query/v5/docs/framework/vue/guides/query-keys))!

> Query keys have to be an Array at the top level, and can be as simple as an Array with a single string, or as complex as an array of many strings and nested objects. As long as the query key is serializable using JSON.stringify, and unique to the query's data, you can use it!

PR #2176, opened 2026-08-12
update @stellar/stellar-sdk to v17

quietbits:
- Updated `@stellar/stellar-sdk` to `17.0.0-rc.2` and migrated the XDR API surface — `toXDR`/`fromXDR` were renamed, byte-returning methods now yield `Uint8Array`, and XDR accessors moved from methods to properties
- Extracted the simulate step's auth-entry replacement into `replaceAuthEntries` with tests — v17 requires replacing entries in place on a `readonly`-typed array, where assigning a copy type-checks while silently dropping every signed authorization
- Fixed `TransactionDetails` throwing on legacy `v0` and muxed-source envelopes by narrowing all three envelope variants exhaustively — muxed (`M…`) sources now render `-` rather than crashing the page
- Unwrapped `Uint256Bytes` with `.toBytes()` before `StrKey.encodeEd25519PublicKey` — v17 made the named byte aliases wrapper classes rather than `Uint8Array`, so the explorer detail page threw `RangeError` on every transaction until rc.2's tightened declarations caught it
- Extracted signature-row resolution out of `Signatures` into `resolveSignatureRows` so hint and signature hex decoding is unit-testable
- Sourced the XDR type dropdown from `StellarXdr.types()` instead of `functions(xdr)` — the list came from the JS SDK while decoding runs through `@stellar/stellar-xdr-json`, and v17 drove them apart to 1058 entries with 608 the decoder rejects (was 437/4 on v16)
- Seeded the network in `submitStepContent` tests, which had depended on `NetworkSelector`'s default-network effect landing before the first assertion, unlike the sibling test files
- Added `@exodus/bytes` to the Jest ESM transform allowlist, a new transitive dependency of the SDK
- Marked the CAP-71 v2 auth decision with `TODO: handle CAP-71 v2 auth flag` at five call sites rather than resolving it here — rc.2 defaults `useUpgradedAuth`/`authV2` to `true`, so `prepareTransaction` now requests `ADDRESS_V2` while `useSimulateTx`'s hand-rolled JSON-RPC call still doesn't send the flag

> [!IMPORTANT]
> The CAP-71 v2 auth flag is deliberately left unresolved in this PR, but the behaviour has already changed: `simulateTransaction` now always sends `useUpgradedAuth` on the wire (it previously omitted the field when unset), which older or custom RPC endpoints may reject. The five `TODO: handle CAP-71 v2 auth flag` markers track where the decision needs to be made. When it lands, `txDashSignatures.test.ts` reads `credentials.address` directly and will need the `addressV2` arm. We need to decide whether to enforce v2 flag always or allow users to select depending on their RPC settings.


src/helpers/fetchTxSignatures.ts
```diff
@@ -128,7 +128,7 @@ export const fetchTxSignatures = async ({
               const hashXSigner = StrKey.decodeSha256Hash(signer.key);
               // eslint-disable-next-line no-case-declarations
               const hashXSignature = hash(sigObj.sig);
-              isValid = hashXSigner.equals(hashXSignature);
+              isValid = Buffer.from(hashXSigner).equals(hashXSignature);
```
Ryang-21: You may want to wrap `hashXSignature` in `Buffer`. Apparently the common Buffer polyfill from feross will throw an error when given a `Uint8array`

jeesunikim:
Screenshot 2026-08-19 at 3 47 36 PM

`Contract Spec` isn't showing up when I checked out this branch locally and via preview link. The contract ids I used 
- `CDL74RF5BLYR2YBLCCI7F5FB6TPSCLKEJUBSD2RSVWZ4YHF3VMFAIGWA` (Kale)
- `CDVQVKOY2YSXS2IC7KN6MNASSHPAO7UN2UR2ON4OI2SKMFJNVAMDX6DP` (Blend)

It's related to `@stellar-expert/contract-wasm-interface-parser/` expecting `Buffer`. Two call sites are affected:
- `ContractSpecMeta.tsx:94` (RPC -> Uint8Array)
- `deploy-contract/page.tsx:127` via rpcWasmBinary (line 317) (RPC -> Uint8Array)

src/helpers/getBuildVerification.ts
```diff
@@ -26,16 +26,20 @@ export const getBuildVerification = async ({
...
-      ?.instance()
-      ?.executable();
+    const ledgerEntryData = ledgerEntries.entries[0]?.val;
 
-    const contractType = executable?.switch()?.name;
-    const isSAC =
-      contractType ===
-      xdr.ContractExecutableType.contractExecutableStellarAsset().name;
+    if (ledgerEntryData?.type !== "contractData") {
+      return "unverified_build";
```
jeesunikim, line 36: can we add a comment that this is more "not_found"/"unknown" rather than "unverified"

tests/e2e/submitStepContent.test.ts
```diff
@@ -103,6 +103,21 @@ const seedSessionStorage = async (page: Page, storeState: object) => {
...
   await page.evaluate((stateJson) => {
     sessionStorage.setItem("stellar_lab_tx_flow_build", stateJson);
+
+    // Seed the network so it is set on first render. Otherwise the store starts
+    // with no network and testnet is only applied by NetworkSelector's
+    // "set default network on launch" effect — anything deriving from
+    // network.passphrase (the transaction hash field) is absent until then.
+    localStorage.setItem(
+      "stellar_lab_network",
+      JSON.stringify({
```
jeesunikim, line 113: did you mean to add this?

src/helpers/fetchTxSignatures.ts
quietbits: Good catch, thank you!

tests/e2e/submitStepContent.test.ts
quietbits: Yes, I added this because I kept getting a flaky test. This fixed it.

jeesunikim (approved):

Ryang-21 (approved):

